PR #45, opened 2017-08-30
Add Label Format feature to Bar

dplain90:
#### 1. Simple access

```javascript
const getLabel = getLabelGenerator('value')
getLabel({ value: 12 }) // => 12
```

#### 2. Nested property

```javascript
const getLabel = getLabelGenerator('node.value')
getLabel({ node: { value: 13 } }) // => 13
```

#### 3. Simple access with d3 formatting

```javascript
const getLabel = getLabelGenerator('value', '.1f')
getLabel({ value: 14 }) // => 14.0
```

#### 4. Custom access function

```javascript
const getLabel = getLabelGenerator(d => d.value[0])
getLabel({ value: [15, 16] }) // => 15
```

#### 5. Custom formatting function

```javascript
const getLabel = getLabelGenerator('value', v => v + 10)
getLabel({ value: 17 }) // => 27
```

#### 6. Custom access & formatting functions

```javascript
const getLabel = getLabelGenerator(d => d.value[1], v => v + 10)
getLabel({ value: [18, 19] }) // => 29
```


src/components/charts/bar/BarItemLabel.js
```diff
@@ -8,7 +8,7 @@
  */
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
-
+import { getLabelGenerator } from '../../../lib/propertiesConverters'
```
plouc: It's not used

src/components/charts/bar/BarItemLabel.js
```diff
@@ -26,13 +26,13 @@ export default class BarItemLabel extends Component {
         height: PropTypes.number.isRequired,
         linkColor: PropTypes.string.isRequired,
         textColor: PropTypes.string.isRequired,
+        textFormat: PropTypes.func.isRequired,
```
plouc: It should be named `getLabel` instead

src/components/charts/bar/Bar.js
```diff
@@ -46,7 +47,8 @@ const Bar = ({
     enableLabels,
     getLabelsLinkColor,
     getLabelsTextColor,
-
+    label,
+    labelFormat,
```
plouc: Can you add a blank line after please, I use it to clearly separate each block

src/lib/propertiesConverters.js
```diff
@@ -9,26 +9,15 @@
...
-
-    const label = d => _.get(d, _label)
+export const getLabelGenerator = (_label, labelFormat = '') => {
+    if (_.isFunction(_label)) return _label
 
-    let formatter
-    if (labelFormat) {
-        formatter = format(labelFormat)
-    }
+    const formatter = format(labelFormat)
```
plouc: We should keep the condition and only apply it if it's necessary, I think it can make a difference when we have large graphs

src/lib/propertiesConverters.js
```diff
@@ -9,26 +9,15 @@
...
 
     return data => {
-        let labelOutput = label(data)
-
-        if (formatter) {
-            labelOutput = formatter(labelOutput)
-        }
-
-        return labelOutput
+        let labelOutput = _.isObject(data) ? labelFn(data) : data
```
plouc: I prefer to only handle data, not plain values

src/lib/propertiesConverters.js
```diff
@@ -9,26 +9,15 @@
 import _ from 'lodash'
 import { format } from 'd3-format'
 
-export const getLabelGenerator = (_label, labelFormat) => {
-    if (_.isFunction(_label)) {
-        return _label
-    }
-
-    const label = d => _.get(d, _label)
+export const getLabelGenerator = (_label, labelFormat = '') => {
```
plouc: In fact I'm don't think we have to change this function

test/lib/getLabelGenerator.test.js
```diff
@@ -0,0 +1,40 @@
...
+        expect(labelGenerator('a')).toBe('abcd')
+    })
+
+    describe('should return label with d3 formatting', () => {
+        it(`should format if data is a object & format string is provided`, () => {
+            const label = 'value'
+            const labelFormat = '.1f'
+            const data = { id: 55, name: 'John', value: 22 }
+            const labelGenerator = getLabelGenerator(label, labelFormat)
+            expect(labelGenerator(labelGenerator(data))).toBe('22.0')
```
plouc: I don't understand the double call

plouc: it should just be `expect(labelGenerator(data)).toBe('22.0')`

test/lib/getLabelGenerator.test.js
```diff
@@ -0,0 +1,40 @@
...
+        })
+
+        it(`should format if data is a value & format string is provided`, () => {
+            const labelFormat = '.1f'
+            const label = 'value'
+            const data = 22
+            const labelGenerator = getLabelGenerator(label, labelFormat)
+            expect(labelGenerator(labelGenerator(data))).toBe('22.0')
```
plouc: same here

test/lib/getLabelGenerator.test.js
```diff
@@ -0,0 +1,40 @@
...
+        })
+    })
+
+    describe('should return data object value of label key', () => {
+        it(`should return value if labelFormat is empty`, () => {
+            const labelFormat = ''
+            const label = 'value'
+            const data = { id: 55, name: 'John', value: 22 }
+            const labelGenerator = getLabelGenerator(label, labelFormat)
+            expect(labelGenerator(labelGenerator(data))).toBe('22')
```
plouc: same here

src/components/charts/bar/BarItemLabel.js
```diff
@@ -54,7 +54,7 @@ export default class BarItemLabel extends Component {
             <g transform={`translate(${x},${y})`} className="nivo_bar_legend" style={labelStyle}>
                 {line}
                 <text x={textX} textAnchor={textAnchor} dy="0.5em" style={{ fill: textColor }}>
-                    {data.value}
+                    {textFormat(data.value)}
```
plouc: should be `getLabel(data)` as the `getLabelGenerator()` function handles both accessing & formatting the value

test/lib/getLabelGenerator.test.js
```diff
@@ -0,0 +1,40 @@
...
+    describe('should return label with d3 formatting', () => {
+        it(`should format if data is a object & format string is provided`, () => {
+            const label = 'value'
+            const labelFormat = '.1f'
+            const data = { id: 55, name: 'John', value: 22 }
+            const labelGenerator = getLabelGenerator(label, labelFormat)
+            expect(labelGenerator(labelGenerator(data))).toBe('22.0')
+        })
+
+        it(`should format if data is a value & format string is provided`, () => {
```
plouc: This case should be removed as we don't have to support plain values

src/lib/propertiesConverters.js
plouc: We could just change the logic for formatting

src/lib/propertiesConverters.js
```diff
@@ -9,26 +9,15 @@
...
+    if (_.isFunction(_label)) return _label
 
-    let formatter
-    if (labelFormat) {
```
plouc: Could be changed for:

``` javascript
let formatter
if (labelFormat) {
    if (_.isFunction(labelFormat)) {
        formatter = labelFormat
    } else {
        formatter = format(labelFormat)
    }
}
```

src/lib/propertiesConverters.js
```diff
@@ -9,26 +9,15 @@
...
 import { format } from 'd3-format'
 
-export const getLabelGenerator = (_label, labelFormat) => {
-    if (_.isFunction(_label)) {
-        return _label
-    }
-
-    const label = d => _.get(d, _label)
+export const getLabelGenerator = (_label, labelFormat = '') => {
+    if (_.isFunction(_label)) return _label
```
plouc: Could be changed for:

```javascript
let getRawLabel
if (_.isFunction(_label)) {
    getRawLabel = _label
} else {
    getRawLabel = d => _.get(d, _label)
}
```

src/lib/propertiesConverters.js
```diff
@@ -9,26 +9,15 @@
...
 
-    let formatter
-    if (labelFormat) {
-        formatter = format(labelFormat)
-    }
+    const formatter = format(labelFormat)
+    const labelFn = d => _.get(d, _label)
 
     return data => {
```
plouc: With the previous changes we can offer 2 levels of customization, accessor & formatting:

```javascript
if (formatter) return d => formatter(getRawLabel(d))
return getRawLabel
```

This way, we avoid the condition to be executed for each label retrieval.

plouc:
@dplain90, I've added the use cases I've got in mind, do not hesitate if you think some are missing or if it could be changed.

src/components/charts/bar/Bar.js
```diff
@@ -73,6 +75,7 @@ const Bar = ({
         })
     }
 
+    const getLabelsFormat = getLabelGenerator(label, labelFormat)
```
plouc: This function should be issued by the *enhancer* (`enhance.js`) as if we re-compute it each time we render the bar, it will force React to render each labels because the reference will change each time, if you compute it inside the *enhancer*, it will be re-computed only if `label`or `labelFormat` changed, so we keep the shallow compare ability.

```javascript
// /src/components/charts/bar/enhance.js
export default Component =>
    compose(
        // …
        withPropsOnChange(['label', 'labelFormat'], ({ label, labelFormat }) => ({
            getLabel: getLabelGenerator(label, labelFormat),
        })),
        pure
    )(Component)
```

test/lib/getLabelGenerator.test.js
```diff
@@ -0,0 +1,40 @@
+import { getLabelGenerator } from '../../src/lib/propertiesConverters'
+
+describe('getLabelGenerator()', () => {
```
plouc: 👍

src/components/charts/bar/Bar.js
plouc: Also, it will make the `getLabel()` function available for `BarCanvas` component (but you'll have to add support for it inside this component).

plouc: Sorry, in fact I didn't added labels support in `BarCanvas` :/ But it will be easier for me to do so with this approach :)

dplain90:
Hi Raphael, 

Everything should be squashed and fixed, let me know if this is what you were looking for.

PR #60, opened 2017-09-22
feat(axes): add onClick handler to axis ticks

martingordon:
I found this useful in heat maps to allow for additional interactivity (for example, to allow sorting by a row/column label).


martingordon:
@plouc I fixed the failing tests. Any chance we can get this merged in soon? Please let me know if there are any other changes you want me to make.

plouc:
@martingordon, seems OK, thank you. However I think it's useful to also pass event with value, it can be used to determine click position for example, or prevent bubbling in case we a have a click handler higher in the tree.

martingordon:
@plouc Good idea, I just pushed an update.

debovis:
@martingordon @plouc Can someone update the documentation to reflect this? Thanks

PR #69, opened 2017-10-12
feat(tooltip): prop to format values in tooltip was added

ahtohbi4:
@plouc property to format values in tooltip was added for components Bar, Bubble, Chord, HeatMap, Line, Pie, Radar, Sankey and Stream.

### DIFF
--- a/src/components/charts/bar/Bar.js
+++ b/src/components/charts/bar/Bar.js
@@ -102,6 +102,7 @@ const Bar = ({
 
     // interactivity
     isInteractive,
+    tooltipFormat,
     onClick,
 }) => {
     const options = {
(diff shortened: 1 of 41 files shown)


plouc:
@ahtohbi4, thank you for this nice addition!
Just 2 suggestions:
- It could be nice to have the same logic as other formatting related properties, supporting `d3.format` or custom function
- I tend to prefer flat properties (appart from specific configs such as axes) because it's easier to diff those, so in this case `tooltipFormat`

ahtohbi4:
@plouc, everything was fixed:
- [x] It could be nice to have the same logic as other formatting related properties, supporting d3.format or custom function
- [x] I tend to prefer flat properties (appart from specific configs such as axes) because it's easier to diff those, so in this case tooltipFormat

plouc:
@ahtohbi4, thank you, and you updated the stories 👍

plouc:
I cannot release it for now, will try to do it next week.

PR #70, opened 2017-10-13
[WIP] Feat: stacked tooltip for Bar

ingro:
As the title I wanted to implement something like the stacked tooltip used in the `Line` chart for the `Bar` chart. The PR is WIP, let me know if I'm on the right track (I plan to support horizontal bars too and write some test).


plouc:
@ingro, sorry for the late reply, it seems very interesting, it can surely helps to easily compare data for a given index.
I'm not sure it can easily implemented for canvas implementation though, but it can be added later (or not at all depending on the complexity).

ingro:
Added support for horizontal bar and updated stories, let me know what else I could do!

I'm not very good at working with canvas so I would leave this option just for regular Bar for the moment.

plouc:
While this was an interesting feature, the code has diverged quite a lot since 2017.

Issue #71, opened 2017-10-17
Feature request: Sankey coloring and interaction


bripkens:
First of: Thank you for this nice charting library. You did a fantastic job not only with the library, but also with the documentation and project setup. Well done!

I would like to request some Sankey chart extensions, which I would be willing to contribute.

1) Custom colors for links between nodes (as part of the data on the link objects).
2) Configurable tooltip content (via a getter property: `<Sankey getNodeTooltip={node => <div>{node.id}</div>} getLinkTooltip={…} />`)
3) Click listeners for links and nodes (via `<Sankey onNodeClick={…} onLinkClick={…} />`)

Please let me know whether this is something you would be willing to accept within this library.

plouc:
@bripkens, the custom colors could be nice, and I'm also trying to add more handlers to various charts so no pb for the click handlers :) 

However for the tooltip, I'd like to find a global way to handle this, there's already a PR regarding formatting, maybe it should be handled separately from the sankey diagram.

bripkens:
Thanks, I'll look into colors and click handlers within the next few days.

> However for the tooltip, I'd like to find a global way to handle this, there's already a PR regarding formatting, maybe it should be handled separately from the sankey diagram.

I didn't know about https://github.com/plouc/nivo/pull/69. That might be sufficient for my use case. I'll give it a try once it is merged/released.

bripkens:
Just noticed that link coloring is already supported via:

```
<Sankey linkColorBy={d => {
    console.log(d);
    return 'red';
  }}
```

plouc:
I also forgot its support…

bripkens:
Everything merged and about to be released. Thanks @plouc!

plouc:
Thanks to you, you did all the work ;)

PR #75, opened 2017-10-20
feat(interactivity): add onClick support for Sankey

bripkens:
As discussed in https://github.com/plouc/nivo/issues/71, this adds onClick support for Sankey charts. onClick handlers are registered for both nodes and links and the onClick parameter order is aligned with the order of the bar and bubble charts.


plouc:
Thank you @bripkens

PR #76, opened 2017-10-20
fix(bar): getLabel is defined twice

bripkens:
That was probably copy/pasted at some point. Not a bug, just something that doesn't need to happen twice.


plouc:
Good catch!

PR #77, opened 2017-10-20
feat(sankey): Support configurable labels

bripkens:
This uses the label handling that the bar chart already has and applies it to sankey charts.

Storyboard result:
screen shot 2017-10-20 at 17 03 43


bripkens:
Thanks for these nice turnaround times @plouc! :)

prahna:
Hi, I want this feature too, I see that it's been merged to a branch in Oct 2017 but it still isn't in the main package today. How can I get it please ?

Issue #61, opened 2017-09-23
Line with missing data


plouc:
https://bl.ocks.org/mbostock/0533f44f2cfabecc5e3a

plouc:
fixed by https://github.com/plouc/nivo/pull/81

PR #81, opened 2017-10-28
feat(line): part of a line with missing data was hidden

ahtohbi4:
Hide part of the Line with missing data #61


plouc:
Thank you for your contribution @ahtohbi4

andreas-groos:
In case one has 2 or more line graphs that include null values there is a 'key' issue, I'll open up an issue.
image

PR #82, opened 2017-10-28
feat(line): new prop enableArea was added to Line chart

ahtohbi4:
New property `enableArea` (and `areaOpacity`) was added to decorate Line charts.

enablearea

### DIFF
--- a/src/components/charts/line/Line.js
+++ b/src/components/charts/line/Line.js
@@ -8,7 +8,7 @@
  */
 import React from 'react'
 import { sortBy } from 'lodash'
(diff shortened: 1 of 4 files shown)


plouc:
Nice! thank you.

yiranx6:
How to fill the area just between two lines? I want to have multiple different area merged but they have different limitation'

plouc:
@yiranx6, that's not possible with the line chart, if you want such behavior, you could consider the `@nivo/scatterplot` package with a `none` offsetType which behaves this way.

Issue #99, opened 2017-12-07
Added yarn.lock to .gitignore


kaadash:
I see that yarn.lock has been moved to .gitignore. I think that it should be still in repo (https://github.com/yarnpkg/yarn/issues/1583 ). I've created PR (https://github.com/plouc/nivo/pull/100). I will be really happy and proud of being contributor of such great library. Really good job!

plouc:
@kaadash, you're right, it offers more consistent behavior.

plouc:
Fixed by https://github.com/plouc/nivo/pull/100.

PR #100, opened 2017-12-07
Fix/remove yarn lock from gitignore

kaadash:
https://github.com/plouc/nivo/issues/99


plouc:
@kaadash, thank you for your contribution!

PR #101, opened 2017-12-07
Fix retro build

kaadash:
I just tried to run retro example but it didn't work so I made two changes to run it. I hope that it is good fix :)


plouc:
It seems so :) I'll also run the build command on travis to ensure nothing is broken then :)

PR #116, opened 2017-12-13
Add ability to compose charts, mixing/customizing

plouc:


dmk255:
Hi @plouc  -  just wanted to check in and see if this is no longer going to be considered as a future feature for Nivo or if a different PR is incoming.

Thank you!

plouc:
Hi @dan-kez,

This PR had too many BC breaks, I want to provide this feature, but I have to find a way to introduce it gradually.

steveswork:
Hi @plouc,

All data columns seem to have the same number of fields values. Is there a way to layer multiple datasets of varying number of fields. For example:
first dataset looks like so:
`[
{x: 1, y: 23},
{x: 2, y: 41},
{x: 3, y: 87},
{x: 4, y: 45},
{x: 5, y: 99}
]`
and the second looks like so:
`[
{x: 10, y: 54},
{x: 20, y: 57},
{x: 30, y: 32},
{x: 40, y: 77},
{x: 50, y: 37},
{x: 60, y: 90},
{x: 70, y: 74},
{x: 80, y: 24},
{x: 90, y: 42},
{x: 100, y: 45},
{x: 110, y: 66},
{x: 120, y: 28},
{x: 130, y: 99},
{x: 140, y: 72},
{x: 150, y: 90}
]`

Let's say that the first dataset will be used for the main chart as a bar chart possibly. The second dataset will be scaled and layered on top of the bar chart as a line chart. Is there any way to achieve this with nivo charts?

PR #129, opened 2018-01-16
BasicTooltip allows to format id

nedimb86:
package: nivo-core - modified format function to allow formatting id.
package: nivo-bar - created story to show an usage example for formatting id


plouc:
@nedimb86, thank you for your contribution and I'm sorry for the late review, you can now completely customize the tooltip for bar charts, I think it solve this in a more versatile way.
I close this MR, but let me know if you think there's still a missing use case with the new implementation.

PR #168, opened 2018-04-10
Added option to place markers behind bars

gopeter:
Sometimes it would be useful if the markers could be placed behind the bars instead in front of it.

Like this:

image

I tried to make use of an inline logical operator (`bars && markersBehindBars`), but this won't render the bars at all, so I had to stick with `{(markersBehindBars) ? null : bars}`. Hope this is ok or do you have another idea?

Last but not least: thank you so much for this great charting library! By far the nicest in the React world.

### DIFF
--- a/packages/nivo-bar/src/Bar.js
+++ b/packages/nivo-bar/src/Bar.js
@@ -261,7 +261,6 @@ const Bar = ({
                             left={axisLeft}
                             {...motionProps}
                         />
-                        {bars}
                         <CartesianMarkers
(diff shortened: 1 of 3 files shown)


plouc:
@gopeter, sorry for the very late reaction, I've added a more versatile option to support this feature (https://github.com/plouc/nivo/pull/342), you'll now be able to define order of element using the `layers` property which default to `['grid', 'axes', 'bars', 'markers', 'legends']`.

PR #169, opened 2018-04-10
[WIP] added option to define `gridValues`

gopeter:
As I think that this feature was often requested in the past, I've added another option that was inspired by `tickValues`: `xGridValues` and `yGridValues`.

At first I thought that we could use the `tickValues` value from the `axis` prop, but since there could be two axis for one dimension, we need an extra option.

So this component `<Bar {...commonProps} axisLeft={{ tickValues: [0,50,100] }} gridYValues={[0,50,100]} />` would produce this chart:

bildschirmfoto 2018-04-10 um 23 19 05

If you're fine with this idea/my approach, I would like to add the props to all other components (I've just added it to the `Bar` component in the moment) and add tests :-)

### DIFF
--- a/packages/nivo-bar/src/Bar.js
+++ b/packages/nivo-bar/src/Bar.js
@@ -72,6 +72,8 @@ const Bar = ({
     axisLeft,
     enableGridX,
     enableGridY,
+    gridXValues,
+    gridYValues,
 
     // customization
     barComponent,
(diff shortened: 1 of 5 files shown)


gopeter:
Oh damn, this PR includes the commits from my other PR. Going to create a new one tomorrow :-)

gopeter:
Closed in favor of #170

PR #171, opened 2018-04-12
Implemented the option to add a tag in nodes properties objects

Azkael:
Hi! this is my first contribution ever and I think this would be helpful for Sankey Graph.
What I done is add a custom label for each nodes, that also help if you want to represent cycling dependencies (A -> B -> A), for instance a person who has changed between telecommunication companies.

sankey

What I you have to do is add a tag property in nodes object 

```
"nodes": [
        {"id": "Movistar"},
        {"id": "Claro", "tag": "Claro"},
        {"id": "Cootel", "tag": "Cootel"},
        {"id": "Otro", "tag": "Movistar"}
],
```

**Note**: The tag is not required and sorry if I make a grammatical mistake

### DIFF
--- a/packages/nivo-sankey/src/Sankey.js
+++ b/packages/nivo-sankey/src/Sankey.js
@@ -93,7 +93,7 @@ const Sankey = ({
 
     data.nodes.forEach(node => {
         node.color = getColor(node)
-        node.label = getLabel(node)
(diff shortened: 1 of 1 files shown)


plouc:
@Azkael, thank you for your contribution! What you're implementing is now available by customizing the label property, you could, for example, pick the tag property from the datum.

Issue #47, opened 2017-09-01
Support for events (onClick etc.)


jurajhrib:
I noticed there is an onClick prop in chart component, but it's probably not used for anything right now. Do you plan on implementing some basic events with access to data of clicked element? I am currently trying to implement bar chart with onClick event (to filter in hourly data). It should also have formatted tooltip. I really like your library, it has beautiful visuals and very useful options, but I sadly need onclick for my use case.

plouc:
Hi @jurajhrib ,
Yes, I'm planning to add support for this kind of use case. As you noticed there are already some but not used, in fact I already have all the necessary bits on many charts as I use it for the tooltips.
Which component are you using?

jurajhrib:
I am mostly interested in Bar charts (grouped and regular). My chart should have clickable bars which trigger filtering in list when clicked, also the bar should be able to change colour or other property when it's selected. But that's not as important as the onclick itself.

plouc:
@jurajhrib I've added onClick support for `Bar` & `Bubble` charts, available in `nivo@0.26.1`, the demo has not been updated accordingly though

plouc:
in `nivo@0.27.0`, you'll also have the event as second argument (`onClick(node, event)`), useful in your case if you wish to position something according to it.

debovis:
@plouc Thanks for updating the library to support more events. I am interested in a onClick event for HeatMap component, as well as being able to edit the information shown with tooltip. Can you point me to use onClick if it is already there? Thanks

plouc:
Added onClick support on `TreeMap` components, @debovis should be available soon enough on `HeatMap`, for the tooltip, I plan to allow customization, but don't know when I'll have the time to add it, however feel free to contribute if it's a requirement for your current usage ;)

chrislombaard:
Hi @plouc, i'm using the Heatmap for a project our company is working on. Have you had a chance to add an onClick event for the Heatmap yet?

orarr:
And also Sunburst please? In a moment i thought onClick, onHover(onMouseEnter, onMouseLeave) could be universal functions across all components.

chrislombaard:
@plouc Re: above ^ please

vdidonato:
I would also love to have support for onClick on slices of the Pie component

debovis:
@vdidonato Check out http://recharts.org/en-US/examples/CustomActiveShapePieChart
I had a great experience with implementing Recharts.

vdidonato:
[...]

PR #194, opened 2018-05-19
feat(heatmap): add support for onClick event

plouc:
see https://github.com/plouc/nivo/issues/47


rizo:
Hi, @plouc. Thanks for adding support for `onClick` events to the heatmap. I was testing this now and realized that the events for axis labels are not handled in the same way they are in the non-canvas implementation. It seems like the event in this PR is terminated if it does not target a value node.

Do you have any plans on completing the click handling logic for labels? If not, do you have any suggestions on how this could be implemented so I could give it a try?

Thanks in advance!

PR #198, opened 2018-05-22
Type declaration module for Heatmap package

kodedevil:
Added typescript type definitions for the heat map package. The declarations are in index.d.ts


plouc:
@kodedevil thank you! will probably try to define some common types from the core later and maybe refine the function's input.

kodedevil:
@plouc sure we will do like that

Issue #197, opened 2018-05-22
TypeScript definitions


plouc:
This issue is used to track down the ongoing work on TypeScript definitions for nivo packages.

As nivo isn't built using TypeScript, it does not include definitions, if you're working in a TS environment and already created one for your use case, please do not hesitate to submit a PR!

- [x] `@nivo/bar` *Responsive***Bar** - https://github.com/plouc/nivo/pull/199
- [x] `@nivo/bar` *Responsive***BarCanvas** -  https://github.com/plouc/nivo/pull/199
- [x] `@nivo/calendar` *Responsive***Calendar** - https://github.com/plouc/nivo/pull/223
- [x] `@nivo/calendar` *Responsive***CalendarCanvas**
- [x] `@nivo/chord` *Responsive***Chord**
- [x] `@nivo/chord` *Responsive***ChordCanvas**
- [ ] `@nivo/circle-packing` *Responsive***Bubble**
- [ ] `@nivo/circle-packing` *Responsive***BubbleHtml**
- [ ] `@nivo/circle-packing` *Responsive***BubbleCanvas**
- [x] `@nivo/heatmap` *Responsive***HeatMap** - https://github.com/plouc/nivo/pull/198
- [x] `@nivo/heatmap` *Responsive***HeatMapCanvas** - https://github.com/plouc/nivo/pull/198
- [x] `@nivo/line` *Responsive***Line** - https://github.com/plouc/nivo/pull/316
- [x] `@nivo/line` *Responsive***LineCanvas**
- [x] `@nivo/pie` *Responsive***Pie** - https://github.com/plouc/nivo/pull/207
- [x] `@nivo/radar` *Responsive***Radar**
- [x] `@nivo/sankey` *Responsive***Sankey** - https://github.com/plouc/nivo/pull/211
- [x] `@nivo/scatterplot` *Responsive***ScatterPlot** - https://github.com/plouc/nivo/pull/319
- [x] `@nivo/scatterplot` *Responsive***ScatterPlotCanvas** - https://github.com/plouc/nivo/pull/319
- [x] `@nivo/stream` *Responsive***Stream**
- [x] `@nivo/sunburst` *Responsive***Sunburst**
- [x] `@nivo/treemap` *Responsive***TreeMap**
- [x] `@nivo/treemap` *Responsive***TreeMapHtml**
- [x] `@nivo/treemap` *Responsive***TreeMapCanvas**
[...]

PR #199, opened 2018-05-23
Add TypeScript definitions for Bar package

wyze:
I based this off of #198 with some changes.

- I used the built in `Partial` mapped-type instead of marking all properties optional with `?`. This is why I separated out `data` property as it is required.
- Add types for Legend.
- Add types for Theme.
- There were a few types missing from `Axis` type in the Heatmap PR, this PR contains those.

Reference #197

### DIFF
--- a/packages/nivo-bar/index.d.ts
+++ b/packages/nivo-bar/index.d.ts
@@ -1,0 +1,149 @@
+declare module '@nivo/bar' {
+    export class Bar extends React.Component<Data & BarProps & Dimensions>{}
+    export class BarCanvas extends React.Component<Data & BarProps & Dimensions>{}
+    export class ResponsiveBar extends React.Component<Data & BarProps>{}
+    export class ResponsiveBarCanvas extends React.Component<Data & BarProps>{}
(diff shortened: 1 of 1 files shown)


plouc:
@wyze, +1 for the use of `Partial`, I'll use it for a TypeScript project I'm currently working on! (were duplicating types :/).

packages/nivo-bar/index.d.ts
```diff
@@ -0,0 +1,149 @@
...
+
+        // canvas specific
+        pixelRatio: number;
+    }>
+
+    export type Axis = Partial<{
+        format: string | Function;
+        legend: string;
+        legendOffset: number;
+        legendPosition: 'start' | 'center' | 'end';
```
benembery: I'm currently `ResponsiveBar` and `legendPosition: 'center'` aligns the label off the canvas. However, `legendPosition: 'middle'` works as expected, but doesn’t pass TypeScript validation? Which is correct.

There is also no property for renderTick should this be exposed?

PR #216, opened 2018-06-01
New types and TypeScript test for Pie component

### LINKED ISSUE #215: Problems with Pie TypeScript definition

Psvensso:
Resolves #215 

The original theme prop was copied from the Bar component's .d.ts, it also have the same problems and is not resolved by this PR. These common types (theme, motion legends etc.) should all actually be defined in the @nivo/core package and imported. So the next step would for sure be to create the @nive/core .d.ts and then components after that. 

Again he motion props are copied from the bar defs but should be refined and moved to core package. 

I crated the theme types from the default theme props in core, when moved/implmented in to core this should probably be reworked and make sure all props are there. But for pie this should be enough?

I also added a types folder and a test .tsx files. I think this is good practice, a types folder can contain more types files if needed for a more complex component, and even local tsconfig.json files or lint. 
Even if the pie is simple its nice to keep it consistent.


plouc:
@Psvensso, thank you. I've improved the package and already adjusted the definitions, sorry for wasting your time and let me know if you encounter more problems.

Psvensso:
@plouc No problems, just great to see this fixed!

Issue #218, opened 2018-06-01
Keys are missing from the legend of a vertical grouped bar chart if the value for the first bar is zero


martinalexrolph:
When `Bar.js` picks out the legend data for keys, it looks at all the bars with index 0, i.e. the first group or stack of bars, then uses these to generate the legend.

```javascript
// nivo/packages/nivo-bar/src/Bar.js line 157

const legendDataForKeys = result.bars
    .filter(bar => bar.data.index === 0)
    .map(bar => ({
        label: bar.data.id,
        fill: bar.data.fill ? bar.data.fill : bar.color,
    }))
    .reverse()
```

However, if the value of a given data key on the first bar is 0 then that bar is not included in `result.bars`, presumably to reduce the rendering load by hiding zero-height bars. However, this means that when `legendDataForKeys` looks for bars, it misses out those keys.

One possible fix would be to include zero-height bars by amending `generateVerticalGroupedBars` in `compute/grouped.js` so that it does not exclude bars if they have no height.

```javascript
// nivo/packages/nivo-bar/src/compute/grouped.js line 88

if (barWidth > 0) {
    keys.forEach((key, i) => {
        range(xScale.domain().length).forEach(index => {
            const x = xScale(getIndex(data[index])) + barWidth * i + innerPadding * i
            const y = getY(data[index][key])
            const barHeight = getHeight(data[index][key], y)

            if (barWidth > 0 /*&& barHeight > 0*/) { // Remove this condition
                const barData = {
                    id: key,
```

This has the disadvantage that we lose the performance benefit of not rendering zero-height bars so it should be possible to filter them out *after* the legend keys have been generated.

```javascript
// nivo/packages/nivo-bar/src/Bar.js line 162

    }))
    .reverse()

const legendDataForIndexes = result.bars.filter(bar => bar.data.id === keys[0]).map(bar => ({
    label: bar.data.indexValue,
    fill: bar.data.fill ? bar.data.fill : bar.color,
}))

// Add filter here
result.bars = result.bars.filter(bar => bar.height === 0)

return (
    <Container isInteractive={isInteractive} theme={theme}>
        {({ showTooltip, hideTooltip }) => {
            const commonProps = {
                borderRadius,
                borderWidth,
```

martinalexrolph:
Just spotted that this is a duplicate of #111 which is fixed by https://github.com/plouc/nivo/pull/128, but in a different way.

plouc:
[...]

PR #128, opened 2018-01-16
Fix legendDataForKeys() missing data #111

stahlmanDesign:
Stacked bar charts currently assume the first bar has all the items for the legend. This fix looks at all bars. https://github.com/plouc/nivo/issues/111

Issue #111, opened 2017-12-12
legendDataForKeys() missing data


lgan1989:
``` javascript
 var legendDataForKeys = result.bars.filter(function (bar) {
        return bar.data.index === 0;
    }).map(function (bar) {
        return {
            label: bar.data.id,
            fill: bar.data.fill ? bar.data.fill : bar.color
        };
    }).reverse();
```

Consider following data:

```json
[
  {
    key1: 0,
    key2: 100,
    key3: 100,
    idx: 0
  },
  {
    key1:100,
    key2:300,
    key3:400,
    idx:1
  }
]
```

Key1 would be missed in legendDataForKeys, as it seems that result.bars only contains those that have visible heights?

knackjax:
@lgan1989 I am seeing missing records as well, seems to be keying off the first data point for me as well

stahlmanDesign:
You guys are right, it only shows legends for the first bar. If that stacked bar only has 1 colour, then the legend only has one item.
![legend-keys](https://user-images.githubusercontent.com/972804/34273120-0aef3036-e661-11e7-8fa3-abd268733ed2.gif)

knackjax:
@stahlmanDesign thanks your fix works for me!

stahlmanDesign:
I thought of a workaround for this issue https://github.com/plouc/nivo/pull/128 while waiting for feedback on the fix: 
Basically, put a small value on the first bar that goes unseen but gets picked up by the legend

jeremypeters:
@stahlmanDesign Thanks - your fix is working for me as well.

leonkyr:
This hack does the trick...

```
if (data.items && data.items.length > 0) {
    Object.keys(data.items[0]).forEach((key: string) => {
         if (!data.items[0][key]) data.items[0][key] = 0.05;
    });
}
```

plingampally:
@stahlmanDesign Fix does work but it produces console log errors, as the data structure has changed. 

The object should look like this:

```
var legendDataForKeys = result.bars.map(function (bar) {
   return {
     id: bar.data.id,
     label: bar.data.id,
     color: bar.color,
     fill: bar.data.fill ? bar.data.fill : bar.color
   };
 }).reverse();```

superdopey:
When will this fix be released?

PR #219, opened 2018-06-01
Missing legend keys on bar charts with zero data values

martinalexrolph:
Fix for https://github.com/plouc/nivo/issues/111 and https://github.com/plouc/nivo/issues/218 (duplicate).

See https://github.com/plouc/nivo/issues/218 for a detailed explanation of the issue and approach but the summary is that by filtering out zero-height bars before generating the legend, if a bar doesn't show in the first group of bars (because it has a value at or close to zero) it doesn't appear in the legend. To fix, remove this filter and add filtering just before the graphs are rendered.

I think this approach is better than https://github.com/plouc/nivo/pull/128 because with that method it appears that if a key has value 0 across the whole range covered by the graph, it will not appear in the legend. In our use case this is quite common as you can set the date range of the graph and some of the keys may have no data in them for certain periods of time. The key should still appear in the legend so it is clear that the value is zero, rather than just not included in the graph.

### DIFF
--- a/packages/nivo-bar/src/Bar.js
+++ b/packages/nivo-bar/src/Bar.js
@@ -167,7 +167,8 @@ const Bar = ({
         fill: bar.data.fill ? bar.data.fill : bar.color,
     }))
 
-    result.bars = result.bars.filter(bar => bar.height === 0)
+    result.bars = result.bars
+        .filter(bar => (layout === 'vertical') ? (bar.height !== 0) : (bar.width !== 0)) 
 
     return (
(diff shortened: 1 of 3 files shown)


taschetto:
@plouc is there a workaround while this PR isn't merged?

plouc:
Fixed by https://github.com/plouc/nivo/pull/362, it doesn't cover the case where all values are `0`, but I'm not sure we should show the legend as nothing will correspond in the chart.

PR #220, opened 2018-06-03
feat(pie): improve pie components

plouc:


plouc:
@MartijnDevNull, `width` & `height` are only required for `Pie`, not `ResponsivePie`,  and defined through `Dimensions` type https://github.com/plouc/nivo/blob/master/packages/nivo-pie/index.d.ts#L76. I forgot to use `Partial` helper for `MotionProps`, `ColorProps<PieDatum>` & `SvgDefsAndFill<PieDatum>` though :/

Issue #177, opened 2018-04-23
Number y value on ResponsiveLine


th0th:
I have data like the following:

```
const data = [
    { x: 1, y: 10 },
    { x: 2, y: 20 },
    { x: 3, y: 30 },
];
```

And I am rendering chart:

```
<ResponsiveLine
    data={[
        { id: 'data', data },
    ]}
    curve="monotoneX"
/>
```

Chart renders OK, but I get a warning on the console:

> Warning: Failed prop type: Invalid prop `slices[0].id` of type `number` supplied to `LineSlices`, expected `string`.

Warning goes away when I use strings for x axis but the document says x can be a number, too.

In my case, x is unix timestamp and values should be handled accordingly.

plouc:
Fixed by https://github.com/plouc/nivo/pull/222

PR #222, opened 2018-06-04
fix(line): fix LineSlices id prop type

plouc:
see https://github.com/plouc/nivo/issues/177


riteshnus:
I still getting the error for this issue. It still required string, whereas we pass number. you changed to oneOftheType in 0.42.1 to accept any of them. But it didn't fix. Could you suggest me something? I have upgraded the @nivo/line to 0.42.1. Also, I cannot upgrade the version to 0.43.0, as npm cannot find the latest version.

jwchia1:
same things happens for responsive pie when i use a number for the data id

PR #231, opened 2018-06-25
fix bug for Microsoft IE11, which does not support str.endsWith

benzitohhh:
Without the below fix, IE11 throws an error on loading the code.

### DIFF
--- a/packages/nivo-core/src/props/curve.js
+++ b/packages/nivo-core/src/props/curve.js
@@ -54,7 +54,12 @@ export const curvePropKeys = Object.keys(curvePropMapping)
 
 export const curvePropType = PropTypes.oneOf(curvePropKeys)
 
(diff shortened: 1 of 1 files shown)


benzitohhh:
Hmmm not sure why integration test failed.

Error message is
```
"The command "make fmt-check --no-print-directory" exited with 2.
0.34s$ make packages-lint --no-print-directory
Running eslint on all packages"
```

Any ideas?

plouc:
@benzitohhh thank you for this, but sorry I won't accept this PR, if I start to include polyfills for various environments directly into nivo, the lib will quickly become bloated :)
You could be able to include a polyfill in your own build.

benzitohhh:
@plouc ok cool no worries - thanks for looking at this

Issue #246, opened 2018-07-18
Calendar crashes when no data is present


mdorr:
If no data is present (yet), the calendar component crashes with the following error:
```
Uncaught TypeError: Cannot read property 'value' of undefined
    at nivo-calendar.js:510
    at new WithPropsOnChange (withPropsOnChange.js:52)
   [...][1]
```
This causes an issue if p.e. the component is still waiting for data to be loaded asynchronously. The error is caused by in `/packages/nivo-calendar/src/enhance.js`:

```js
if (domain === 'auto') {
  colorDomain = [minBy(data, 'value').value, maxBy(data, 'value').value]
}
```

If no data is given, minBy/maxBy will return `undefined`, and then `.value` is called on that. To prevent this from happening, `_.get` can be used, which will safely return `undefined` instead of throwing an error. I'll submit a pull request in a moment.

--
[1] Error message shortened for clarity. Full trace in attachment: [nivo-calendar-error.txt](https://github.com/plouc/nivo/files/2206972/nivo-calendar-error.txt)

PR #247, opened 2018-07-18
Fix calendar crashing when no data is available

mdorr:
#246

### DIFF
--- a/packages/nivo-calendar/src/enhance.js
+++ b/packages/nivo-calendar/src/enhance.js
@@ -12,6 +12,7 @@ import withPropsOnChange from 'recompose/withPropsOnChange'
 import pure from 'recompose/pure'
 import minBy from 'lodash/minBy'
 import maxBy from 'lodash/maxBy'
+import get from 'lodash/get'
 import { scaleQuantize } from 'd3-scale'
 import { withTheme, withDimensions } from '@nivo/core'
 import { CalendarDefaultProps } from './props'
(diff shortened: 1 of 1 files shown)


packages/nivo-calendar/src/enhance.js
```diff
@@ -24,7 +25,7 @@ export default Component =>
         withPropsOnChange(['data', 'domain', 'colors'], ({ data, domain, colors }) => {
             let colorDomain
             if (domain === 'auto') {
-                colorDomain = [minBy(data, 'value').value, maxBy(data, 'value').value]
+                colorDomain = [get(minBy(data, 'value'), 'value'), get(maxBy(data, 'value'), 'value')]
```
plouc: a more simple way to handle this could be:

```javascript
if (data.length === 0) {
    colorDomain = [0, 0] 
}
```

This way, we do not have to call four functions when we know we'll get nothing, what do you think?

mdorr:
Good point, that is better.

plouc:
The build [fails on travis](https://travis-ci.org/plouc/nivo/builds/405512169), can you please run `make fmt`?

mdorr:
Sorry, missed that step, thanks for the heads up. Build should be good now.

plouc:
@mdorr, no pb :) thank you for your help!

mdorr:
Thank you for this great tool as well!

PR #255, opened 2018-08-02
Custom tooltips on ScatterPlot + optionally pass a function to the 'symbolSize' prop

mjsarfatti:
First of all, sorry if I post two improvements in one pull request, it's the first time I do this. Let me know if you prefer it done in some other ways.

**Improvement 1** adds custom tooltips to ScatterPlot (I basically followed what had been done in the Bar component).
I added them to the SVG component, but not to the canvas one. That's something that might have to be done before merging? Let me know your thoughts.

**Improvement 2** allows the passing of a function to 'symbolSize'. This is something that's often useful with scatter plots because it allows representing a third dimension.
I've touched @nivo/core, so I would feel better if you reviewed it. I'm not 100% sure of what I've done :)

For neither of them I've written tests. But you can see from the two added stories ('symbol size function' and 'custom tooltip') that they work.
But again, let me know if you'd prefer to have some tests (and any guidance on that would be super appreciated).

Manuele

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,18 +22,18 @@ yarn add @nivo/bar @nivo/sankey ...
 
 ## Features
 
--   Highly customizable
--   Motion/transitions, powered by [react-motion](https://github.com/chenglou/react-motion)
--   [Component playground](http://nivo.rocks)
--   [Exhaustive documentation](http://nivo.rocks)
--   Isomorphic rendering
--   [SVG charts](http://nivo.rocks/components?filter=svg)
--   [HTML charts](http://nivo.rocks/components?filter=html)
--   [Canvas charts](http://nivo.rocks/components?filter=canvas)
--   [server side rendering API](https://github.com/plouc/nivo-api)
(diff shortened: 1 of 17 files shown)


plouc:
@mjsarfatti the build fails because I think you used a newer version of prettier, maybe you're using `npm` and the repo only contains `yarn.lock` file

plouc:
would you mind trying to install deps using yarn and re-formatting? It adds a little bit of noise to the PR.

mjsarfatti:
So I deleted the ./node_moduels folder, then _$ yarn_ and _$ make fmt_ and it seems to have done the trick.

plouc:
@mjsarfatti, SVG support can be added later, SVG is far more used than canvas for now :)
For tests, I need to add some basic ones on all packages and add some guidance in the contribution guide.
👍 for added stories

plouc:
I've added the ability to use custom tooltip from the website demo, in the "interactivity" pane.

PR #259, opened 2018-08-12
fix(heatmap): better handling of NaN values

martingordon:
When a NaN value is included in heatmap data, no errors appear but two warnings are raised:
* The prop `node.color` is marked as required in `HeatMapCellTooltip`, but its value is `undefined`.
* Received NaN for the `children` attribute. If this is expected, cast the value to a string.

This commit addresses both of these, and adds a `nanColor` prop to `HeatMap` to allow specifying the color used for a cell that is NaN (defaults to #000000).


plouc:
@martingordon, thank you for your contribution

PR #265, opened 2018-08-23
LineDots labelFormat PropType can be string or function

plemarquand:
Supplying a formatting function to `dotLabelFormat` on the Line graph type
results in a prop type validation error, however the label is formatted
correctly.

The `getLabelGenerator` function is responsible for actually formatting
this label, and accepts a string or a function as a formatter.


plouc:
Right! Thank you for the fix @plemarquand!

Issue #243, opened 2018-07-12
ScatterPlot: minX, maxX, minY, maxY exist in default props but are never used


plouc:

plouc:
Fixed by https://github.com/plouc/nivo/pull/266

PR #266, opened 2018-08-24
feat(scatterplot): remove unused min/max x/y

plouc:
https://github.com/plouc/nivo/issues/243


serchavalos:
Hey @plouc 👋 Could someone update these docs too? I got a bit confused when I saw the `minY` property in the docs thinking it was still used
http://nivo.rocks/line

PR #273, opened 2018-08-25
add legend support for @nivo/waffle package

plouc:


packages/waffle/src/enhance.js
```diff
@@ -54,6 +54,14 @@ const commonEnhancers = [
             }),
         }
     }),
+    withPropsOnChange(['computedData'], ({ computedData }) => ({
+        legendData: computedData.map(datum => ({
+            id: datum.id,
+            label: datum.id,
```
optyler, line 60: This shall be `label: datum.label`

PR #287, opened 2018-09-11
feat(scatterplot-markers): add markers to scatterplot SVG

mjsarfatti:
As the title said, simply added markers to scatterplot SVG

### DIFF
--- a/packages/scatterplot/src/ScatterPlot.js
+++ b/packages/scatterplot/src/ScatterPlot.js
@@ -9,7 +9,7 @@
 import React from 'react'
 import { TransitionMotion, spring } from 'react-motion'
 import { Container, SvgWrapper } from '@nivo/core'
-import { Grid, Axes } from '@nivo/core'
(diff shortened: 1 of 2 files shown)


plouc:
@mjsarfatti, thank you! however there are some conflicts, you probably created the branch from your other PR regarding mouseenter/mouseleave, can you please rebase against master?

packages/scatterplot/src/ScatterPlot.js
```diff
@@ -36,6 +36,9 @@ const ScatterPlot = ({
     enableGridX,
     enableGridY,
 
+    // markers
+    markers,
```
plouc: The corresponding propType should be added (see https://github.com/plouc/nivo/blob/master/packages/line/src/props.js#L64).

mjsarfatti:
Alright, I think I'm starting to get a hang of git and pull requests :) 
Rebased the branch on v0.49.1 and added proptypes.

PR #291, opened 2018-09-13
WIP feat(annotations): working example for scatterplot SVG

mjsarfatti:
So, this is the famouse **annotations** :)

It's basically a wrapper for react-annotation, which you can see here: http://react-annotation.susielu.com/
I found that library flexible, clean and well structured.

The library provides lower level highly customizable components, and a dozen pre-packaged annotation _types_ (which are simply preset combinations of the low level components). You can see these pre-packaged _types_ in the section "Annotation Types" of the link above.

Given Nivo is a high level library, I thought best to wrap the high level _types_, and not give access to the low level components.

I have worked only on the scatterplot at the moment, and you can see an example in the storybook.

The way you add annotations to your chart is by simply passing an object which mirrors react-annotation props, plus a "type" property to define which of the _types_ you want to render (the storybook is a lot clearer than I can explain I think!)

### DIFF
--- a/packages/annotations/src/props.js
+++ b/packages/annotations/src/props.js
@@ -1,0 +1,67 @@
+/*
+ * This file is part of the nivo project.
+ *
+ * Copyright 2016-present, Raphaël Benitte.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+import PropTypes from 'prop-types'
+
+export const AnnotationPropTypes = {
+    type: PropTypes.oneOf([
+        'label',
+        'callout',
+        'calloutCircle',
+        'calloutCurve',
+        'calloutElbow',
+        'calloutRect',
+        'xyThreshold',
+        'badge',
+        'bracket',
+        'calloutCustom',
+    ]).isRequired,
+    xScale: PropTypes.func.isRequired, // ought to be a d3 scale fn
+    yScale: PropTypes.func.isRequired, // ought to be a d3 scale fn
+
+    // to scale
+    x: PropTypes.number.isRequired,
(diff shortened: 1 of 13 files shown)


packages/annotations/src/props.js
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * Copyright 2016-present, Raphaël Benitte.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+import PropTypes from 'prop-types'
+
+export const AnnotationPropTypes = {
+    type: PropTypes.oneOf([
```
plouc: Might be easier to maintain if we use `Object.keys(annotationComponentsById)`

plouc:
@mjsarfatti, how does the library behaves regarding animation? I'm afraid it will be weird if the chart move smoothly from one state to another while the annotations immediately move to their destination.

mjsarfatti:
Right, I didn't tackle that at all.
Being a third party library, it might be difficult and CPU-expensive to transition the position of annotations from a state to the next, but an alternative could be to fade-out/fade-in.
The whole experience would be:
1. All annotations start to fade out as the bars/circles/elements transition to their new position and size
2. New annotations start to fade in and complete at the same time as the rest of the elements
I will have to give it a try.
It would be cool if they transition-moved to their new locations though.

plouc:
@mjsarfatti, thank you for your time, finally I've created a new package not based on react-annotations, which supports animations/canvas and avoid manipulating the DOM to get element's dimensions as it's not compatible with the HTTP rendering API. See https://github.com/plouc/nivo/commit/e0dc149dc14e34c18d9816dfb769875b255123be

PR #305, opened 2018-10-03
Pass `tickIndex` to Axis’ renderTick method

benembery:
Decide whether or not to render a tick based on the index. 
For example dependent on viewport width you could modulo the `tickIndex` to only render *n* ticks.


benembery:
@plouc I’ve added to the non animated too now.

PR #340, opened 2018-11-12
Add tooltip to LineProps

ArjanJ:
Allows the `tooltip` prop to be used on `<Line>` or `<ResponsiveLine>` when you want to have a custom tooltip. I'm not sure if the `(data: LineDatum): React.ReactNode` is the best annotation, but it compiles fine for me, open to someone improving that though.

### DIFF
--- a/packages/line/index.d.ts
+++ b/packages/line/index.d.ts
@@ -76,6 +76,7 @@ declare module '@nivo/line' {
         enableStackTooltip?: boolean
 
         legends?: LegendProps[]
+        tooltip?(data: LineDatum): React.ReactNode
     }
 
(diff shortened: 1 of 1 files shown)


plouc:
@ArjanJ, thank you! `ReactNode` seems fine, I guess it's versatile enough to cover all kind of needs :)

ArjanJ:
@plouc I'm not sure why the build is failing. It mentions `itemOpacity` doesn't exist on the `LegendProps` type, but according to this file https://github.com/plouc/nivo/blob/master/packages/legends/index.d.ts it exists 🤔

zepplock:
Any update?

PR #362, opened 2018-11-30
Fix missing legend in Bar chart 

### LINKED ISSUE #111: legendDataForKeys() missing data

AhyoungRyu:
Another trial to fix #111 or #218 

So the problem that I aim to fix here is, if some keys have `0` as value in first column, it doesn't generate legend at all even if other columns have non-zero value. 

Related conversation is also in #128 which is a first trial to fix this issue. I saw #219 as well and I understand there might be more cases we can improve. 
But I think many ppl including me, this missing legend with zero value in 1st column issue is more urgent I think. 

tl; dr 

#### Before 
screen shot 2018-11-30 at 2 17 13 pm

#### After 
screen shot 2018-11-30 at 2 17 04 pm

Also applied in Canvas 
screen shot 2018-11-30 at 2 12 51 pm

### DIFF
--- a/packages/bar/src/Bar.js
+++ b/packages/bar/src/Bar.js
@@ -7,7 +7,7 @@
  * file that was distributed with this source code.
  */
 import React, { Fragment } from 'react'
-import uniqBy from "lodash/uniqBy"
+import uniqBy from 'lodash/uniqBy'
 import { TransitionMotion, spring } from 'react-motion'
 import { bindDefs, Container, SvgWrapper, Grid, CartesianMarkers } from '@nivo/core'
 import { Axes } from '@nivo/axes'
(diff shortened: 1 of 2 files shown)


justincrow:
awesome - when will this fix be published?

AhyoungRyu:
Yeah I also wondered too ;) 
@plouc When will be the next release ?

stahlmanDesign:
I saw this was finally fixed with v0.52.0!
capture d ecran 2018-12-10 a 17 37 06

AntonPuko:
ehhh..can this be evaluated not only  in the first bar?
with current calculations its not possible to use legend in realtime chart like this:
image
where the first bar can be empty.
:(

PR #372, opened 2018-12-07
Fix wrong legend item color in ScatterPlotCanvas

### LINKED ISSUE #371: ScatterPlotCanvas legend item color is different with actual plot's
See how it's different with SVG one 

#### SVG 
screen shot 2018-12-07 at 6 50 02 pm

#### Canvas
screen shot 2018-12-07 at 6 50 15 pm

AhyoungRyu:
Aiming to fix #371 

### Before 
screen shot 2018-12-07 at 6 50 15 pm

### After
screen shot 2018-12-07 at 6 54 51 pm

### DIFF
--- a/packages/scatterplot/src/ScatterPlotCanvas.js
+++ b/packages/scatterplot/src/ScatterPlotCanvas.js
@@ -216,13 +216,14 @@ class ScatterPlotCanvas extends Component {
         const legendData = data.map(serie => ({
             id: serie.id,
             label: serie.id,
-            color: getColor(serie),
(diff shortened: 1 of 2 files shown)


plouc:
The ScatterPlotCanvas example should be updated too, https://github.com/plouc/nivo/blob/master/website/src/components/charts/scatterplot/ScatterPlotCanvas.js#L50, using `serie.id` instead of `id`.

AhyoungRyu:
@plouc I see. Thanks for the explanation!

PR #375, opened 2018-12-08
Make scatter plot item size have always positive value by wrapping it with Math.abs()

### LINKED ISSUE #374: Can't handle negative value in ScatterPlot
I saw some related issues like #249 and #337 but I think it'll be nicer there is safer workaround to handle this kind of situation. 
In my case, Chrome browser just becomes unresponsive.

 - With ScatterPlot Canvas api 
screen shot 2018-12-07 at 11 45 30 pm

 - With SVG api 
screen shot 2018-12-07 at 11 46 20 pm

AhyoungRyu:
Aiming to resolve #374 

### How to replicate 
Put some negative values in sample data in `ScatterPlot.stories.js` & `ScatterPlotCanvas.stories.js` -> Add 
```
yScale: {
  type: 'linear',
  min: 'auto',
  max: 'auto',
},
```
to `commonProps` to see the negative sample data is displayed in the graph  -> Go to "ScatterPlot - Varying Symbol Size" example (same in Canvas storybook example)

#### Before 
In SVG ScatterPlot example 
screen shot 2018-12-08 at 2 58 47 pm

In ScatterPlotCanvas example 
screen shot 2018-12-08 at 2 58 27 pm

Those error should be gone with this patch

### DIFF
--- a/packages/scatterplot/src/ScatterPlotCanvas.js
+++ b/packages/scatterplot/src/ScatterPlotCanvas.js
@@ -200,7 +200,7 @@ class ScatterPlotCanvas extends Component {
 
         points.forEach(point => {
             this.ctx.beginPath()
-            this.ctx.arc(point.x, point.y, getSymbolSize(point.data) / 2, 0, 2 * Math.PI)
+            this.ctx.arc(point.x, point.y, Math.abs(getSymbolSize(point.data) / 2), 0, 2 * Math.PI)
             this.ctx.fillStyle = getColor(point.data)
             this.ctx.fill()
         })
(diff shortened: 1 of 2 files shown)


plouc:
@AhyoungRyu, thank you, however while I see a case where we end up having negative values for the canvas implementation when animating (where the values often exceed limits while transitioning), I don't see how it could happen in the canvas implementation, except if we use an improper symbol sizing function, which is the user's responsibility IMHO.

AhyoungRyu:
@plouc Yeah I was using customized function to calculate `symbolSize`. That was the problem. Thank you for the explanation!

Issue #151, opened 2018-02-13
Bundle size


cedmax:
Hello, 
I might be wrong, but it looks like Nivo is importing the whole lodash-es in the bundle to use only a tiny subset of it. 

According to WebPackBundleAnalyzerPlugin the size of lodash-es exceeds the one of Nivo itself.
Is there any chance you would consider optimising the imports to include only the needed parts?

This is the screenshot from the webpack plugin with the responsive line example from your website
<img width="1192" alt="screenshot 2018-02-13 17 13 13" src="https://user-images.githubusercontent.com/415593/36163643-de9334f0-10e1-11e8-8ed0-1fdbb109be6a.png">

marcosfede:
Is there any update on this?  The size of the imported javascript is very large, and you also download it multiple times if you are going to use a different type of graph

![image](https://user-images.githubusercontent.com/16232610/38647106-00f7b4c4-3dc1-11e8-9c6b-1906f12c08f3.png)

those two bundles correspond to the @nivo/line and @nivo/bar modules, which are not sharing common code

marcosfede:
I'd like to help solving this. Could you point me in a general direction of what needs to be done?

plouc:
Moved to rollup & use now use scoped imports (`@nivo/<package>@0.35.2+`).

cedmax:
Thanks, sir :)

PR #387, opened 2018-12-16
fix(lodash): use scoped imports

cymen:
I think #151 came back with a few non-scoped lodash imports. Prior #191.


cymen:
I tried adding an entry in Makefile checking if `git grep lodash|grep import|grep {` returned 0 (found) or 1 (not found) but got confused in the Makefile syntax. It's a bit of a hacky grep but might be nicer to avoid repeats?

PR #391, opened 2018-12-18
Add number support to gridXValues and gridYValues on Line

jasonwells:
This commit adds gridCount/integer support to `gridXValues` and `gridYValues` for line charts. This allows for grid lines and axes ticks to be calculated the same when using a number that is passed. For example:

```javascript
<ResponsiveLine
  axisRight={{
    tickValues: 6,
  }}
  gridYValues={6}
  ...
/>
```

I tried to follow all of the contributing guidelines, happy to contribute to a great library!

### DIFF
--- a/packages/line/src/props.js
+++ b/packages/line/src/props.js
@@ -66,11 +66,15 @@ export const LinePropTypes = {
     enableGridY: PropTypes.bool.isRequired,
     gridXValues: PropTypes.oneOfType([
         PropTypes.number,
-        PropTypes.arrayOf(PropTypes.oneOfType([PropTypes.number, PropTypes.string])),
+        PropTypes.arrayOf(
+            PropTypes.oneOfType([PropTypes.number, PropTypes.string, PropTypes.instanceOf(Date)])
+        ),
     ]),
     gridYValues: PropTypes.oneOfType([
         PropTypes.number,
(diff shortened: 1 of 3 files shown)


packages/line/src/props.js
```diff
@@ -64,8 +64,14 @@ export const LinePropTypes = {
 
     enableGridX: PropTypes.bool.isRequired,
     enableGridY: PropTypes.bool.isRequired,
-    gridXValues: PropTypes.arrayOf(PropTypes.oneOfType([PropTypes.number, PropTypes.string])),
-    gridYValues: PropTypes.arrayOf(PropTypes.oneOfType([PropTypes.number, PropTypes.string])),
+    gridXValues: PropTypes.oneOfType([
+        PropTypes.number,
+        PropTypes.arrayOf(PropTypes.oneOfType([PropTypes.number, PropTypes.string])),
```
plouc: The line chart also supports Date objects, so you should add `PropTypes.instanceOf(Date)` to both `gridXValues` and `gridYValues`.

plouc:
@jasonwells sorry for the late review

packages/line/src/props.js
jasonwells: Thanks for the feedback, I wasn't sure about that. I've added it now.

jasonwells:
@plouc Thanks for the input and no worries!

plouc:
Thank you @jasonwells!

Issue #393, opened 2018-12-18
Adding `defaultCurrentNode` prop in HeatMap


ncknuna:
We have a need to call out a specific row/column/cell in a heatmap based on external data (e.g. "this is the particular cell that's in effect based on your current settings"), and I thought that the best way to do that would be to allow providing a `defaultCurrentNode` prop which would get used in the `computeNodes` process here (https://github.com/plouc/nivo/blob/262a8ee96bb870fe388e64d1453248aed94bf445/packages/heatmap/src/computeNodes.js#L58) if `currentNode` is falsy

Would you be open to a PR that adds this? Or is there an alternative you would prefer?

plouc:
Yes, I'm ok with this approach with the current implementation, but I'd like to change the heatmap behaviour as it could be more versatile to have the ability to customise node attributes the way we want (using onMouseEnter/onMouseMove/onMouseLeave + overrides).

ncknuna:
@plouc Could you elaborate a bit more on your desired implementation? If it's not too big of a lift I'd consider taking some of it on

plouc:
The idea is to completely remove the currentNode from the state and let users decide the styles of the nodes from the outside, for example you could have default color/border/size and have the ability to override those properties using an `overrides` property for example which will accept a matcher and properties, the same way patterns and gradients work, then you could use mouse event to define the matcher managing the state by yourself, it will require more customisation but will allow to cover almost every use cases.

ncknuna:
Ah, I see... that does seem like a bigger lift. I'll just submit a PR for the easy fix with the current implementation for now :)

PR #394, opened 2018-12-18
Add ability to set highlight on heatmap by default

ncknuna:
Wasn't sure if you wanted me to also update the website, since this is a feature you want to be done in a different way in the longer term


ncknuna:
Fixes #393

ncknuna:
@plouc Could you take a look at this when you get the chance? :)

roger-schaer:
I am really interested in this feature as well: I would like to highlight a specific column in the heatmap upon hovering on a point in a DIFFERENT chart (scatter plot), to "synchronize" the two charts in a way. Is there a reason why this pull request was never approved? Thank you very much for your feedback, it would be great to have this working.

Issue #370, opened 2018-12-07
Unexpected mouse cursor type in ScatterCanvas when "isInteractive" is true 


AhyoungRyu:
![scatterplot](https://user-images.githubusercontent.com/10060731/49634017-db5cab80-fa3e-11e8-81b8-9c76d61540e6.gif)

As you can see, if `isInteractive` is true, then the mouse cursor type is changed to `crosshair` from `normal`. It's because of this line https://github.com/plouc/nivo/blob/master/packages/scatterplot/src/ScatterPlotCanvas.js#L247. 
I just wonder it's intended behaviour or not. If not, I wanna just set as `normal` regardless `isInteractive` value.

PR #402, opened 2018-12-22
Set pointer as 'normal' not crosshair in ScatterPlot canvas chart

AhyoungRyu:
Fixes https://github.com/plouc/nivo/issues/370 

#### Before
record_before

#### After
record_after

### DIFF
--- a/packages/scatterplot/src/ScatterPlotCanvas.js
+++ b/packages/scatterplot/src/ScatterPlotCanvas.js
@@ -243,7 +243,7 @@ class ScatterPlotCanvas extends Component {
                         style={{
                             width: outerWidth,
                             height: outerHeight,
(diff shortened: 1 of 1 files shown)


AhyoungRyu:
/cc @plouc I'm not sure this behaviour is intended or not(If so, plz let me know I'll close this one). 
But I think it'll be better if we have consistent behaviour between SVG and Canvas chart.

plouc:
I liked the crosshair, but you're right, consistency is preferable. However the property should just be removed, this way we'll be able to customise it from the parent container.

AhyoungRyu:
@plouc Thanks for the reply. I just removed it in latest commit f37f34183c45d0cb90fddbb69208936241958fb8

PR #404, opened 2018-12-23
Feature: Decouple Sankey Node coloring and Link coloring

ByronBecker:
This PR aims to make it much easier to customize node and link colorings for the sankey plot, to be able to create a plot like this

screen shot 2018-12-23 at 14 02 54

Now one can customize link gradients from the node color when using enable gradient by adding the startColor and endColor properties to each link's data. Otherwise, this will default to the color of each node.

### DIFF
--- a/packages/sankey/src/SankeyLinksItem.js
+++ b/packages/sankey/src/SankeyLinksItem.js
@@ -59,6 +59,12 @@ const SankeyLinksItem = ({
     path,
     width,
     color,
+
+    //Decouples link colors from node color
+    defaultColor, //If no linear gradient
+    startColor, //Use startColor & endColor if linear gradient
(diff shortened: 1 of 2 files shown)


ByronBecker:
@plouc One thing that I did find quite confusing was that using the getColor and getLinkColor properties don't point to functions with the same name in Sankey.js, as those functions are mapped to respectively by the colorBy, and linkColorBy properties. It took some time to figure out, and would be helpful if the sankey documentation was updated somewhere to reflect this.

ByronBecker:
That makes sense - it was just something new to me and took awhile before I found the connection.

plouc:
Thanks @ByronBecker, that's a useful feature 👍

PR #406, opened 2019-01-01
Updates slice width to allow for unevenly spaced data.

### LINKED ISSUE #229: Tooltips on line graphs that are to dense dont line up with the mouse
If a line graph which uses tooltips that are less then 40px apart the LineSliceItems start to overlap.  Hovering over a point means you select the points hitbox to the right that overlaps it. 

image

https://github.com/plouc/nivo/blob/f04660f2ff1cd965b0a9d2609782e0409cb486d5/packages/nivo-line/src/LineSlicesItem.js#L39

allisonlucas:
The slice width is currently hard coded to be 40px wide. This causes problems when the data on the graph is either closer together than 40px or farther apart than 40px. The slices then either overlap, causing the tooltip position to be off, or are too far apart, causing there to be spaces where the tooltip does not show. This pull request fixes this issue by making the width of the slices variable and determined by the position of the previous data point and the next data point on the graph.

Fixes #229.

### DIFF
--- a/packages/line/src/LineSlices.js
+++ b/packages/line/src/LineSlices.js
@@ -11,30 +11,62 @@ import PropTypes from 'prop-types'
 import pure from 'recompose/pure'
 import LineSlicesItem from './LineSlicesItem'
 
-const LineSlices = ({
-    slices,
-    height,
-    showTooltip,
-    hideTooltip,
-    theme,
-    tooltip,
-    tooltipFormat,
-}) => (
-    <g>
-        {slices.map(slice => (
-            <LineSlicesItem
-                key={slice.id}
-                slice={slice}
-                height={height}
-                showTooltip={showTooltip}
-                hideTooltip={hideTooltip}
-                theme={theme}
-                tooltipFormat={tooltipFormat}
-                tooltip={tooltip}
-            />
-        ))}
-    </g>
-)
+class LineSlices extends React.Component {
+    render() {
+        const {
+            slices,
+            height,
+            showTooltip,
+            hideTooltip,
+            theme,
+            tooltip,
+            tooltipFormat,
+        } = this.props;
+        // sliceWidth: half the distance from the previous point + half the distance to the next point
+        let sliceWidth;
+        // xOffset: half the distance from the previous point
+        let xOffset;
+        let lineSlicesItems = [];
+        for (let i = 0; i < slices.length; i++) {
+            const currentXPosition = slices[i].data[0].position.x;
+            const previousXPosition = i > 0 ? slices[i - 1].data[0].position.x : null;
+            const nextXPosition = i < slices.length - 1 ? slices[i + 1].data[0].position.x : null;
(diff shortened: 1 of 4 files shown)


BogdanTaranenko:
+1

plouc:
Please, do not add comment containing `+1`, it doesn't add any value to the original issue, you can use reactions instead.

packages/line/src/LineSlices.js
```diff
@@ -11,30 +11,62 @@ import PropTypes from 'prop-types'
...
+        // sliceWidth: half the distance from the previous point + half the distance to the next point
+        let sliceWidth
+        // xOffset: half the distance from the previous point
+        let xOffset
+        let lineSlicesItems = []
+        for (let i = 0; i < slices.length; i++) {
+            const currentXPosition = slices[i].data[0].position.x
+            const previousXPosition = i > 0 ? slices[i - 1].data[0].position.x : null
+            const nextXPosition = i < slices.length - 1 ? slices[i + 1].data[0].position.x : null
+            const widthBetweenCurrentAndNext = nextXPosition - currentXPosition
```
R-K-H: What happens if two nulls are subtracted? It might be better to instead have zeroed values or have a minimum default for intended behavior?

packages/line/src/LineSlices.js
```diff
@@ -11,30 +11,62 @@ import PropTypes from 'prop-types'
...
+            theme,
+            tooltip,
+            tooltipFormat,
+        } = this.props
+        // sliceWidth: half the distance from the previous point + half the distance to the next point
+        let sliceWidth
+        // xOffset: half the distance from the previous point
+        let xOffset
+        let lineSlicesItems = []
+        for (let i = 0; i < slices.length; i++) {
```
R-K-H: This might be intense for load and calculation for many points? Would it be better to instead say at larger slices it is defaulted? Or a tiered process?

packages/line/src/LineSlicesItem.js
```diff
@@ -25,41 +25,45 @@ Chip.propTypes = {
...
-        {isHover && (
-            <line
-                x1={0}
-                x2={0}
-                y1={0}
-                y2={height}
-                stroke="#000"
-                strokeOpacity={0.35}
-                strokeWidth={1}
+const LineSlicesItem = ({ slice, height, width, showTooltip, hideTooltip, isHover, xOffset }) => {
```
R-K-H: Is this the correct order would you rather have xOffset next to height and width?

packages/line/src/LineSlices.js
R-K-H: I think it would also depend on the chart width in the window as well. A smaller div would mean you have less space for the points to work?

plouc:
@allisonlucas, Thank you and sorry but finally I've used another approach and now compute slices using points instead of series, I also added support for both `x` and `y` slices, see https://github.com/plouc/nivo/pull/529/commits/1c085bf0bd8e13852f59b869a10b8a99a5ba5362

PR #412, opened 2019-01-07
fix(typescript): Allow axis to be AxisProps or `null`

DrBenton:
Setting an axis to `null` is the only way to disable it, so we have to specify it as an authorised value in the TypeScript definitions.


DrBenton:
Oh, thank YOU for this awesome library! :-)

mcantor:
This same problem crops up in the BarSvgProps.  Currently can't compile the demo code for the RepsonsiveBar for this reason.

Issue #419, opened 2019-01-13
TickValues setting doesn't work in canvas bar chart 


AhyoungRyu:
I draw a bar chart with 
```
keys:  ['job(sum)']
data: [
    {id: "19", "job(sum)": 1},
    {id: "20", "job(sum)": 2},
    {id: "17", "job(sum)": 4},
    {id: "1", "job(sum)": 3},
    {id: "2", "job(sum)": 99},
    {id: "3", "job(sum)": 60},
    {id: "4", "job(sum)": 79},
    {id: "5", "job(sum)": 8},
    {id: "6", "job(sum)": 9},
    {id: "23", "job(sum)": 10},
    {id: "29", "job(sum)": 11},
    {id: "27", "job(sum)": 12},
    {id: "33", "job(sum)": 16},
    {id: "32", "job(sum)": 20},
    {id: "87", "job(sum)": 3},
    {id: "65", "job(sum)": 23},
    {id: "55", "job(sum)": 49},
    {id: "45", "job(sum)": 30},
    {id: "49", "job(sum)": 37},
]
```

and set `axisBottom`
```
axisBottom: {
        legendPosition: 'middle',
        legendOffset: 50,
        tickSize: 0,
        tickPadding: 12,
        tickValues: ["1", "5", "17", "33", "87"], <- this is not applied in BarCavas chart but works in Bar (SVG) 
    },
```

Same data and prop setting work in Bar chart but not in BarCanvas.

PR #420, opened 2019-01-13
[BarCanvas] Fix tickValues setting in BarCanvas chart

AhyoungRyu:
Fixes https://github.com/plouc/nivo/issues/419

I tried `tickValues` setting in `ScatterPlotCanvas` chart (as another example of canvas chart) and found that the setting works in there.
I think it's because two canvas chart components use different `renderAxesToCanvas` util func.

 - ScatterPlotCanvas : `renderAxesToCanvas` is come from  `@nivo/axes` -> https://github.com/plouc/nivo/blob/master/packages/scatterplot/src/ScatterPlotCanvas.js#L11
 - BarCanvas : `renderAxesToCanvas` is come from  `@nivo/core` -> https://github.com/plouc/nivo/blob/master/packages/bar/src/BarCanvas.js#L13

As you can see `nivo/core`'s `renderAxesToCanvas` doesn't accept `tickValues` whereas `nivo/axes`'s does. 
 - `nivo/axes/canvas` -> https://github.com/plouc/nivo/blob/master/packages/axes/src/canvas.js#L22
 - `nivo/core/lib/canvas/axes` -> https://github.com/plouc/nivo/blob/master/packages/core/src/lib/canvas/axes.js#L15 

I don't know which one is legacy or deprecated one. Or there must be a reason that there are two `renderAxesToCanvas` in two places `nivo/axes` &  `nivo/core`?

Anyways, #419 is gone with this patch.

####  - Without this patch
screen shot 2019-01-13 at 5 29 14 pm

####  - With this patch
screen shot 2019-01-13 at 5 28 41 pm

### DIFF
--- a/packages/bar/src/BarCanvas.js
+++ b/packages/bar/src/BarCanvas.js
@@ -10,13 +10,13 @@ import React, { Component } from 'react'
 import uniqBy from 'lodash/uniqBy'
 import setDisplayName from 'recompose/setDisplayName'
 import {
-    renderAxesToCanvas,
     getRelativeCursor,
     isCursorInRect,
     Container,
     BasicTooltip,
     renderGridLinesToCanvas,
 } from '@nivo/core'
+import { renderAxesToCanvas } from '@nivo/axes'
 import { renderLegendToCanvas } from '@nivo/legends'
 import { generateGroupedBars, generateStackedBars } from './compute'
 import { BarPropTypes } from './props'


plouc:
@AhyoungRyu, you used the good one :), `@nivo/axes` is the new package while `@nivo/core` axes is deprecated, but not all components have been updated.
Thank you!

PR #425, opened 2019-01-15
feat(line): pass showTooltip and hideTooltip functions to custom layers

johnolos:
Allow showing tooltips from custom layers.

Edit: BasicTooltip and TableTooltip was not exported from ts-definitions so added that as well.


plouc:
Thank you @johnolos, the typescript definitions will also be useful for other packages 👍

PR #428, opened 2019-01-17
fix(typescript): Fix TableTooltip and LineComputedSerieDatum-type on line

johnolos:
Slight fix for TableTooltip, sorry about that.

Also, I found `LineComputedSerieDatum` useful exported subtype when creating custom layers upon Line. Basically for when customizing computedData with additional effects.

```jsx
  getCustomLayer = (props: CustomProps) => {
    const { xScale, yScale, computedData } = props;
    const { train, overrides } = this.props;
    const areaGenerator = area<LineComputedSerieDatum>()
      .x(d => xScale(d.data.x))
      .y0(() => yScale(0))
      .y1(d => yScale(d.data.y))
      .curve(curveMonotoneX);
    ...
    return <path
        ...
        d={areaGenerator(data)}
      />;
  };
```

@plouc How accepting would you be of a PR adding `defs` to `Line`?
Line isn't using it directly, but custom layers might, and therefore it would be helpful to add patterns to the main SVG element.

Work-around is to have a seperate svg element with pattern defined in the custom layer itself.

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -97,7 +97,7 @@ declare module '@nivo/core' {
 
     export interface TableTooltipProps {
         title?: React.ReactNode
-        rows?: React.ReactNode[]
+        rows: React.ReactNode[][]
         theme: Pick<Theme, 'tooltip'>
         renderContent?: () => React.ReactNode
     }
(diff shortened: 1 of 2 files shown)


plouc:
@johnolos `defs` definition should be defined inside the `core` package IMO

johnolos:
@plouc Fair. I'll leave it for now.

plouc:
Thank you for your contribution @johnolos

PR #432, opened 2019-01-21
fix(imports): replace root "lodash" package imports with more specific ones

DrBenton:
This prevents the whole "lodash" package to be imported in the resulting JavaScript files when using bundlers that don't do advanced tree-shaking.


plouc:
@DrBenton, I'm closing this as it was already covered by https://github.com/plouc/nivo/pull/387 which has been merged. Thank you

PR #441, opened 2019-01-28
Fix the wrong title on the guides/axes page

frooeyzanny:


plouc:
Good catch @frooeyzanny!

Issue #64, opened 2017-10-02
Custom tick and label formatter (Bar chart)


devmilind:
Is there a way to format tick and label values? I noticed there is `sliceLabel` and `radialLabel` which can be a `function` in pie charts. Can we have a similar option for bar chart? The tick and label values I have is high 1000s. So I would like to display them with a `K` suffix. For example. `130K` instead of `130000` which it would display by default

plouc:
Yes you can, for example:

```javascript
<Bar axisLeft={{ format: v => `${v}%` }} />
```

SachaG:
That would be really helpful to have in the docs :)

plouc:
@SachaG, you're right, the feature has been added without updating the documentation, but the website is now part, it should now be easier to align documentation with new features.
BTW, I'll add it.

plouc:
Reopened until documentation reflect this feature.

spredemann:
I still can't figure out how to format the tick values to be abbreviated numbers. The labelFormat=".0s" works fine but using a function for axisLeft format am I expected to use my own format implementation or can I tap into the same built in function?

plouc:
@spredemann, you can use either custom function or d3-format specifier, you can see both in action [here](https://codesandbox.io/s/nrw0xpvw9j).

spredemann:
@plouc thanks for the quick response. Very easy, don't know how I missed that. I was trying to return the d3-fromat specifier from the function, duh

stahlmanDesign:
I tried `labelFormat` on a heatmap and it doesn't seem to be supported (it works on the axis). It seems `colorBy` is not supported either. Often I'd like to have a different label than the value used to colour it. For example, show `13/15` as the label but have the value be 13.

stahlmanDesign:
In the line chart's dotLabelFormat I can't use `dotLabelFormat: '.0%'` because my data is already '12.0', not '0.12' %. So I used the function ```dotLabelFormat: v => `${v}%` ``` which works, but throws a prop types error because it expects a string.
``` Invalid prop `labelFormat` of type `function` supplied to `LineDots`, expected `string`.```

I'm sure a fix would be as simple as allowing a string OR function as accepted prop types.

Stekot:
I have a same issue with `dotLabelFormat` expecting `string`. Also documentation on the site could be updated.

gijsbotje:
[...]

PR #449, opened 2019-02-04
doc(Bar): Add labelFormat prop to Bar docs

philjestin:
Added the formatLabel prop to the Bar chart documentation. I was looking for this and didn't notice it in the diverged staked story, and didn't see the prop was an option until I saw the conversation in #64 . Figured this might save someone some time in the future.


plouc:
Thank you! keeping the docs up-to-date is a hard task :)

Issue #301, opened 2018-09-28
IE 11 Compatibility


fabrusso:
Hello,
I'm trying to integrate a ResponsivePie into my React Project.
It's not working in IE 11.

The component, with the graph inside of it, is never loading (I'm using react-router) while the others pages are still ok.

I've created a Component with the  ResponsivePie, and imported it into another Component, to put the graph in a <div> with a specific height (as suggested for the responsive graph).

P.S.: I'm not able to load the "nivo rocks" website with IE11

plouc:
@fabrusso, I'm really sorry to read that :/ what error do you have?

AlexCLeduc:
Would it be possible to update the demo website to work in IE11 ? The served JS isn't transpiled at all right now

plouc:
@AlexCLeduc, sorry but I have to disagree :) https://github.com/plouc/nivo/tree/gh-pages/static/js, of course the code is transpiled.

plouc:
Will try to test it using a windows VM

AlexCLeduc:
Woops, you're right, TIL IE11 supports source maps! 

When I open up the demo website in IE11, I get an issue with `.includes` being undefined. Including one of those catch-all polyfills like babel-polyfill would definitely fix that issue.

plouc:
I've updated `browserslist` to include `not ie < 11`.

serchavalos:
Hello! I would like to re-open this issue; I saw this error when I visited the site

![IE11](https://user-images.githubusercontent.com/587388/55520885-1bb8f700-56ca-11e9-8ba5-81c769ad538f.JPG)

Platform: Win7

fabrusso:
Hello,

sorry if I answer late.

Right now, the site is displayed on IE11 but seems to report numerous errors like:

> 
> Syntax Error
> CONST must be initialized
> "Missing resources for /"

When I try to enter the pie demo page, this error is shown in the console:

> Missing resources for / pie /
> And it is not possible to view either the graph or the source code.

I would like to understand if it is only the DEMO site that does not work or it is not possible to use the library on IE11 ... because I am having problems visualizing the components on the Microsot browser.

AlexCLeduc:
@fabrusso it's just the demo site, the library itself works fine in IE11.

plouc:
Well, finally I dropped IE support for the demo website as it requires far too much work and maintaining the documentation is already quite time consuming.
[...]

PR #453, opened 2019-02-08
Fixed the website for internet explorer

meganhng:
Fixes: #301 

Babel was transpiling stuff properly, but IE was still lacking basic functions like `Array.prototype.includes` 

The css is still defective but the website is now functional on IE 11 and older browsers.


plouc:
The website make use of create-react-app, which uses @babel/preset-env under the hood.
Sorry but I don't plan to support older browsers for the website, maintaining the packages + examples + documentation + storybook is already pretty time consuming :/

PR #463, opened 2019-02-19
Allow non-inherited Sunburst child colors

### LINKED ISSUE #332: Sunburst child colorization
Hi,

Is it possible to customize secondary node colorization on sunburst chart ?
The demo dataset is setting `color` keys for children elements, but seems never used.

Function `colorBy` seems to be called only on primary nodes.

Is there a way to customize secondary nodes ?
`childColor` behaviour is strange when using function, as the argument passed is always a primary node, why not pass the secondary node too ?

Why not calling `colorBy` method on each node (primary, secondary, n-ary, ...) ?

**Nivo version**

```
"@nivo/sunburst": "^0.50.0",
```

retypepassword:
Fix #332

### DIFF
--- a/packages/sunburst/src/Sunburst.js
+++ b/packages/sunburst/src/Sunburst.js
@@ -157,7 +157,7 @@ const enhance = compose(
     })),
     withPropsOnChange(
         ['data', 'partition', 'getIdentity', 'getChildColor'],
-        ({ data, partition, getIdentity, getColor, getChildColor }) => {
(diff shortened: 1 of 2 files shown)


plouc:
Thank you @retypepassword, can you please add an example of this in the storybook?

retypepassword:
@plouc I've added an example in the storybook. Thanks for the awesome library!

plouc:
@retypepassword, thank you for your contribution!

PR #482, opened 2019-03-18
Update TS declaration for LineSliceDate

serchavalos:
Hello! 👋

I noticed the type definition for `LineSliceData` was not matching the actual implementation. Here what I saw in the console:
Skärmavbild 2019-03-19 kl  10 11 01 fm

Looking at the code also confirmed my suspicions
https://github.com/plouc/nivo/blob/master/packages/line/src/LineSlices.js#L39-L73

It's a quick fix, but do let me know if there's something else

### DIFF
--- a/packages/line/index.d.ts
+++ b/packages/line/index.d.ts
@@ -40,8 +40,10 @@ declare module '@nivo/line' {
         id: string | number | Date
         x: number
         data: Array<{
-            x?: string | number | Date
-            y?: string | number | Date
+            data: {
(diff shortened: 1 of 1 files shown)


serchavalos:
hello @plouc! do you agree with this change?

plouc:
@serchavalos, yes :)

serchavalos:
@plouc thank you! I hope I didn't sound like rushing; I copied-pasted this file on my repo and seeing duplicated code makes me shiver 😆

PR #488, opened 2019-03-21
Add a minBarLength (Sizing) Option to @nivo/bar

ByronBecker:
I've found that when dealing with large data quantities, the d3-scaleLinear function will sometimes return bars that have no width.

An example of this is seen below.

Screen Shot 2019-03-15 at 13 24 59

To fix this, one can set `minBarLength={x}`, where x is a number in order to ensure that all bars width a measureable quantity will render with that min length. If a data value is zero, it will still not render a bar, as expected.

The result of this change with `minBarLength={3}` is below

Screen Shot 2019-03-15 at 13 25 40

Note that this change works for both vertical and horizontal chart orientations. In the case of a horizontal chart, `minBarLength` will apply to each bar's width, whereas in a vertical chart, `minBarLength` will apply to each bar's height

Additionally, this feature will work for both grouped and stacked bars, with an example of this below.

Stacked bar chart w/o `minBarLength`

Screen Shot 2019-03-15 at 13 25 14

Stacked bar chart with `minBarLength={3}`

Screen Shot 2019-03-15 at 13 25 48

The one possible edge case where users can run into trouble is if they are using a stackedBarChart with a either an extremely large minBarLength, or many stacked BarItem sections that trigger the minBarLength resize adjustment. This can potentially cause bars to expand past the max size of the chart. I think, however, if developers have this in mind when testing it out, and are reasonable in their `minBarLength` usage, this will be an extremely helpful feature.

### DIFF
--- a/packages/bar/index.d.ts
+++ b/packages/bar/index.d.ts
@@ -117,6 +117,8 @@ declare module '@nivo/bar' {
             legends: Array<{ dataFrom: 'indexes' | 'keys' } & LegendProps>
 
             markers: CartesianMarkerProps[]
+
+            minBarLength: number
         }>
 
     export type Axis = Partial<{
--- a/packages/bar/src/Bar.js
+++ b/packages/bar/src/Bar.js
@@ -121,7 +121,7 @@ const Bar = props => {
         getColor,
         padding,
         innerPadding,
-        minBarLength
+        minBarLength,
     }
     const result =
         groupMode === 'grouped' ? generateGroupedBars(options) : generateStackedBars(options)
--- a/packages/bar/src/Bar.js
+++ b/packages/bar/src/Bar.js
@@ -104,6 +104,8 @@ const Bar = props => {
         onMouseEnter,
         onMouseLeave,
 
+        minBarLength,
+
         legends,
     } = props
     const options = {
@@ -119,6 +121,7 @@ const Bar = props => {
         getColor,
         padding,
         innerPadding,
+        minBarLength
     }
     const result =
         groupMode === 'grouped' ? generateGroupedBars(options) : generateStackedBars(options)
--- a/packages/bar/src/compute/grouped.js
+++ b/packages/bar/src/compute/grouped.js
@@ -94,7 +94,7 @@ export const generateVerticalGroupedBars = ({
                 let y = getY(data[index][key])
                 let barHeight = getHeight(data[index][key], y)
 
-                if (minBarLength && minBarLength > 0 && barHeight < minBarLength) {
+                if (minBarLength && minBarLength > 0 && barHeight < minBarLength && 0 !== data[index][key]) {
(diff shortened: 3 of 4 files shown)


plouc:
@ByronBecker I understand the problem, however I already had several issues suggesting min/max width/height… and it's very weird to me to have bars which have appears to have a positive value while they don't.
I appreciate the effort you put in this, but I'm not willing to integrate that kind of *hack*, sorry.
If you want to display zero values, you might use another chart type, for example an heatmap, which will show a cell for every item and use color encoding to express the *strength*.

plouc:
Also, as you mentioned, as this chart supports a bunch of combinations, layout/groupMode/reverse, it's almost impossible to get a robust solution.

ByronBecker:
@plouc I may not have gotten this across in my initial post, but the main problem is described in the first image I linked, the value for the bar Native Hawaiian or Other Pacific Islander tick is **not zero**, it has a value of 2. However, due to the scaling by the d3.scaleLinear method and the size of the chart, all values that are deemed too small to warrant a pixel are not rendered. For a user then, they are unable to hover over this non-existent bar (which has a positive value) and view a tooltip describing that the value is non-zero. To the user, the value perceived is zero, even though it isn't zero in reality. To clarify what was included in the PR, the changes still respect the fact that **values that are equal to zero are not rendered**. This is assigned to the variable ```const doesBarHaveData = d.data[stackedDataItem.key```, and evaluated for its truthiness before deciding if it should be given a height/width of minBarLength, and a bar rendered. 

While I do agree that in some cases the HeatMap is a valid chart to use, they can be slightly more difficult to interpret from a UX point of view, and aren't always as clean when you have a dynamically changing number of categories that may become HeatMap cells. In my opinion, the HeatMap is amazing at displaying data that has 2 categories with distinct values, and 1 category with continuous values, and doesn't necessarily fit the case provided above as well.

I agree that the pr is a tad _"hackish"_, as you described, but I'm open to feedback on how to improve it if you are willing to go in that direction. Also, I'm willing to expand the pr to work for the reverse option, as well as other options (it already works for groupMode stacked/grouped, and layout vertical/horizontal). However, I respect if you don't want this type of feature included (and will keep it as a fork if so), but if you're not willing to incorporate this feature, I'm curious about how you recommend handling the _"non-zero bar value not rendering a bar because of linear scaling and chart size"_ issue described in the first paragraph.

Thanks again for the review 👍

akosborn:
@plouc Given @ByronBecker's further explanation, are you still opposed the proposed change?

An example of the issue can be seen here: https://codesandbox.io/s/nivo-basic-demo-256lb?file=/src/index.js

In the example, you can see that values of 0 (Quarter 2) are treated identically to relatively small non-zero values (Quarter 3), which is a bit misleading.

If you're still opposed to the change, how would you recommend handling this?

Issue #437, opened 2019-01-23
Vertical Sankey


cl-evgenyb-rnd:
Is there an easy way to rotate the Sankey diagram 90 degrees?

Something like this:
![image](https://user-images.githubusercontent.com/46893495/51611518-17cd6380-1f28-11e9-8f89-1ce27ec51bb8.png)

plouc:
Not for now, but it could be interesting to have, however your example is more a dendogram (which is a work in progress).

gpalrepo:
Listing here for the convenience of someone else who might have had a similar need:
Nivo now supports vertical layouts for Sankey diagrams. 
**Include a layout prop and pass either "vertical" or "horizontal" into it.**

PR #490, opened 2019-03-22
Improve @nivo/sankey package

plouc:
- improve nodes sorting feature with shortcuts
- add sort property to the *component playground*
- move the layout computation outside of the rendering component
- add support for vertical layout | fix https://github.com/plouc/nivo/issues/437


ByronBecker:
Looks good so far, thanks for adapting the sort to support the null option.

Issue #496, opened 2019-03-27
Issue with Sankey in newly released versions


ByronBecker:
Upon checking out the newest release of the sankey packages (having merged in #401 & #404), I was planning on updating to the current version of the @nivo/sankey package, but am having this issue when testing out the most recent version.

<img width="1374" alt="Screen Shot 2019-03-27 at 16 11 52" src="https://user-images.githubusercontent.com/17368530/55118654-47295980-50ac-11e9-9eb3-7a01b26dc4af.png">

Here's the transpiled esm code (although I'm not sure that this helps)

<img width="560" alt="Screen Shot 2019-03-27 at 16 11 41" src="https://user-images.githubusercontent.com/17368530/55118631-3678e380-50ac-11e9-8d70-f6ce474d54af.png">

I'm going to investigate this issue, but thought I'd post it first here. @plouc Let me know if you have any ideas as to what might be causing this error.

ByronBecker:
Also, on a side note, I've noticed on storybook a small bug that the sankey stories that needs to be updated with the code change. This led me to find what I think might be a bug in the `packages/sankey/src/enhance.js` file...

```
withPropsOnChange(['sort'], ({ sort }) => {
            let sortFunction
            if (sort === 'input') {
                sortFunction = null
            } else if (sort === 'ascending') {
                sortFunction = (a, b) => a.value - b.value
            } else if (sort === 'descending') {
                sortFunction = (a, b) => b.value - a.value
            }

             return { sortFunction }
        }),
```
You aren't actually returning the sorting function if the user specifies one :)

Secondly, I would recommend making the code in the non-story book (roll the dice) playground contain stacked nodes, so that the sorting asc/desc/input has a more noticeable effect.

plouc:
@ByronBecker, you'll have to upgrade React.
You should had a warning when you installed the new version.

plouc:
There's definitely a bug, I forgot to handle this case, thank you for noticing.

plouc:
Fixed by https://github.com/plouc/nivo/pull/498

PR #498, opened 2019-03-28
Fix custom sankey node sorting

ByronBecker:
Fixes issue/bug referenced in comments of #496


plouc:
Thank you @ByronBecker!

PR #506, opened 2019-04-04
Add custom tooltips for Chord

yanatan16:
This PR adds the `arcTooltip` and `ribbonTooltip` properties to the `Chord` chart in the same style as `tooltip` props to other chart types. I could not get the storybook to show the Chord (it only built with 4 component types, instead of all of them...), but I was able to test locally in another project.

### DIFF
--- a/packages/chord/src/Chord.js
+++ b/packages/chord/src/Chord.js
@@ -43,6 +43,8 @@ const Chord = ({
 
     isInteractive,
     tooltipFormat,
+    arcTooltip,
+    ribbonTooltip,
 
     animate,
     motionDamping,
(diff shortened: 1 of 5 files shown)


plouc:
@yanatan16, as chord is the 5th required file in the storybook config, I guess there's a problem with the new stories.

yanatan16:
Ok makes sense. I'll debug it

yanatan16:
@plouc This Chord bug in the storybook appears to be in master. Its somewhere in `Chord.js` but I can't get any error output of the storybook.

yanatan16:
no problem @plouc thanks!

Issue #507, opened 2019-04-04
ScatterPlotCanvas onClick TypeError in plot margin


jvansan:
### Minor Bug Report

While tinkering with nivo ScatterPlotCanvas in my app I discovered a react-dom warning when building with webpack development mode. (Does not appear in sandbox).

```
react-dom.development.js?61bb:289 Uncaught TypeError: Cannot read property 'data' of null
    at eval (nivo-scatterplot.esm.js?95a7:631)
    at HTMLUnknownElement.callCallback (react-dom.development.js?61bb:149)
    at Object.invokeGuardedCallbackDev (react-dom.development.js?61bb:199)
    at invokeGuardedCallback (react-dom.development.js?61bb:256)
    at invokeGuardedCallbackAndCatchFirstError (react-dom.development.js?61bb:270)
    at executeDispatch (react-dom.development.js?61bb:561)
    at executeDispatchesInOrder (react-dom.development.js?61bb:583)
    at executeDispatchesAndRelease (react-dom.development.js?61bb:680)
    at executeDispatchesAndReleaseTopLevel (react-dom.development.js?61bb:688)
    at forEachAccumulated (react-dom.development.js?61bb:662)
```

I will submit a PR with a simple fix, adding a check for point in null.

jvansan:
Sorry, just to clarify: This happens when you click on the canvas element, but outside of the drawn plot.

plouc:
Fixed by https://github.com/plouc/nivo/pull/508

PR #508, opened 2019-04-04
Minor bug fix for when clicking in canvas margin

jvansan:
Referencing https://github.com/plouc/nivo/issues/507

An alternative solution would be to check for Truthyness of the point object.
I.E. At line 122
```
if (point) { ...
```

It's unclear to me at this point which is a better solution.


plouc:
Thank you @jvansan!

PR #513, opened 2019-04-11
Make LineDots datum key more unique

sighrobot:
# Problem

React `key` values created by `LineDots` are not unique enough for some `data`.

# Example

```jsx
const data = [{
  data: [
    {
      name: 'Jane',
      x: '2016-07-15'
      y: 'Birthday'
    },
    {
      name: 'Joe',
      x: '2016-07-15'
      y: 'Birthday'
    },
  ],
  id: 'birthday',
}]

const MyLine = ({ data }) => (
  <ResponsiveLine
    axisBottom={{ format: '%Y' }}
    data={data}
    xScale={{
      type: 'time',
      format: '%Y-%m-%d',
      precision: 'day',
    }}
    yScale={{ type: 'point' }}
  />
)
```

If `data` contains two birthdays on the same day (e.g. `2016-07-15`), the `key`s created by the `LineDots` component are not unique and React throws a warning. However these warnings are generated hundreds at a time, and never stop. I have seen upwards of 60,000 identical warnings thrown in matter of minutes.

# Solution

The proposed solution is to add the iteration index from `map` to the `key` value:

```jsx
// before
'birthday.Fri Jul 15 2016 00:00:00 GMT-0400 (Eastern Daylight Time)' 
'birthday.Fri Jul 15 2016 00:00:00 GMT-0400 (Eastern Daylight Time)' 

// after
'0.birthday.Fri Jul 15 2016 00:00:00 GMT-0400 (Eastern Daylight Time)'
'1.birthday.Fri Jul 15 2016 00:00:00 GMT-0400 (Eastern Daylight Time)' 
```

### DIFF
--- a/packages/line/src/LineDots.js
+++ b/packages/line/src/LineDots.js
@@ -40,9 +40,9 @@ const LineDots = ({
             ...acc,
             ...data
                 .filter(datum => datum.position.x !== null && datum.position.y !== null)
-                .map(datum => {
+                .map((datum, datumIdx) => {
                     return {
-                        key: `${id}.${datum.data.x}`,
+                        key: `${datumIdx}.${id}.${datum.data.x}`,
                         x: datum.position.x,
                         y: datum.position.y,
                         datum: datum,


plouc:
I've tried this approach, but it breaks transitions if you modify the input data, for example, if you have the following initial data:
```javascript
['B', 'C', 'D']
// ['0.B', '1.C', '1.D']
```
Then if you prepend another datum:
```javascript
['A', 'B', 'C', 'D']
// ['0.A', '1.B', '2.C', '3.D']
```
The identity for all previous node is broken :/

I think the values should be uniques as there's no way to represent those visually.

plouc:
No, sorry, the problem with this is that it will involve more computing, if you have several identical dates they'll all have to be parsed, computing the scales will become more costly too when using the \'auto\' mode, and it will have no use as it won't be visible on the chart.

If you use some opacity for example, it will behave strangely as the dots will appear stacked and more opaque where you have several identical values.

Also I'm not sure the underlying d3 path generator will take care of deduplication, which means more complex paths, and for some curve interpolations like catmull rom, having several points at the same position might affect the tension/shape of the line.

Just out of curiosity, why don't you want to perform deduplication before passing it to the component?
Maybe I'm missing something, I'm not sure I understand your use case.

plouc:
Sorry to close this, thank you for taking the time to create a PR.

PR #531, opened 2019-05-04
add @nivo/network package

plouc:
network-canvas
network


JeffBaumgardt:
Sorry to jump into a pull request. I can file an issue if it's better. I was noticing that you only account for a single root node. Is it a possibility to have multiple root nodes, each with its own children?

plouc:
@JeffBaumgardt, you can only have one root node for now

PR #540, opened 2019-05-10
Added domain prop to Choropleth and Choropleth Canvas

tg970:
Added domain prop to the choropleths so the color grading can be controlled from the top level.

Went ahead made it a required prop. Let me know if it would be preferable to make it optional.

### DIFF
--- a/packages/geo/src/props.js
+++ b/packages/geo/src/props.js
@@ -119,7 +119,6 @@ const commonChoroplethDefaultProps = {
     label: 'id',
     value: 'value',
     colors: 'PuBuGn',
-    domain: [0, 1000000],
     unknownColor: '#999',
     tooltip: ChoroplethTooltip,
     layers: ['graticule', 'features', 'legends'],
--- a/packages/geo/src/props.js
+++ b/packages/geo/src/props.js
@@ -63,6 +63,7 @@ const commonChoroplethPropTypes = {
     value: PropTypes.oneOfType([PropTypes.string, PropTypes.func]).isRequired,
     valueFormat: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
     colors: quantizeColorScalePropType.isRequired,
+    domain: PropTypes.arrayOf(PropTypes.number).isRequired,
     unknownColor: PropTypes.string.isRequired,
     layers: PropTypes.arrayOf(
         PropTypes.oneOfType([PropTypes.oneOf(['graticule', 'features', 'legends']), PropTypes.func])
@@ -118,6 +119,7 @@ const commonChoroplethDefaultProps = {
     label: 'id',
     value: 'value',
     colors: 'PuBuGn',
+    domain: [0, 1000000],
     unknownColor: '#999',
     tooltip: ChoroplethTooltip,
     layers: ['graticule', 'features', 'legends'],
--- a/website/src/data/components/choropleth/props.js
+++ b/website/src/data/components/choropleth/props.js
@@ -70,9 +70,7 @@ const props = [
         help: 'Defines uppper and lower bounds of color shading',
         type: 'number[]',
         required: true,
-        defaultValue: ChoroplethDefaultProps.domain,
-        controlType: 'domainColors',
-        group: 'Style',
+        group: 'Base',
     },
     {
         key: 'unknownColor',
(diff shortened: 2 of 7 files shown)


plouc:
Nice! It's perfect as I'll need this for the upcoming State of CSS survey report ;)

packages/geo/src/props.js
```diff
@@ -118,6 +119,7 @@ const commonChoroplethDefaultProps = {
     label: 'id',
     value: 'value',
     colors: 'PuBuGn',
+    domain: [0, 1000000],
```
plouc: This is quite an arbitrary domain, I think we can remove this and let users add their own,  otherwise I'm afraid that people won't understand where does the steps come from.

website/src/data/components/choropleth/props.js
```diff
@@ -65,6 +65,15 @@ const props = [
         controlType: 'quantizeColors',
         group: 'Style',
     },
+    {
+        key: 'domain',
+        help: 'Defines uppper and lower bounds of color shading',
+        type: 'number[]',
+        required: true,
+        defaultValue: ChoroplethDefaultProps.domain,
+        controlType: 'domainColors',
```
plouc: There is not such control unfortunately, you can remove this key, I'll add it later if I add one

website/src/data/components/choropleth/props.js
```diff
@@ -65,6 +65,15 @@ const props = [
         controlType: 'quantizeColors',
         group: 'Style',
     },
+    {
+        key: 'domain',
+        help: 'Defines uppper and lower bounds of color shading',
+        type: 'number[]',
+        required: true,
+        defaultValue: ChoroplethDefaultProps.domain,
```
plouc: We can remove default domain

website/src/data/components/choropleth/props.js
```diff
@@ -65,6 +65,15 @@ const props = [
...
+        controlType: 'domainColors',
+        group: 'Style',
```
plouc: Maybe it's more something for the `Base` group as it relates to the data.

website/src/data/components/choropleth/props.js
tg970: Removed.

website/src/data/components/choropleth/props.js
tg970: Done.

website/src/pages/choropleth/index.js
```diff
@@ -29,6 +29,7 @@ const initialProperties = {
     },
 
     colors: 'nivo',
+    domain: [0,1000000],
```
plouc: It should be added to the canvas implementation too

website/src/data/components/choropleth/props.js
```diff
@@ -65,6 +65,13 @@ const props = [
         controlType: 'quantizeColors',
         group: 'Style',
     },
+    {
```
plouc: Can you move this to next to other `Base` props please?

website/src/data/components/choropleth/props.js
tg970: Changed to Base group.

website/src/data/components/choropleth/props.js
tg970: You got it! Should we add a description since the other base props all have a description?

plouc: Yes! it could help users understand how does this value affect the chart

website/src/pages/choropleth/index.js
tg970: Done!

website/src/data/components/choropleth/props.js
tg970: Ok moved it next to the other base props and added a description.

plouc:
@tg970 the build is failing, you can easily fix this by running `make fmt`.

tg970:
Thanks for all of your help @plouc, I really appreciate it! It's exciting to be able to make a contribution.

plouc:
Well, Thank you for your contribution ;)

plouc:
@tg970, you can use `@nivo/geo@0.57.2`, the website has been updated too.

Issue #123, opened 2018-01-04
Allow custom symbol shapes for ScatterPlot


Firenze11:
The `@nivo/line` package allows custom `dotSymbol` shape ([link](http://nivo.rocks/storybook/?knob-curve=monotoneX&knob-stacked=true&selectedKind=Line&selectedStory=custom%20dot%20symbol&full=0&down=1&left=1&panelRight=0&downPanel=storybooks%2Fstorybook-addon-knobs)). `ScatterPlot` should allow the same functionality too.

PR #549, opened 2019-05-16
symbol function

Diogo-Ferreira:
Hi :wave: @ch-apptitude we use Nivo's `ScatterPlot` and we need to provide a custom item shape like explained in this issue https://github.com/plouc/nivo/issues/123 .

This PR is the implementation of that. 

Therefore you can pass a function that returns an SVG to the `symbol` prop. You can also do conditional shape as the point data is also passed:

```
const PointShape = ({
  x,
  y,
  size,
  color,
  point,
  fill,
  ...etc
}) => {
  const { data: { formulaType } } = point;

  if (formulaType === 'PREDEFINED') {
    return (
      <g height="16" width="16" transform={`translate(${x}, ${y})`}>
        <path d="M8 0 L16 16 L0 16 Z" fill={color} {...etc} />
      </g>
    );
  }
  return (
    <circle
      cx={x}
      cy={y}
      r={size / 2}
      fill={fill}
      {...etc}
    />
  );
};
```

Let me know what you think about it.

### DIFF
--- a/packages/scatterplot/index.d.ts
+++ b/packages/scatterplot/index.d.ts
@@ -24,17 +24,17 @@ declare module '@nivo/scatterplot' {
         }
     }
 
-    export type SymbolSvg = (
-        x: number,
-        y: number,
-        size: number,
-        color: string,
-        onMouseEnter: ScatterPlotMouseHandler,
-        onMouseMove: ScatterPlotMouseHandler,
-        onMouseLeave: ScatterPlotMouseHandler,
-        onClick: ScatterPlotMouseHandler,
+    export type SymbolSvg = (arg: {
+        x: number
+        y: number
+        size: number
+        color: string
+        onMouseEnter?: ScatterPlotMouseHandler
+        onMouseMove?: ScatterPlotMouseHandler
+        onMouseLeave?: ScatterPlotMouseHandler
+        onClick?: ScatterPlotMouseHandler
         point: Point
-    ) => React.ReactNode
+    }) => React.ReactNode
 
     export type SymbolCanvas = (
         ctx: CanvasRenderingContext2D,
--- a/packages/scatterplot/index.d.ts
+++ b/packages/scatterplot/index.d.ts
@@ -39,8 +39,8 @@ declare module '@nivo/scatterplot' {
     export type SymbolCanvas = (
         ctx: CanvasRenderingContext2D,
         point: Point,
-        getSymbolSize: (data: Point.data) => void,
-        getColor: (data: Point.data) => void
+        getSymbolSize: ScatterPlotSizeGetter,
+        getColor: (data: ScatterPlotDatum) => void
     ) => void
 
     export type ScatterPlotMouseHandler = (
--- a/packages/scatterplot/index.d.ts
+++ b/packages/scatterplot/index.d.ts
@@ -16,6 +16,33 @@ declare module '@nivo/scatterplot' {
         color: string
     }
 
+    interface Point {
+        id: string
+        data: {
+            x: number
+            y: number
+        }
+    }
+
+    export type SymbolSvg = (
+        x: number,
+        y: number,
+        size: number,
+        color: string,
+        onMouseEnter: ScatterPlotMouseHandler,
+        onMouseMove: ScatterPlotMouseHandler,
+        onMouseLeave: ScatterPlotMouseHandler,
+        onClick: ScatterPlotMouseHandler,
+        point: Point
+    ) => React.ReactNode
+
+    export type SymbolCanvas = (
+        ctx: CanvasRenderingContext2D,
+        point: Point,
+        getSymbolSize: (data: Point) => void,
+        getColor: (data: Point) => void
+    ) => void
+
     export type ScatterPlotMouseHandler = (
         data: ScatterPlotDatum,
         event: React.MouseEvent<any>
@@ -24,13 +51,7 @@ declare module '@nivo/scatterplot' {
     export type ScatterPlotSizeGetter = (data: ScatterPlotDatum) => number
 
     export interface ScatterPlotProps {
-        data: Array<{
-            id: string
-            data: Array<{
-                x: number
-                y: number
-            }>
-        }>
+        data: Point[]
 
         xScale?: Scale
         yScale?: Scale
@@ -50,6 +71,7 @@ declare module '@nivo/scatterplot' {
 
         symbolSize?: number | ScatterPlotSizeGetter
         symbolShape?: 'circle' | 'square'
+        symbol?: SymbolCanvas | SymbolSvg | React.MemoExoticComponent<any>
 
         isInteractive?: boolean
         useMesh?: boolean
--- a/packages/scatterplot/src/ScatterPlotCanvas.js
+++ b/packages/scatterplot/src/ScatterPlotCanvas.js
@@ -200,7 +200,7 @@ class ScatterPlotCanvas extends Component {
 
         points.forEach(point => {
             if (symbol) {
-                symbol(this.ctx, point);
+                symbol(this.ctx, point)
                 this.ctx.fill()
             } else {
                 this.ctx.beginPath()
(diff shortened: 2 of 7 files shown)


packages/scatterplot/src/ScatterPlotItem.js
```diff
@@ -21,18 +21,33 @@ const ScatterPlotItem = ({
...
+            size,
+            color,
+            onMouseEnter,
+            onMouseMove,
+            onMouseLeave,
+            onClick,
+            point,
+        })
+    ) : (
+        <circle
```
plouc: This could be the default symbol instead of having to repeat all the props

packages/scatterplot/src/props.js
```diff
@@ -60,6 +60,8 @@ export const ScatterPlotPropTypes = {
     symbolSize: PropTypes.oneOfType([PropTypes.func, PropTypes.number]).isRequired,
     symbolShape: PropTypes.oneOfType([PropTypes.oneOf(['circle', 'square'])]).isRequired,
 
+    symbol: PropTypes.func,
```
plouc: What about canvas implementation? Also, if we use something like `memo()` to wrap the symbol component (which should be the case IMO), you'll have an object not a function so we should support both types

plouc: The TypeScript definition should be updated too.

packages/scatterplot/src/ScatterPlotItem.js
```diff
@@ -21,18 +21,33 @@ const ScatterPlotItem = ({
...
-        onMouseEnter={onMouseEnter}
-        onMouseMove={onMouseMove}
-        onMouseLeave={onMouseLeave}
-        onClick={onClick}
-    />
-)
+    symbol,
+    point,
+}) =>
+    symbol ? (
```
plouc: I guess we could easily add a test for this searching for a matching number of custom symbols.

plouc:
@Diogo-Ferreira, thank you for this, it will be useful

packages/scatterplot/src/ScatterPlotItem.js
Diogo-Ferreira: I put the default symbol as default prop, is this what you wanted ? let me know.

packages/scatterplot/src/props.js
Diogo-Ferreira: Both function and object are allowed, will do the TS def soon.

packages/scatterplot/src/ScatterPlotCanvas.js
```diff
@@ -198,10 +199,15 @@ class ScatterPlotCanvas extends Component {
         })
 
         points.forEach(point => {
-            this.ctx.beginPath()
-            this.ctx.arc(point.x, point.y, getSymbolSize(point.data) / 2, 0, 2 * Math.PI)
-            this.ctx.fillStyle = getColor(point.data)
-            this.ctx.fill()
+            if (symbol) {
```
Diogo-Ferreira: For the canvas implementation I propose something like this, let me know what you think.

plouc: We could use the same approach as for the SVG implementation, setting a default symbol function and making it required.
Also `this.ctx.fill()` should be removed as it's part of the symbol rendering function.

packages/scatterplot/src/ScatterPlotItem.js
```diff
@@ -34,6 +49,23 @@ const ScatterPlotItem = ({
...
+    onMouseEnter: PropTypes.func,
+    onMouseMove: PropTypes.func,
+    onMouseLeave: PropTypes.func,
+    onClick: PropTypes.func,
+}
+
+DefaultSymbol.propTypes = commonProps
+
+ScatterPlotItem.defaultProps = {
+    symbol: DefaultSymbol,
```
plouc: Maybe we could set this directly in the `ScatterPlot` default props.

packages/scatterplot/stories/ScatterPlotCanvas.stories.js
```diff
@@ -125,6 +125,11 @@ const commonProps = {
     height: 500,
     margin: { top: 20, right: 20, bottom: 80, left: 80 },
     symbolSize: 10,
+    symbol: (ctx, point) => {
```
plouc: It could be useful to have a demo of custom symbol, but probably not by default IMO

packages/scatterplot/index.d.ts
```diff
@@ -16,6 +16,33 @@ declare module '@nivo/scatterplot' {
...
 
+    interface Point {
+        id: string
+        data: {
+            x: number
+            y: number
+        }
+    }
+
+    export type SymbolSvg = (
```
plouc: The properties should be passed as an object, not as several arguments.

packages/scatterplot/stories/ScatterPlotCanvas.stories.js
Diogo-Ferreira: True. I removed it.

Do you want me to create another story or to add it to an existing one ?

packages/scatterplot/index.d.ts
```diff
@@ -24,13 +51,7 @@ declare module '@nivo/scatterplot' {
...
 
     export interface ScatterPlotProps {
-        data: Array<{
-            id: string
-            data: Array<{
-                x: number
-                y: number
-            }>
-        }>
+        data: Point[]
```
plouc: The `data` property is an array of series, not an array of points

packages/scatterplot/index.d.ts
```diff
@@ -16,6 +16,33 @@ declare module '@nivo/scatterplot' {
...
+        size: number
+        color: string
+        onMouseEnter?: ScatterPlotMouseHandler
+        onMouseMove?: ScatterPlotMouseHandler
+        onMouseLeave?: ScatterPlotMouseHandler
+        onClick?: ScatterPlotMouseHandler
+        point: Point
+    }) => React.ReactNode
+
+    export type SymbolCanvas = (
```
plouc: For other components, when allowing to customize canvas rendering, I used this approach:

```typescript
export type SymbolCanvas = (ctx: CanvasRenderingContext2D, {
    // ...props
}) => void
```

packages/scatterplot/src/ScatterPlotItem.js
```diff
@@ -34,6 +58,19 @@ const ScatterPlotItem = ({
     />
 )
 
+const commonProps = {
```
plouc: you can set this directly on `DefaultSymbol.propTypes`, I used this approach when we have several implementation of the same component (canvas, html, …)

packages/scatterplot/src/enhance.js
```diff
@@ -55,13 +54,23 @@ const commonEnhancers = [
...
         pure,
         setDisplayName('ScatterPlot')
     )(Component)
 
 export const enhanceCanvas = Component =>
     compose(
+        defaultProps({
+            ...ScatterPlotDefaultProps,
+            symbol: (ctx, point, getSymbolSize, getColor) => {
+                this.ctx.beginPath()
```
plouc: You can remove `this` ;)

packages/scatterplot/stories/ScatterPlotCanvas.stories.js
plouc: That would be nice!

plouc:
@Diogo-Ferreira, I'm really sorry but I had to update the `@nivo/scatterplot` package for a project I'm working on, and I also migrated the package to the new architecture (using hooks).
I've changed a lot of things :/ as you already took some time to implement this feature, I've also implemented it while refactoring.
Pleas see https://github.com/plouc/nivo/pull/584

plouc:
This feature is available in `@nivo/scatterplot@0.59.0`, the website has been updated accordingly, the new prop is named `renderNode`. Thank you for your work and sorry again.

Diogo-Ferreira:
@plouc No problem

PR #568, opened 2019-05-28
Adding a tooltip typing to the LineProps interface (Line/ResponsiveLine)

iliyanyotov:
Adding a missing typing for the tooltip on the Line and ResponsiveLine components.

[TS error of the missing property definition](https://i.imgur.com/EuMHjYg.png)

### DIFF
--- a/packages/line/index.d.ts
+++ b/packages/line/index.d.ts
@@ -89,8 +89,8 @@ declare module '@nivo/line' {
             index: number
             serieId: string | number
             serieColor: string
-            x: number
-            y: number
+            x: number | string | Date
+            y: number | string | Date
             color: string
             borderColor: string
             data: {
@@ -98,7 +98,7 @@ declare module '@nivo/line' {
                 x: string | number
                 y: number
                 yStacked: number
-                xFormatted: string
+                xFormatted: string | number
                 yFormatted: string | number
             }
         }
--- a/packages/line/index.d.ts
+++ b/packages/line/index.d.ts
@@ -83,6 +83,8 @@ declare module '@nivo/line' {
     export type LineCustomLayer = (props: LineCustomLayerProps) => React.ReactNode
     export type Layer = LineLayerType | LineCustomLayer
 
+    export type TooltipProp = React.StatelessComponent<LineComputedSerieData[]>
+
     export interface LineProps {
         data: LineSerieData[]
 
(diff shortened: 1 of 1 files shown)


JakeSidSmith:
I could really do with this fix, however I'm not sure if the proposed type is correct. I'm a little unsure of the changes that have been made to the tooltip prop.

It seems the type was (accidentally?) removed here: https://github.com/plouc/nivo/commit/d56881b82f5fa480830f3dcb4d22645337f9008d#diff-4fed95f31b1c07652a6a8ca3acc9885eL133

I would not recommend using `StatelessComponent` as the type also, as this is deprecated, and adds some additional types that are not necessary. I'd stick with a custom arrow function `() => {}`.

packages/line/index.d.ts
```diff
@@ -83,6 +83,8 @@ declare module '@nivo/line' {
     export type LineCustomLayer = (props: LineCustomLayerProps) => React.ReactNode
     export type Layer = LineLayerType | LineCustomLayer
 
+    export type TooltipProp = React.StatelessComponent<LineComputedSerieData[]>
```
wyze: I would change this to `React.FC<LineTooltipProps>`.

plouc:
@wyze is right, the previous tooltip is now `sliceTooltip`, and the other one: `tooltip` is used for single points.

iliyanyotov:
@plouc @wyze Thanks for the comments, the pull request is updated.

packages/line/index.d.ts
```diff
@@ -83,6 +83,29 @@ declare module '@nivo/line' {
...
+            index: number
+            serieId: string | number
+            serieColor: string
+            x: number
+            y: number
+            color: string
+            borderColor: string
+            data: {
+                color: string
+                x: string | number
```
plouc: x and y values can be numbers, strings or dates, so the type should be `number | string | Date`.

packages/line/index.d.ts
```diff
@@ -83,6 +83,29 @@ declare module '@nivo/line' {
...
+                y: number
+                yStacked: number
+                xFormatted: string
```
plouc: `xFormatted` can also be a number.

iliyanyotov: Updated.

packages/line/index.d.ts
iliyanyotov: Updated.

karenqinn:
@plouc Any other fixes I can help with in order to make this PR acceptable? :)

plouc:
Thank you for your contributions.

PR #574, opened 2019-06-03
Respect useUTC option on x/y scale property

wyze:
If I am to disable UTC for the scale, then I would expect the tick values to also not use UTC. It looks like from the code we always use the UTC helpers when generating tick values. I'm not sure if this is the best way to pass that option from the `scale` package to the `axes` package. So feel free to tell me to do it a different way.

Here is my first CodeSandbox with my simple test case, and you can see the values are incorrect on the ticks. Also the first tick is missing a value. https://codesandbox.io/s/jyz1h

I used `patch-package` to apply the code changes in the PR here, but those weren't respected in CodeSandbox. I have a deployed version here (https://csb-jyz1h-crmzrjqkcn.now.sh/) where you can see the tick values are correct, including the first point.

To view the source for the above, you can go here: https://zeit.co/deployments/csb-jyz1h-crmzrjqkcn.now.sh/source to see that the code is the same as the first CodeSandbox, but the patches are applied.

### DIFF
--- a/packages/axes/src/compute.js
+++ b/packages/axes/src/compute.js
@@ -102,7 +102,7 @@ export const getScaleTicks = (scale, spec) => {
                 // UTC is used as it's more predictible
                 // however local time could be used too
                 // let's see how it fits users' requirements
-                const timeType = timeByType[matches[2]][1]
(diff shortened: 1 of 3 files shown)


plouc:
Thank you for your contribution @wyze!

PR #576, opened 2019-06-03
feat(line): add onClick for slices

warmeta:
Add onClick event for slices in Line graph. Actually onClick handler doesn't work with slice tooltip, with this functionallity onClick return the slices object.


plouc:
What about Canvas implementation?
You forgot to update TypeScript definitions accordingly
I think it's best to separate handlers for points and those for slices.

One more thing, please do not contact me on my professional email regarding nivo, it's a personal project and if I don't review your PR that's simply because I don't have the time to do so.

PR #578, opened 2019-06-04
Legend theme fill color always being overwritten

jefferymills:
The `textColor` prop has a default value of black which means the `fill` in the style property was always overwriting the `fill` property coming from the theme.

### DIFF
--- a/packages/legends/src/svg/LegendSvgItem.js
+++ b/packages/legends/src/svg/LegendSvgItem.js
@@ -178,7 +178,7 @@ LegendSvgItem.defaultProps = {
     direction: 'left-to-right',
     justify: false,
 
-    textColor: 'black',
+    textColor: '#333333',
     background: 'transparent',
     opacity: 1,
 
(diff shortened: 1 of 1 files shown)


wyze:
@plouc What value should take priority here, the theme fill color or the defaultProp from the legend item?

Also, @jefferymills, looks like there are a few tests that broke with your change.

jefferymills:
Thanks @wyze, I wasn't able to get things running locally because the `make init` fails so I haven't been able to see why my change breaks those tests. I made the change directly in GitHub. Should I just make an issue? Basically, right now updating the legend text color with the theme isn't working.

plouc:
@jefferymills, how does `make init` fails?

jefferymills:
@plouc Sorry for the delay. I've been out of town. I will share the output I get when running `make init`. I'm sure it's something wrong with my setup but I'm not as experienced with build systems and most the output is over my head :)

https://docs.google.com/document/d/1d9AmyO81RTWGtdKcbzJTO63Hy2_lZ-dmKISfzOunW3Q/edit?usp=sharing

plouc:
This issue has been fixed in master and the fix will be available soon in the upcoming release, I'm very sorry for the late reply and thank you for your contribution!

PR #587, opened 2019-06-10
feat(radar): Tooltip format function gets key

tloszabno:
I needed functionality to format tooltips regards key in radar chart.


plouc:
Thank you for your contribution @tloszabno!

PR #454, opened 2019-02-08
Shows labels when bar height/width is zero

meganhng:
Fixes: #448 

The bar values with no height wouldn't have their labels displayed because the data wouldn't be pushed onto the graph.

PR #591, opened 2019-06-11
Improve with a props to show/hide bar when it's zero

Amitom:
Improvement of https://github.com/plouc/nivo/pull/454, with resolved conflicts (I don't assume the potential next version released with this PR).

I added a prop to decide if zero bar is "shown" or not, by default it's the same behavior than currently.

### DIFF
--- a/packages/bar/src/Bar.js
+++ b/packages/bar/src/Bar.js
@@ -64,6 +64,7 @@ const Bar = props => {
         outerHeight,
         padding,
         innerPadding,
+        hideZeroValues,
 
         axisTop,
         axisRight,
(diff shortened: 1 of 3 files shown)


cookkate:
We could really use this PR, if #454 isn't tenable.

Stephen-ONeil:
Oh, +1 this, actually

Mihier-Roy:
+1 this. Could we have this released anytime soon?

akornmeier:
This update would be a great addition to the chart component. Would like to see the same ability to show the `0 bar` when the graph is in grouped mode too. That would necessitate a change to the `./compute/grouped.js` file.

Blerd:
+1 would definitely love this out :)

justinro-underscore:
Any updates on this?

PR #607, opened 2019-06-26
added gridXValues and gridYValues to line typings

vichugunov:
Added missing typings. Done accordingly to:
https://github.com/plouc/nivo/blob/master/packages/axes/src/components/Grid.js#L51
https://github.com/plouc/nivo/blob/master/packages/line/src/Line.js#L139

### DIFF
--- a/packages/line/index.d.ts
+++ b/packages/line/index.d.ts
@@ -114,7 +114,9 @@ declare module '@nivo/line' {
         axisLeft?: AxisProps | null
 
         enableGridX?: boolean
+        gridXValues?: number | number[] | string[] | Date[]
         enableGridY?: boolean
+        gridYValues?: number | number[] | string[] | Date[]
 
         enablePoints?: boolean
         pointSize?: number


packages/line/index.d.ts
```diff
@@ -114,7 +114,9 @@ declare module '@nivo/line' {
         axisLeft?: AxisProps | null
 
         enableGridX?: boolean
+        gridXValues?: number | number[] | string[] | Date[]
```
plouc: As the corresponding component is in the `axes` package, I think we should put this in that package.

vichugunov: Yes, you're right. I've done that deliberately for the couple of reasons:
1. When you're using `Line` and you want to set it there, typescript compile shows the error. That's why these props should be there, because it has to be set for the component firstly (but later it'll be used by another module within `axes` package).

2. If I'm not mistaken, these values are used by `Grid` component itself from `axes` package, but I don't see any typings for `Grid` within that package, only `Axis` and `AxisProps`, that's why I haven't added anything new there. Maybe `Grid` hasn't exposed deliberately, I don't know will be used or not

Should I  create typing for `Grid` from scratch? (I haven't seen anything else that will use this `grid(X|Y)Values` from line within `axes` package)

plouc: You're right, the `Grid` component being more an _internal_ component, it's not necessary to add this definition in the `axes` package.

PR #612, opened 2019-07-02
Provide theme to renderTick function

### LINKED ISSUE #579: [Storybook] Bar Custom Axis Ticks doesn't work
**Describe/explain the bug**
Hey, I've just noticed that one of your storybook stories throws an exception.

**To Reproduce**
1. Go to Storybook
2. Expand `Bar`
3. Click on `custom axis ticks`

See url: https://nivo.rocks/storybook/?path=/story/bar--custom-axis-ticks

**Expected behavior**
No error 🤷‍♂ 

**Screenshots**
image

 - OS: MacOS
 - Browser Chrome 
 - Version (74.0.3729.169 (Official Build) (64-bit))

wyze:
Close #579

### DIFF
--- a/packages/axes/src/components/Axis.js
+++ b/packages/axes/src/components/Axis.js
@@ -119,6 +119,7 @@ const Axis = ({
                         rotate: tickRotation,
                         textBaseline,
                         textAnchor: textAlign,
+                        theme,
                         ...tick,
                         ...(onClick ? { onClick } : {}),
                     })
@@ -161,6 +162,7 @@ const Axis = ({
                                         format: formatValue,
                                         textBaseline,
                                         textAnchor: textAlign,
+                                        theme,
                                         ...tick,
                                         ...style,
                                         ...(onClick ? { onClick } : {}),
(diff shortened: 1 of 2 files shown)


plouc:
@wyze, in fact you do not need to pass down the theme property, you can use the dedicated hook to access the theme, `import { useTheme } from '@nivo/core` and then `const theme = useTheme()`.

wyze:
The bug is the `renderTick` function doesn't receive a `theme` property. The `Axis` component already is using the `useTheme` hook, so it will provide that to the `renderTick` function, which defaults to `AxisTick`, when it isn't provided by the user.

plouc:
You should use `useTheme` in the `renderTick` function.

wyze:
So, you're saying to revert the above, and change the story here https://github.com/plouc/nivo/blob/0a90609bd4c7f9fb29cd95b4879a857736dbb680/packages/bar/stories/bar.stories.js#L205 to import the `useTheme` hook and use that?

plouc:
@wyze, yes, that's the correct approach, however I've just realized that `renderTick` isn't treated as a React component, I'm currently fixing this.

plouc:
I've created a PR => https://github.com/plouc/nivo/pull/635

PR #634, opened 2019-07-11
Fix domain not in index.d.ts file, even though it is required

rccoe:
I'm getting an error when running ResponsiveChloropleth when using it without a domain prop. Typescript should have caught this, since it is a required prop, but it didn't because it's not in the index.d.ts file.


plouc:
Thank you @rccoe!

PR #653, opened 2019-07-29
Fix PropType for Line markers when using dates in the X axis.

nfvs:
When using a Line chart with time in the X axis, passing a date as a marker's x value works but raises a PropTypes warning. This change allows dates in the markers' values.


plouc:
Thank you for your contribution @nfvs!

PR #656, opened 2019-08-02
feat(line): pass the state to a custom layer

proAlexandr:
Hi!
The changes pass state variables to a custom layer in the Line component. For example, it allows me to replace crosshair with a custom implementation.


wyze:
Thanks for this @proAlexandr! Will go out in the next release.

PR #657, opened 2019-08-05
Update tooltip TypeScript types.

nfvs:
- removed theme= from Basic/TableTooltip
- define Chip

### DIFF
--- a/packages/tooltip/index.d.ts
+++ b/packages/tooltip/index.d.ts
@@ -17,15 +17,21 @@ declare module '@nivo/tooltip' {
         color: string
         format?: (value: number | string) => number | string
         renderContent?: () => React.ReactNode
-        theme: Pick<Theme, 'tooltip'>
     }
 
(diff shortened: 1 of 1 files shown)


nfvs:
@wyze maybe not the best forum to ask this, but what's the best way to use such a fork in a project? Importing the git repo in `package.json` doesn't seem work.

wyze:
We have it set in `tsconfig.json` like so:

```json
{
  "files": [
    "@nivo/core.d.ts",
    "@nivo/sankey.d.ts",
    "@nivo/scales.d.ts",
    "@nivo/tooltip.d.ts",
    "global.d.ts"
  ]
}
```

And then directories are relative to that. So you would have a folder called `@nivo`, then in that folder have files that correspond to above. Example, `@nivo/tooltip.d.ts`:

```typescript
declare module '@nivo/tooltip' {
  export interface ChipProps {
    size?: number
    color: string
    style?: Partial<React.CSSProperties>
  }

  export class Chip extends React.Component<ChipProps> {}
}
```

PR #659, opened 2019-08-08
Fixed no fontSize definition in Themes and Added Legends property in geo module

tmahmood:
I have added the fontSize definition and story to make sure it's working as expected.

Added the legend property definition to Geo module.

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -35,6 +35,7 @@ declare module '@nivo/core' {
 
     export type Theme = Partial<{
         background: string
+        fontSize: string | 0
         axis: Partial<{
             domain: Partial<{
                 line: Partial<React.CSSProperties>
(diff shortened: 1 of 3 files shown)


packages/geo/index.d.ts
```diff
@@ -116,6 +117,7 @@ declare module '@nivo/geo' {
         onMouseMove?: ChoroplethEventHandler
         onMouseLeave?: ChoroplethEventHandler
         onClick?: ChoroplethEventHandler
+        legends: Array<{ dataFrom: 'indexes' | 'keys' } & LegendProps>
```
plouc, line 120: This package doesn't support `dataFrom`, the type should be `legends: LegendProps[]`

packages/core/index.d.ts
```diff
@@ -35,6 +35,7 @@ declare module '@nivo/core' {
 
     export type Theme = Partial<{
         background: string
+        fontSize: string | 0
```
plouc, line 38: Actually there are other props missing from this definition:

```typescript
    fontFamily: string
    textColor: string
```

It would be great if you could add them too!

PR #664, opened 2019-08-10
fix(tooltip): fix tooltip overflow

proAlexandr:
Hi!
The purpose of the PR is to avoid that tooltip is not visible when it is out of the browser window.

GIF examples:
<details>
<summary>
Before:
</summary>

Aug-10-2019 12-13-16
</details>

<details>
<summary>
After:
</summary>

Aug-10-2019 12-09-22
</details>

### DIFF
--- a/packages/core/src/hocs/withContainer.js
+++ b/packages/core/src/hocs/withContainer.js
@@ -50,7 +50,11 @@ const Container = ({
                         <div style={containerStyle} ref={container}>
                             {children}
                             {isTooltipVisible && (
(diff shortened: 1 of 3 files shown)


rpnna:
hey @proAlexandr , looking at the image you posted above, which is the correct approach to achieve the same effect? I need to have two different dots (one per line) that move on mouse hover in a ResponsiveLine chart.
Thanks

wyze:
Hi @proAlexandr, thanks for the contribution. It looks like there was a fix done already that shifts the position of the tooltip depending on the position of the cursor. It displays it either on the right or the left.

PR #670, opened 2019-08-21
Adding a ScatterPlotScale interface

thomasJang:
Adding a wrong typing for the ScatterPlot's scale prop

image

So, I Added ScatterPlotScale interface and updated ScatterPlotProps interface.

PS)  It's an honor to be able to use this wonderful open source.


wyze:
Thanks for the contribution @thomasJang! It looks like this was a duplicate, and it was farther back so #935 was merged before this and they fix the same thing.

PR #681, opened 2019-09-06
Add missing pointSymbol prop to line typings

latema:
The `pointSymbol` is missing in the line typings. Fixes the following error:

pointSymbol


plouc:
Thank you for your contribution @latema!

Issue #525, opened 2019-04-21
Button to copy code/data in clipboard on all the examples


jonattfin:
I think that a copy button on the code/data would be useful for rapid taking that code/data & putting it in your own editor, etc. See attached picture.
![Screenshot_2019-04-21 Bar-1](https://user-images.githubusercontent.com/3768739/56473632-a6de0f00-6476-11e9-8e6e-5108f8f764ec.png)

plouc:
You're right, in fact it's already on my todo ;)

Larissagilliane:
I can help and work on that issue

PR #683, opened 2019-09-11
Create copy button with pure js

Larissagilliane:
This pull request adds a button to copy code/data in clipboard on all the examples, as asked at issue #525 

For this solution i've used pure JS, but my suggestion is refinying the code using a dependencie like "react-clipboard.js" as you can see in the code below, is no needed any method, i just pass the content to the imported component:

`<Clipboard data-clipboard-text={copyContent} button-title="Copied!">
                Copy
   </Clipboard>`

let me know if i can add this dependency so i can refactor this code. 

@plouc

### DIFF
--- a/website/src/components/components/ActionsLoggerLog.js
+++ b/website/src/components/components/ActionsLoggerLog.js
@@ -22,7 +22,7 @@ const ActionHeader = styled.div`
     cursor: pointer;
 
     &:hover {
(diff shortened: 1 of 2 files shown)


wyze:
@Larissagilliane How about we use something like [wsmd/use-clipboard-copy](https://github.com/wsmd/use-clipboard-copy) instead?

PR #687, opened 2019-09-17
feat: add id to symbolShape

andresmechali:
This simple modification adds access to the legend id in the symbolShape property. This is useful because it enables the possibility of rendering a custom icon next to each legend, based on the id. It is then possible to implement a toggling button on each legend, which can show and hide part of the data from the graph.


plouc:
@andresmechali, good idea, thank you for your contribution!

PR #690, opened 2019-09-20
Add `useUTC` to types

eknowlton:
After some messing around with Nivo and line charts, I had a issue where I discovered the dates I was passing in were being converted to the clients timezone, even though I wasn't supplying a timezone with the date and these dates I was passing in were actually already correct. To get around this we discovered a `useUTC` option which we set to `false`. 

Although it seems like its missing from the types.

### DIFF
--- a/packages/scales/index.d.ts
+++ b/packages/scales/index.d.ts
@@ -23,6 +23,7 @@ declare module '@nivo/scales' {
         type: 'time'
         format?: string
         precision?: 'millisecond' | 'second' | 'minute' | 'hour' | 'month' | 'year' | 'day'
+        useUTC: boolean
     }
 
     export interface LogScale {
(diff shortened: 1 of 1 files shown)


packages/scales/index.d.ts
```diff
@@ -23,6 +23,7 @@ declare module '@nivo/scales' {
         type: 'time'
         format?: string
         precision?: 'millisecond' | 'second' | 'minute' | 'hour' | 'month' | 'year' | 'day'
+        useUTC: boolean
```
plouc: Sorry for the late review :/ This property is optional, it should be `useUTC?: boolean`.

plouc:
Thank you @eknowlton!

eknowlton:
Thank you for such a awesome library!

PR #700, opened 2019-10-05
Added export classes for canvas calendar

SourceCipher:
I have noticed that these classes were never exported and I could not import them in typescript. After adding these 2 lines it was working very well. We had big data to display and with the svg canvas it was very laggy. As soon as I exported these 2 classes, calendar was displayed in canvas and the response speed was amazing!


plouc:
Thank you for your contribution @SourceCipher!

PR #703, opened 2019-10-09
Added custom `colorScale`

### LINKED ISSUE #693: Dynamic Colors in Calendar (Custom `colorScale` fn)
I am building a calendar where I have my own `colorScale` function to get `colors` for each `value`. The bands I am working are not linear so the default implementation (which is a static `colorScale` function based on `domain`) does not fit the use case 

I would like to be able to override the `colorScale` function with something like 
```
/**
 * @params {*} value  = the actual value at the box 
 * @returns {*} color = color in hex formatt e.g. #121212
(value) => myCustomColorScale(value)

```

I have failed to find my way around this. 

**Additional context**
N/A 

The code that needs to be updated is here:
https://github.com/plouc/nivo/blob/dbd44e270b5643acfa65319d09ea16a7c29e63c0/packages/calendar/src/enhance.js#L24-L35

I will be happy to open a PR, might need some assistance with changing the code structure though.

jashanbhullar:
### NEED HELP 

I needed a custom `color` on the calendar heatmap based on my scale so I added this functionality. I have made it successfully work in my codebase. 
I need more clarity on what needs to be changed and if there's a better way to do this rather than what I have currently done. 

It also needs an update in the documentation.

Resolves #693 

image

### DIFF
--- a/packages/calendar/src/enhance.js
+++ b/packages/calendar/src/enhance.js
@@ -22,15 +22,16 @@ const commonEnhancers = [
     withTheme(),
     withDimensions(),
     withPropsOnChange(
-        ['data', 'minValue', 'maxValue', 'colors'],
-        ({ data, minValue, maxValue, colors }) => {
+        ['data', 'minValue', 'maxValue', 'colors', 'colorScale'],
+        ({ data, minValue, maxValue, colors, colorScale }) => {
+            if (colorScale) return { colorScale }
             const domain = computeDomain(data, minValue, maxValue)
 
-            const colorScale = scaleQuantize()
+            const defaultColorScale = scaleQuantize()
                 .domain(domain)
                 .range(colors)
 
-            return { colorScale }
+            return { colorScale: defaultColorScale }
         }
     ),
     withPropsOnChange(


plouc:
That's a useful feature, thank you for your contribution @jashanbhullar!

jashanbhullar:
@plouc when will this show up on the npm package?

jashanbhullar:
@plouc still awaiting the update on the version number.

gdevrs:
Is there any example how this works? Or what the custom colorScale function is supposed to look like? I cannot figure it out.

koders:
@gdevrs I figured it out, you have to have a function, which returns a color, and also have a tick array under the function.
```
const colors = [
    "#ff0000",
    "rgba(255, 255, 255, 0.5)",
    "#003300",
    "#005500",
    "#007700",
    "#009900",
    "#00bb00",
    "#00ff00",
  ];

  const ticks = ["Loss", 0, 200, 400, 600, 800, 1000];
  const colorScaleFn = (value) => {
    if (value < 0 || value === "Loss") {
      return colors[0];
    }
    if (value == 0) {
      return colors[1];
    }
    for (let i = 2; i < ticks.length; i++) {
      if (value < ticks[i]) {
        return colors[i];
      }
    }
    return colors[colors.length - 1];
  };

  colorScaleFn.ticks = () => ticks;
```

and then `<ResponsiveTimeRange colorScale={colorScaleFn} />`

PR #704, opened 2019-10-11
add borderColor type definition for nivo/bar

alexsteele110:
contributes to #197 where borderColor was missing from TypeScript definitions in nivo/bar. From what I could see in the documentation borderColor has the same definitions as labelTextColor so I just simply reused them


plouc:
Thank you @alexsteele110!

PR #720, opened 2019-10-15
Adding showEmpty prop

thefzn:
Regular Bar component won't display 0 values (or values with a size near to 0), but sometimes, we need to display those values, or at least, the labels/tooltips.

This update allows us to pass the `showEmpty` prop (boolean) so we can force display 0 (or near to 0) values.


ghmeier:
@wyze any thoughts on this? It'd be useful functionality to have.

PR #731, opened 2019-10-27
Add option to format x/y values in tooltip

yotov:
This PR add support of xFormat/yFormat in LineCanvas chart. Currently Tooltip in LineCanvas could not be used with time series data due to missing option to use formatter


plouc:
Thank you for your contribution @yotov!

PR #734, opened 2019-10-28
fix(timescale): allow custom non-native min/max value

pheel:
Most likely a typo. Previous code would not let user set min/max value on a timescale unless format was set to 'native'.


wyze:
Thanks for the submission @pheel! Went with #743 which included this fix as well as some typings.

Issue #596, opened 2019-06-12
gridXValues or gridYValues not a recognized prop for Bar Chart


shihlinlu:
I'm getting a prop error when adding `gridXValues` and/or `gridYValues`. As per documentation, they are optional props and I found a closed issue where they were implemented. Has this since been removed in the latest version?

1. Add `gridXValues` or `gridYValues` prop

**Expected behavior**
There should be no error for prop.

![Prop Error](https://user-images.githubusercontent.com/4715629/59389221-0ea03000-8d2b-11e9-8e22-4e3c71fd4047.png)

wyze:
They still exist. It looks like the TypeScript definitions just need to be updated.

shihlinlu:
Thanks for the update. Is anyone in the process of updating the TS definitions? Or is this something I can open a PR if I have time?

wyze:
A PR is more than welcome! There is a long running issue to get types added and up to date, so every little bit helps.

shihlinlu:
Thanks for confirming! I'll take a stab at it when I have time to work on this and submit a PR! 

I believe the TS definitions may need to be updated for these other `ResponsiveBar` props as well: `colors`. I have noticed importing  the linear gradient pattern from the core Nivo package is also causing errors for TS.

jcalkins89:
Was anyone able to get to this to update the TS definitions? I'm experiencing similar issues for both the `colors` prop for ResponsiveBar and the `gridYValues` prop

chriscasella:
@jcalkins89 @shihlinlu I'm sure this is not the "right thing to do" but I can't wait for a PR and I need this now. So I went into `node_modules\@nivo\bar\index.d.ts` and in the BarProps Partial I added
```
        gridXValues: Number | Array[Number] | null
        gridYValues: Number | Array[Number] | null
``` 

It works perfectly for me so far. I hope it helps you as well! Cheers :)

shihlinlu:
@chriscasella @jcalkins89 I submitted a PR a few days ago, but I need to fix the typing again. @chriscasella Your typing should be correct, I'll update it with that! Thanks!

kaeland:
[...]

PR #737, opened 2019-10-30
fix(bar): Update PropTypes and TypeScript typing for Bar to make grid…

shihlinlu:
Need to update TypeScript typing for `gridXValues` and `gridYValues`. 

Should close https://github.com/plouc/nivo/issues/596

### DIFF
--- a/packages/bar/src/props.js
+++ b/packages/bar/src/props.js
@@ -42,8 +42,8 @@ export const BarPropTypes = {
     axisRight: axisPropType,
     axisBottom: axisPropType,
     axisLeft: axisPropType,
-    enableGridX: PropTypes.bool.isRequired,
-    enableGridY: PropTypes.bool.isRequired,
+    enableGridX: PropTypes.bool,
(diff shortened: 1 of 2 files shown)


audiolion:
I am all for this PR, but #596 is actually referencing the missing `gridXValues` and `gridYValues` types missing from Bar

packages/bar/src/props.js
```diff
@@ -42,8 +42,8 @@ export const BarPropTypes = {
     axisRight: axisPropType,
     axisBottom: axisPropType,
     axisLeft: axisPropType,
-    enableGridX: PropTypes.bool.isRequired,
-    enableGridY: PropTypes.bool.isRequired,
+    enableGridX: PropTypes.bool,
```
plouc: You can leave those as required as there is a default value for both

wyze:
Thanks @shihlinlu, but it looks like this has already been fixed. I think there were several outstanding PRs that were fixing the same thing.

PR #743, opened 2019-11-10
fix(scales): fix timeScale min/max values and typings

grogi:
* updated the TypeScript definition to allow for `min` and `max` properties in the `TimeScale`
* updated the implementation to behave correctly with custom `min`/`max` values when the `TimeScale` has `format` property specified.


grogi:
Pull request #734 is a subset of this one.

PR #758, opened 2019-11-20
Update package.json

vianneyguesdon:
add keywords to be able to find Nivo via npm search bar.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "name": "Raphaël Benitte",
     "url": "https://github.com/plouc"
   },
-  "keywords": [],
+  "keywords": ["react", "charts", "dataviz","react-component", "d3", "graphs", "data-visualization"],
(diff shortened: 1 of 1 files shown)


coooraph:
Thank you for this @vianneyguesdon, however now this file is only used to install infrastructure dependencies, the `nivo` package is deprecated and scoped packages (like `@nivo/treemap`) should be used instead

coooraph:
It's now [private](https://github.com/plouc/nivo/blob/master/package.json#L4)

vianneyguesdon:
Understood. I should be delete or hidden to avoid mistakes imo !
https://www.npmjs.com/package/nivo still accessible

plouc:
@vianneyguesdon, 361 repos apparently have it as a dependency, I don't want to break those, people can still use the old one.

Issue #794, opened 2019-12-12
Piechart legends label is not respected


cerber717:
**Describe/explain the bug**
While using `ResponsivePie` with data : 
`data = [
          {
            'id': 'id_of_metric',
            'label': 'Metric,
            'value': 434,
          }}`
the label is not set to the what is in the data object but rather to the id. 
I dug into it a little bit and found that the issue could be in here 
https://github.com/plouc/nivo/blob/41dd564fcc740e99fa4a1a58ba707b22a877a7b8/packages/pie/src/PieLegends.js#L50

cerber717:
https://github.com/plouc/nivo/pull/797

PR #797, opened 2019-12-13
PieLegend label fix

cerber717:
This is a fix for the issue 
https://github.com/plouc/nivo/issues/794


le0nik:
@wyze looks like it is in fact required https://github.com/plouc/nivo/issues/808 but it doesn't look like it should be.

plouc:
PropTypes define constrains for both internal and external API, while typings only define external/public API

cerber717:
I am gonna add the fall back for.

wyze:
Thanks for this @cerber717, but I went ahead and fixed this and something else in #967.

PR #802, opened 2019-12-18
Various TypeScript fixes across several packages

wyze:
These are some of the fixes we still had locally after upgrading to the latest version of the charts we use.

If you review each commit, they should be descriptive enough, but if not, I can expand further.


packages/legends/index.d.ts
```diff
@@ -93,4 +94,6 @@ declare module '@nivo/legends' {
     }
 
     export type QuantileLegendSvg = React.FunctionComponent<QuantileLegendProps>
+
+    export const BoxLegendSvg: React.FC<LegendProps & Dimensions>
```
agustingabiola: Looking into `https://github.com/plouc/nivo/blob/master/packages/legends/src/svg/BoxLegendSvg.js` Proptypes I don't think these match.

wyze: Which ones specifically?

agustingabiola: https://github.com/plouc/nivo/blob/master/packages/legends/src/svg/BoxLegendSvg.js#L113

You don't pass `height` & `width` but `containerWidth` & `containerHeight`

mahatosourav91:
@wyze @agustingabiola @plouc 

Would be possible to get this merged and have a minor release of legends?

packages/legends/index.d.ts
wyze: Good catch, thank you.

Issue #840, opened 2020-01-23
Bars sizes are rounded and don't take the whole chart area


constgen:
Bars lines in `Bar` chart do not cover all area of a chart (SVG element). Their sizes are rounded. That's why we can see gaps on the left and right. This screenshot shows this issue comparing Histograms with Timeline

**Screenshots**
![image](https://user-images.githubusercontent.com/673144/72987183-89b5e280-3df2-11ea-8fb6-3dc67434c226.png)

To reproduce this you can use for example `width=200` and 92 child items

**Expected behavior**
Bars should take all space from left to right. The dimensions are integers `<g transform="translate(8, 20)">`. But float numbers are expected

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Firefox
 - Version 0.61.1

constgen:
The proposed PR fixes the issue. Now Histograms look correct 
![image](https://user-images.githubusercontent.com/673144/72988346-bec33480-3df4-11ea-9ffe-2d0e46acb1c0.png)

Demensions look as expected `<g transform="translate(2.150537634408602, 14)">`

mwinkels-onegini:
👍 
maybe make this an option

constgen:
Maybe not. Because coordinates on all other charts are float numbers

constgen:
bump

wyze:
@constgen Can you provide a CodeSandbox demonstrating above, or close to it, so I can check it out against your proposed PR please?

constgen:
I will try

wyze:
Here is one that represents the issue: https://codesandbox.io/s/busy-violet-s19us

constgen:
My example https://codesandbox.io/s/nivo-bar-chart-ki9h7

wyze:
This was added in #1282. You can use `indexScale={{ type: 'band', round: false }}`. This will be available in the next version.

PR #841, opened 2020-01-23
Don't round a bar scale

constgen:
Fixes https://github.com/plouc/nivo/issues/840


constgen:
Snapshots tests failed. Please, help me to update them if you think that this PR doesn't brake anything.

constgen:
bump

wyze:
@plouc Was there a specific reason to use rounded values for the index scale? Is this change going to be okay, or should we look into making some prop like `indexScale={{ round: false }}` that users can opt into?

constgen:
Because rounding is **disabled** for all other charts. More details in the referred issue.

plouc:
@wyze, @constgen I remember adding it for a specific use case I had, should maybe not be the default, but I think we should keep the option.

constgen:
Do you expect it to brake something?

wyze:
This was added in #1282. You can use `indexScale={{ type: 'band', round: false }}`. This will be available in the next version.

wyze:
Thanks for the contribution, and sorry it didn't work out landing it this way.

constgen:
The new option may help

PR #842, opened 2020-01-23
Fix bar legend order

manner:
When using the `vertical stacked bar` chart with a `row` legend, the legend order was in reverse. This PR fixes this issue by preventing the legend data to be reversed, when it is in `row` format. 

**Before**
image

**After**
image


coooraph:
@manner thank you!

dooleyb1:
Would love if this could be approved and merged 🙏

plouc:
Thank you for your contribution @manner!

PR #843, opened 2020-01-24
feat(calendar): allow day color to be assigned

sergioramos:


diegosps:
This can be done passing a custom `colorScale` function

PR #844, opened 2020-01-24
feat(line): add gradient support to line areas

AnteWall:
Add Gradient Support to Line Areas

### DIFF
--- a/packages/line/src/Line.js
+++ b/packages/line/src/Line.js
@@ -7,7 +7,14 @@
  * file that was distributed with this source code.
  */
 import React, { Fragment, useState, useMemo } from 'react'
-import { withContainer, useDimensions, useTheme, SvgWrapper, CartesianMarkers } from '@nivo/core'
+import {
+    bindDefs,
+    withContainer,
+    useDimensions,
+    useTheme,
+    SvgWrapper,
+    CartesianMarkers,
+} from '@nivo/core'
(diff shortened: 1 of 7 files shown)


SnarkieDesign:
@plouc this is working well, can you review?

packages/line/src/Line.js
```diff
@@ -62,6 +69,9 @@ const Line = props => {
         pointLabelFormat,
         pointLabelYOffset,
 
+        defs,
```
SnarkieDesign: Do we need to add defs and fill to the Type definitions? i.e.

`export interface LineSvgProps extends LineProps, MotionProps, **SvgDefsAndFill<Datum>** { ... }`

catalinberta:
@plouc 
Really looking forward to this, could we have this reviewed asap? I can imagine you are super busy, thank you!

panutat:
Would be incredible to get this in

packages/line/src/Line.js
wyze: Thanks, good catch!

wyze:
Great work @AnteWall! Let me know if the storybook example can be better, maybe something you have used? Right now it just fades from the area color to transparent.

riencoertjens:
😻 when do these kind of changes get released?

wyze:
It will go out in the next release. I can't give a definite answer, but I've talked with plouc and it will be soon.

PR #853, opened 2020-02-02
add grid{X/Y]Values to ScatterPlot

ericsliu:
These props are defined here: https://nivo.rocks/scatterplot/
but they're not actually in the ScatterPlot component.


plouc:
Thank you for your contribution @ericsliu!

Issue #856, opened 2020-02-05
Typescript error in scatterplot when `noImplicitAny` enabled


stuartlong:
**Describe/explain the bug**
The typings for ScatterPlot create errors for Typescript projects that have the [noImplicitAny](https://www.typescriptlang.org/docs/handbook/compiler-options.html) compiler option enabled.

Specifically thje typing for `CustomTooltip`:
```ts
export type CustomTooltip = ({ node: Node }) => React.ReactNode
```

which gives the error:
![image](https://user-images.githubusercontent.com/2327459/73884594-12814500-481b-11ea-8377-728a71228eec.png)

**To Reproduce**
Enable `noImplicitAny: true` in the tsconfig of the nivo project. Code sandbox demonstrating the error: https://codesandbox.io/s/noimplictany-error-h1lj2

**Expected behavior**
Nivo can be imported into Typescript projects using `noImplicitAny`

PR #857, opened 2020-02-06
fix(scatterplot): fix no implicit any error on CustomTooltip

stuartlong:
Fix for #856.

The `CustomTooltip` typing from scatterplot was creating an error in Typescript projects with the `noImplicitAny` compiler optoin enabled. This error was because the argument to the `CustomTooltip` lambda wasn't being typed:

```ts
export type CustomTooltip = ({ node: Node }: /* NEEDS TYPING */) => React.ReactNode
```

So the fix is to create a specific `TooltipProps` for the argument and enable `noImplicitAny` across the project to prevent regressions.

### DIFF
--- a/packages/scatterplot/index.d.ts
+++ b/packages/scatterplot/index.d.ts
@@ -84,7 +84,9 @@ declare module '@nivo/scatterplot' {
         sizes: [number, number]
     }
 
-    export type TooltipProps = { node: Node };
+    export interface TooltipProps {
+        node: Node
+    }
     export type CustomTooltip = (props: TooltipProps) => React.ReactNode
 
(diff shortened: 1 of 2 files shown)


stuartlong:
@plouc , can you retrigger the build on this? Looks like it was a transient network isse:
```
error An unexpected error occurred: "https://registry.yarnpkg.com/ua-parser-js/-/ua-parser-js-0.7.18.tgz: ETIMEDOUT".
```

laura-mso:
Hi, I don't want to create a duplicate PR, but I am in need of this fix too. It would be awesome if you could rerun the checks and merge this in :)

PR #868, opened 2020-02-10
Add missing typings

volkerchartier:
For onMouseEnter, onMouseLeave


wyze:
Thanks for the submission @volkerchartier, but we went with #939 instead as it was more complete.

PR #878, opened 2020-02-21
feat(pie): add radialLabelsNoClip flag, minAngle parameter

jazzgil:
1. add radialLabelsNoClip flag (default false) to prevent clipping at top or bottom. This flag allows to create a pie with no margins at top and bottom, and re-positions clipped radial labels along sides of the pie.
2. add minAngle parameter (default 0) to set minimum angle of any slice.


jazzgil:
bump...

PR #887, opened 2020-03-03
fix(line): typings fix for custom layer props

stuartlong:
Typings fixes for two areas of Line:
1. adds "catmullRom" as a supported curve string

2. Populates `CustomLayerProps` with some more properties
  - it's still missing `areaGenerator` but I'm not using that so wasn't sure what the proper typing for it would be


wyze:
Thanks @stuartlong!

PR #846, opened 2020-01-24
Add colorBy type declaration in BarProps type

sciamp:
`BarProps` type lacks the `colorBy` type definition, the goal of this PR is to add it to allow users to specify a custom value for `colorBy` while using typescript.

Thanks for your time :)

PR #904, opened 2020-03-27
Add missing colorBy property in Bar in TypeScript definition

codler:


JabbyPanda:
Deiplicated of #846

plouc:
`colorBy` has been deprecated in favor of directly using `colors`, you can see an example of it [here](https://nivo.rocks/storybook/?path=/story/bar--using-custom-color). And sorry for the late feedback.

RoyBkker:
> `colorBy` has been deprecated in favor of directly using `colors`, you can see an example of it [here](https://nivo.rocks/storybook/?path=/story/bar--using-custom-color). And sorry for the late feedback.

But why is colorBy still used than in the example in the storybook?

PR #915, opened 2020-04-09
Added crosshair definitions

vladdes:
Will add more as I come across it. Love your solution btw!


plouc:
Thank you for your contribution @vladdes!

PR #916, opened 2020-04-10
feat(bar): pass showTooltip and hideTooltip functions to custom layers

### LINKED ISSUE #914: Adding tooltips to customised layers
I have added a custom layer to mark up a bar chart showing a high/mid/low indicator on top of the bar chart.

Here's an example of the end product:

bar-with-benchmarks

I would like to add a tooltip to the customised layer to display information about the high/mid/low indicators, however I noticed that `showTooltip`/`hideTooltip` from the `<Container>` component is not passed through to the custom layer.

I have looked through quite a few of the custom layer solutions on code sandbox and haven't come across any that tackle this situation.

Am I missing something or do I have to roll my own solution? Thanks

dancallaghan:
In the same vein as #425, expose `showTooltip`/`hideTooltip` functions to custom layers in the bar chart.

Resolves #914

### DIFF
--- a/packages/bar/src/Bar.js
+++ b/packages/bar/src/Bar.js
@@ -318,7 +318,11 @@ const Bar = props => {
                     >
                         {layers.map((layer, i) => {
                             if (typeof layer === 'function') {
-                                return <Fragment key={i}>{layer({ ...props, ...result })}</Fragment>
+                                return (
+                                    <Fragment key={i}>
+                                        {layer({ ...props, ...result, showTooltip, hideTooltip })}
+                                    </Fragment>
+                                )
                             }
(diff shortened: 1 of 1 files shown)


plouc:
@dancallaghan thank you for your contribution, please note that in the future (probably not soon though :)) you'll be able to retrieve those from the context. Scatterplot [already uses this approach](https://github.com/plouc/nivo/blob/master/packages/scatterplot/src/NodeWrapper.js#L30).

dancallaghan:
No worries! :)

Also, interesting about the `useContext`, I guess it's a big task moving all the packages over to the hook paradigm.

PR #937, opened 2020-04-27
fix(heatmap): add missing type for tooltip property

thomasthiebaud:


plouc:
Thank you for your contribution @thomasthiebaud!

PR #944, opened 2020-05-05
#197 Updated @nivo/line typescript definition to add all curve options

escobar5:
Some curve options where missing (basis, cardinal, catmullRom)


plouc:
Thank you for your contribution @escobar5!

PR #963, opened 2020-05-18
chore(build): fix typescript-migration branch build

wyze:
This PR addresses a few things and allows the build to pass.

- Move from `dependencies` to `dev` + `peer` dependencies for workspace packages
- Switch from ts-jest to babel for jest tests
- Upgrade CI to run on node 12
- Run `make fmt` to get everything formatted correctly
- Upgrade to TypeScript 3.9.3
- Add `make packages-clean` command

Need to figure out the best way to remove the cyclical dependencies between core and tooltip. Right now I commented out tooltip from core.

~~Also, I am trying to get `yarn build` command to work, but it is failing on the radar package saying it can't find `@nivo/tooltip`. Radar uses core, colors, and legends just fine though. So I am confused why it is saying it cannot find tooltip.~~

Looks like there was still some using in `withContainer`.

Trying to model this architecture off of what the xstate project is doing.


packages/core/src/components/Container.tsx
```diff
@@ -135,9 +135,9 @@ export const Container = ({
                 stiffness={motionStiffness}
                 damping={motionDamping}
             >
-                <TooltipContext.Provider value={[showTooltip, hideTooltip]}>
-                    {content}
-                </TooltipContext.Provider>
+                {/* <TooltipContext.Provider value={[showTooltip, hideTooltip]}> */}
```
wyze, line 138: We will need to address tooltip usage here.

packages/core/src/hocs/withContainer.js
```diff
@@ -25,39 +25,56 @@ const Container = ({
...
-    const {
-        showTooltipAt,
-        showTooltipFromEvent,
-        hideTooltip,
-        isTooltipVisible,
-        tooltipContent,
-        tooltipPosition,
-        tooltipAnchor,
-    } = useTooltipHandlers(container)
+    // const {
```
wyze, line 28: We will need to address tooltip usage here.

PR #964, opened 2020-05-19
feat(calendar): add monthSpacing prop

fguitton:
This PR adds the prop `monthSpacing` to `@nivo/calendar` as an option to add spacing between months in the rendered output.
image


fguitton:
Hello ? Anybody here ? 😄 
Not sure who is supposed to review this, would it be @plouc ?

wyze:
Looks good! I added the same prop to the website so users can play with it, as well as updated the types.

Issue #978, opened 2020-06-03
 Add custom layer support to LineCanvas charts


ojvilau:
I need to add hundreds, even thousands, of horizontal and vertical lines, like a grid, that may or not match with the axises points in a Line chart.

In order to do it, maintaining performance, it's necessary to use LineCanvas and a custom layer.

Currently, custom layers are not supported in LineCanvas charts.

Looking forward to seeing this feature, anyway, amazing library. Thanks for the hard work

PR #987, opened 2020-06-08
feat(line): add custom layer support to LineCanvas

jguddas:
2020-06-08-194724_934x742_scrot

This might already resolve @ojvilau 's problem #978  
but should definitely be looked at by someone who knows their way around this codebase.


plouc:
Thank you very much @jguddas, I'll have a look!

wyze:
Looks great @jguddas!

PR #1039, opened 2020-06-29
feat(tooltip): make tooltip provider component

wyze:
By moving the provider for tooltip into the `@nivo/tooltip` package we can avoid having the current circular dependencies between core and tooltip. This is the same functionality that is in `withContainer` HOC as it looks to be the newer way to handle tooltips.


wyze:
Looks like the build is passing but the checks still say pending.

Issue #884, opened 2020-02-28
Tracking issue for removing `recompose` dependency


stephent:
Updating React now leads to a deprecation warning in the console:

```
react.development.js:315 Warning: React.createFactory() is deprecated and will be removed in a future major release. Consider using JSX or use React.createElement() directly instead.
    printWarning @ react.development.js:315
    warn @ react.development.js:278
    createFactoryWithValidation @ react.development.js:1827
    (anonymous) @ shouldUpdate.js:18
    pure @ pure.js:22
    (anonymous) @ compose.js:13
    enhance @ nivo-voronoi.esm.js:82
    ./node_modules/@nivo/voronoi/dist/nivo-voronoi.esm.js @ nivo-voronoi.esm.js:170
```

This is due to the `recompose` package which is still used for some of the nivo packages, we have to migrate the following packages to React hooks (providing the same features) to get rid of it:

- [x] Core
- [x] Pie
- [x] Voronoi (#1364)
- [x] Waffle (#2320)
- [x] Bullet (#1252)
- [x] Bar (#1619)
- [x] Circle Packing (#1360)
- [x] Sunburst (#1264)
- [x] Calendar
- [x] Legends (Listed in package.json, but not used)
- [x] Parallel Coordinates, indirect use via core (#2329)

codler:
Im getting same warning

abhaykumar01234:
Getting the same issue, How to remove the error

hichana:
Same issue :(

neverdane:
Same here :(

jungans:
Same...

mattcarlotta:
This isn't a direct problem with the nivo packages, but instead of with one of its dependencies: **recompose**. 

Specifically `recompose/shouldUpdate`: https://github.com/acdlite/recompose/blob/master/src/packages/recompose/shouldUpdate.js#L6 

and `recompose/withPropsOnChange`: https://github.com/acdlite/recompose/blob/master/src/packages/recompose/withPropsOnChange.js#L9 

Unfortunately, the recompose package appears to be [mostly unmaintained](https://github.com/acdlite/recompose#a-note-from-the-author-acdlite-oct-25-2018).

codler:
Reading recompose readme, it suggest you use React Hooks instead

AmmarHSufyan:
Same

jungans:
[...]

Issue #745, opened 2019-11-12
Tooltips are not consistent for all components


taschetto:
**Is your feature request related to a problem? Please describe.**
It is not possible to render a custom tooltip within a Calendar day without wrapping it with a BasicTooltip (https://github.com/plouc/nivo/blob/master/packages/calendar/src/CalendarDay.js#L92-L104).

**Describe the solution you'd like**
We should be able to use a custom component alone, without the need to wrap it with a BasicTooltip, just like the Line/Mesh tooltip (https://github.com/plouc/nivo/blob/master/packages/line/src/Mesh.js#L43-L47).

**Describe alternatives you've considered**
I haven't found an alternative.

**Additional context**

Custom tooltip being used with line charts
![Custom tooltip being used with line charts](https://user-images.githubusercontent.com/5279182/68678134-25e28400-053c-11ea-8570-aead37e8f3ab.png)

The same tooltip being used with calendars, wrapped inside a BasicTooltip
![The same tooltip being used with calendars, wrapped inside a BasicTooltip](https://user-images.githubusercontent.com/5279182/68678229-54605f00-053c-11ea-8df5-aa2ec36db137.png)

plouc:
That's also the case for scatterplot, see https://github.com/plouc/nivo/issues/754

plouc:
@taschetto I took the liberty to rename this issue instead of having one for each component type behaving differently.

ianstormtaylor:
@plouc I think this same issue affected the `<ResponsiveTreemap>`? I'd like to have the same tooltip behavior from line charts to have complete control over the rendering, but it's getting wrapped in an extra box.

Issue #744, opened 2019-11-12
How to format legend labels within a Calendar?


taschetto:
Hi! I've searched the docs, storybook and this repo to find a way to do it but I was not successful.

I need to format legend labels within a Calendar. Currently, my raw values range from 0 to ~70,000.

![image](https://user-images.githubusercontent.com/5279182/68676095-fcbff480-0537-11ea-953d-caaeec45eda5.png)

I'd like to format these values as follows:

![image](https://user-images.githubusercontent.com/5279182/68676205-401a6300-0538-11ea-9dc2-8206ea00f783.png)

Is there anyway to do it?

I've tried to pass a function inside the legend object, but it didn't work:

```javascript
    // ...
    legends: [
      {
        // ...
        legendFormat: ({ value }) => myFormatter(value),
        // ...
      },
    ],
```

Thanks! :-)

oscargws:
Bump, any information on this? Trying to achieve the same thing

rovaniemi:
I also needed this feature, but the library does not support it yet. However, this feature can be used with a little workaround, but **we should do a PR** if we want to use it the proper way.

#### Workaround

**@nivo/calendar** uses colorScale function when it does create legends. 

```js
const legendData = colorScale.ticks(legend.itemCount).map(value => ({
  id: value,
  label: value,
  color: colorScale(value),
}))
```

We can pass new colorScale function on **@nivo/calendar** props
```js
import { scaleQuantize } from 'd3-scale'

const computeDomain = (data, minSpec, maxSpec) => {
  const allValues = data.map(d => d.value)
  const minValue = minSpec === 'auto' ? Math.min(...allValues) : minSpec
  const maxValue = maxSpec === 'auto' ? Math.max(...allValues) : maxSpec
  return [minValue, maxValue]
}

const createColorScale = (data, minValue, maxValue, colors) => {
  const domain = computeDomain(data, minValue, maxValue)
  const defaultColorScale = scaleQuantize()
      .domain(domain)
      .range(colors)
  const colorScale = value => {
      const number = prettyIntegerToNumber(value) // Formatted number back to number
      return defaultColorScale(number)
  }
  colorScale.ticks = count => {
     return defaultColorScale.ticks(count).map(e => prettyInteger(e)) // Format number
  }
  return colorScale
}
```

Genarito:
Another workaround is to make your custom Legend as proposed [here](https://github.com/plouc/nivo/issues/791#issuecomment-587051981).

[...]

PR #1040, opened 2020-06-30
feat(calendar): remove recompose and convert to hooks

diegosps:
Hi! 
I removed the dependency from recompose and ported the code to use hooks. This way, Calendar should be free from  https://github.com/plouc/nivo/issues/884
I tried my best to follow the commit pointed by @wyze on that issue, but I have no experience with `Typescript`. Please, double check `packages/calendar/index.d.ts` (and everything else, to be honest)

I also changed:

- Tooltip padronization https://github.com/plouc/nivo/issues/745
- Possibility¹ to use custom formats for tooltips and legend independently https://github.com/plouc/nivo/issues/744
- Removed the "duplicate" rectangle rendering from svg²

1 - I think the `legendFormat` might be better suited at the legend component, but since I was following some pattern with `useValueFormatter`, I wasn't able to put it there. Maybe I need a little more thinking.

2 - For each day, the SVG Calendar rendered two rectangles: one with the spacing and the color fill, one with transparent fill and mouse events, but without any spacing. The point here was to be able to raise events while the mouse was over the "empty" space created by `daySpacing`. 
After migrating to hooks, I was experiencing a bug where, sometimes, the tooltip didn't show. After I eliminated the second `rect` and put the mouse events on the first one, this bug disappeared. Also, during my limited visual tests, the tooltip fluidity was way better this way. The 'downside' of this change is: we are no longer able to see tooltips while hovering the empty space created by `daySpacing`. I'm ok with that.

Other possible break change is: `tooltipFormat` is gone.

I also created some stories for the above changes and for the custom color scale function.

I'm not sure if I need to change anything for website and api.

Please, feel free to tell me what I can do to improve this patch.

Thanks


wyze:
@diegosps Can you get the build passing please? Looks like it's just a formatting issue.

diegosps:
All passing now

packages/calendar/src/Calendar.js
```diff
@@ -7,111 +7,146 @@
...
  */
 import React from 'react'
-import { Container, SvgWrapper } from '@nivo/core'
+import { SvgWrapper, useTheme, useDimensions, withContainer, useValueFormatter } from '@nivo/core'
 import { BoxLegendSvg } from '@nivo/legends'
-import { setDisplayName } from 'recompose'
-import { CalendarPropTypes } from './props'
-import enhance from './enhance'
+import { CalendarPropTypes, CalendarDefaultProps } from './props'
+
```
wyze, line 13: Remove extra new line.

packages/calendar/src/Calendar.js
```diff
@@ -7,111 +7,146 @@
...
 import CalendarYearLegends from './CalendarYearLegends'
 import CalendarMonthPath from './CalendarMonthPath'
 import CalendarMonthLegends from './CalendarMonthLegends'
+
```
wyze, line 17: Remove extra new line.

packages/calendar/src/Calendar.js
```diff
@@ -7,111 +7,146 @@
...
+        height: innerHeight,
+        from,
+        to,
+        direction,
+        yearSpacing,
+        monthSpacing,
+        daySpacing,
+        align,
+    })
+    colorScale = useColorScale({ data, minValue, maxValue, colors, colorScale })
```
wyze, line 81: Let's avoid reassigning function parameter (prop).

packages/calendar/src/Calendar.js
```diff
@@ -7,111 +7,146 @@
...
-    months,
-    days,
 }) => {
+    const theme = useTheme()
+    const { margin, innerWidth, innerHeight, outerWidth, outerHeight } = useDimensions(
+        width,
+        height,
+        partialMargin
+    )
+    let { months, years, days } = useCalendarLayout({
```
wyze, line 70:
```suggestion
    const { months, years, ...rest } = useCalendarLayout({
```

Avoid reassignment when needed, and we can make this a `const`.

packages/calendar/src/Calendar.js
```diff
@@ -7,111 +7,146 @@
...
+    })
+    colorScale = useColorScale({ data, minValue, maxValue, colors, colorScale })
+    const monthLegends = useMonthLegends({
+        months,
+        direction,
+        monthLegendPosition,
+        monthLegendOffset,
+    })
+    const yearLegends = useYearLegends({ years, direction, yearLegendPosition, yearLegendOffset })
+    days = useDays({ days, data, colorScale, emptyColor })
```
wyze, line 89:
```suggestion
    const days = useDays({ days: rest.days, data, colorScale, emptyColor })
```

packages/calendar/src/CalendarCanvas.js
```diff
@@ -6,225 +6,291 @@
...
+            height: innerHeight,
+            from,
+            to,
+            direction,
+            yearSpacing,
+            monthSpacing,
+            daySpacing,
+            align,
+        })
+        colorScale = useColorScale({ data, minValue, maxValue, colors, colorScale })
```
wyze, line 101: Avoid reassigning function parameter.

packages/calendar/src/CalendarCanvas.js
```diff
@@ -6,225 +6,291 @@
...
+    }) => {
+        const canvasEl = useRef(null)
+        const { innerWidth, innerHeight, outerWidth, outerHeight, margin } = useDimensions(
             width,
             height,
-            outerWidth,
-            outerHeight,
+            partialMargin
+        )
+        let { months, years, days } = useCalendarLayout({
```
wyze, line 90:
```suggestion
        const { months, years, ...rest } = useCalendarLayout({
```

packages/calendar/src/CalendarCanvas.js
```diff
@@ -6,225 +6,291 @@
...
+            monthLegendPosition,
+            monthLegendOffset,
+        })
+        const yearLegends = useYearLegends({
+            years,
+            direction,
+            yearLegendPosition,
+            yearLegendOffset,
+        })
+        days = useDays({ days, data, colorScale, emptyColor })
```
wyze, line 114:
```suggestion
        const days = useDays({ days: rest.days, data, colorScale, emptyColor })
```

packages/calendar/src/compute.js
```diff
@@ -380,6 +380,7 @@ export const bindDaysData = ({ days, data, colorScale, emptyColor }) => {
                 day.value = dayData.value
                 day.color = colorScale(dayData.value)
                 day.data = dayData
+                return
```
wyze, line 383: What was this needed for?

packages/calendar/src/hooks.js
```diff
@@ -0,0 +1,142 @@
...
+            bindDaysData({
+                days,
+                data,
+                colorScale,
+                emptyColor,
+            }),
+        [days, data, colorScale, emptyColor]
+    )
+
+export const useDaysHandlers = ({
```
wyze, line 91: Looks like this hook isn't used so we can remove it.

packages/calendar/src/compute.js
diegosps: To break from the foreach loop. Since there isn't a duplicate day in a calendar, there is no need to evaluate the rest of the data after finding the right day.
This way we can save n² - (n²)/2 operations. At least, this is my undesrtanding.

packages/calendar/src/hooks.js
diegosps: That's true. I put those inside Calendarday.js after and forgot to remove it.

packages/calendar/src/Calendar.js
diegosps: If the value is assigned, the same value is returned inside the useColorScale. If not, a new one is created.
I'm not sure if this is a problem, but I will change the name any way to be more clear.

packages/calendar/src/compute.js
diegosps: We can go further and put a binary search there instead of the foreach. This way instead of O(n²) this part would be O(nlogn). But i'm not sure if it is worth the extra lines for such small data size. It's not like we have millions of day in a single calendar component. The return statement was a simple one.

wyze:
Yeah just resolve the comments as you make the changes and push! :)

czonios:
Hello everyone,

any word on when the release containing this change will be released to npm? I specifically need the granularity feature, to show only a few months.

Keep up the great work, love this library!

wyze:
Unfortunately I cannot make releases to npm yet, so we are waiting on @plouc to get a release out.

mthanasi:
Hey guys,

Firstly, awesome work!
I really need the granularity feature, to show only a 1-2 months max. Any update on when will this change be released to npm?

PR #1047, opened 2020-07-04
Prevent Line drawing outside data area using clip

### LINKED ISSUE #660: Line Y-axis overlap
Hey @plouc ! Thank you for this great library! The performance is really really great!!
However, I experience a similar problem but with the line chart. As you can see in the image below, I dynamically set min and max on the x-axis (time-based, varies according to user selection) which causes the line and the x-axis to overlap. Is there any fix for this which I just couldn't find yet? If not, could you give me a hint on how and where to implement an option to enable/disable this behavior? Of course I'd be willing to share the code/create a PR if it fits your needs as well.
image

Thank you in advance!!

P.S. Setting min/max for the x-axis is not part of the documentation but works - except for the overlap - really well!

trepidacious:
Fixes #660
Add a clip-path to the container g of lines and points in Lines graph to prevent drawing outside the data area. This requires Lines and Points to accept a width and height - if either one is omitted, clipping is not done.

The implementation is probably not ideal, but shows the general approach.

### DIFF
--- a/packages/line/src/Line.js
+++ b/packages/line/src/Line.js
@@ -176,7 +176,14 @@ const Line = props => {
         ),
         areas: null,
         lines: (
-            <Lines key="lines" lines={series} lineGenerator={lineGenerator} lineWidth={lineWidth} />
+            <Lines 
+                key="lines" 
+                lines={series} 
+                lineGenerator={lineGenerator} 
+                lineWidth={lineWidth} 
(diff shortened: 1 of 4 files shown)


bmwebster:
Is there any chance at looking to pull this or a better solution? Generally a zoomable graph needs to clip data outside the graph to give a reasonable appearance.

PR #1057, opened 2020-07-14
feat(bar): add prop to include empty values

### LINKED ISSUE #874: [ResponsiveBar] - Show Entries in Legend for 0 Value Data Points
I'm currently rendering a `<ResponsiveBar>` element that does not contain stacked values across the range and is similar. However, for entries in the bar chart that take the value 0 there is no entry in the legend corresponding to this option.

I think it'd be nice if there was an option to display a legend for 0 values with a color that that they would take in the case that they did have a non-zero value.

Read around the docs and previous issues but couldn't find anything that matched what I was looking for.

**Additional context**
Screenshot 2020-02-14 at 17 30 32

ghmeier:
Adds the `includeEmpty` prop to `ResponsiveBar` and `Bar` component so that groups where the value is 0 will be included in the results.

Fixes #874

I chose this method to avoid breaking changes and ensure bars are correctly formatted with 0 values.

### DIFF
--- a/packages/bar/src/Bar.js
+++ b/packages/bar/src/Bar.js
@@ -64,6 +64,7 @@ const Bar = props => {
         outerHeight,
         padding,
         innerPadding,
+        includeEmpty,
 
         axisTop,
         axisRight,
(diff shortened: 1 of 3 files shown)


thisizkp:
@plouc any thoughts on this PR? When are we planning to merge this?

PR #1059, opened 2020-07-15
chore(docs): Add Testing section in CONTRIBUTING.md

phacks:
Finding out how to launch tests was only a quick look at `Makefile` away, but I think it makes sense to have it in the CONTRIBUTING docs—it also has the benefits of stating that there _are_ tests to check your PR against.


wyze:
I agree, good place to have it is in the contributing docs.

PR #1060, opened 2020-07-15
fix(line): change points ordering on stacked lines

### LINKED ISSUE #1051: Misleading points displayed in stacked line chart if some values are zero
**Describe/explain the bug**

In a stacked line graph with points, the top-most point will always be displayed on top (and covering entirely) points below it. This can be misleading if some intermediate values in that graph are 0, and can result in an incorrect reading of the graph, as illustrated in the screenshot below.

**To Reproduce**
I’ve set up a repro on this CodeSandBox: https://codesandbox.io/s/nivostacked-line-misleading-points-ublb7

**Expected behavior**

In a stacked line graph with points enabled, points with a value of 0 should not appear on top of (and covering entirely) points that have a value that is non-0.

I think that giving priority to the bottom-most point will solve this problem.

**Screenshots**
image

In that screenshot, reading the graph makes me think that there was 1 PR merged and 0 PR opened, although the reality (as shown in the tooltip) is the other way around.

 - OS: macOS
 - Browser: Chrome latest
 - Version: 0.62.0

I’m willing to work on a PR if you think this is indeed an issue.

phacks:
Fixes #1051 

Reversing the `lines` and `points` array makes the lines and points belonging to lower series appear on top—and makes the graph less misleading.

See this before/after screenshot comparison (the area of interest is the red and yellow lines at the NG and LS ticks):

**Before**

image
We get the impression that the yellow line is increasing, although its value is 0.

**After**

image
We get the right impression (the red line is increasing).

### DIFF
--- a/packages/line/src/Lines.js
+++ b/packages/line/src/Lines.js
@@ -11,16 +11,18 @@ import PropTypes from 'prop-types'
 import LinesItem from './LinesItem'
 
 const Lines = ({ lines, lineGenerator, lineWidth }) => {
-    return lines.map(({ id, data, color }) => (
-        <LinesItem
-            key={id}
-            id={id}
-            points={data.map(d => d.position)}
-            lineGenerator={lineGenerator}
-            color={color}
-            thickness={lineWidth}
-        />
-    ))
+    return lines
+        .reverse()
+        .map(({ id, data, color }) => (
+            <LinesItem
+                key={id}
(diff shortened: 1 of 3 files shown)


phacks:
@wyze Hello! Just checking in to see whether you had time to check this PR out. No worries if not, and please let me know if you need anything else to make reviewing it easier!

PR #1062, opened 2020-07-15
feat(bullet): support ranges that support < 0

jdjkelly:
- Adds support for < 0 values in bullet graphs by including a value < 0 in the range
- First time committing to this repo, not sure if there's anything else I should do, here's an example where this would be useful (using a bullet graph to display standard deviations or z-scores)

Screen Shot 2020-07-15 at 7 03 04 PM


plouc:
@jdjkelly, thank you for your contribution!

PR #1065, opened 2020-07-21
feat(line): add option to enable tooltip when hovering over points with holes in data

### LINKED ISSUE #975: Enable slice tooltip for holes in data on mouse-hover
Hi, first of all, thank you for this amazing library.
Recently, I've come across the need to show "Unavailable" on the slice tooltip for holes in data.
As holes in data can only be shown with `null` values, it also means that the tooltip will not render null values over a `null` point. 

I've tried using the `sliceTooltip` prop in the `<Line/>` component, but it only allows me to customize the tooltip without the ability to control how it should render over a null x axis value.

Is there a way to do this with nivo?

Something like this : 
image

Thank you!

cheeyang:
This Pull Request closes #975.

- feat(line): add option to enable tooltip when hovering over points with null values

Screenshot 2020-07-21 at 10 51 14 AM

### DIFF
--- a/packages/line/src/Line.js
+++ b/packages/line/src/Line.js
@@ -88,6 +88,7 @@ const Line = props => {
         tooltip,
 
         enableSlices,
+        enableTooltipOnNullValues,
         debugSlices,
         sliceTooltip,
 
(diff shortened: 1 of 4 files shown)


cheeyang:
with debugSlices=true

Before:
Screenshot 2020-07-21 at 11 12 18 AM
After:
Screenshot 2020-07-21 at 11 12 08 AM

Issue #1031, opened 2020-06-25
Legends Disappear


HadiKhai:
Hello, is there a way to show the legends despite the data being 0, for example, if I have the following data for the bar chart: `[{
date: ...,
key1: 0,
key2: 2,
key3: 10
}]
`
Only the legends for keys 2 and 3 will appear while I need all the keys to be shown. Thank you

diegosps:
This might be a duplicate.

sibelius:
is there a workaround for this?

Issue #986, opened 2020-06-07
Values not rendered in bar chart will affect selected colors, legend and maybe other properties


diegosps:
**Describe/explain the bug**
When the height of the chart is not exaggerated, the component will skip the first color from scheme if the key value is so small that cant be rendered.

**To Reproduce**
Both charts with same data and same color scheme. The only difference is the height.
It seems the first color of the set is skiped in the first chart because it is unable to render it. resulting in different set of colors for both charts.

https://codesandbox.io/s/vigilant-lake-ml67g

**Expected behavior**
It shouldn't matter if the first key is small or not, the component should respect the correct order from selected scheme.

Thanks!

**Edit**
After seeing #1031,  I made another example to include the legend component:
https://codesandbox.io/s/adoring-snowflake-2ojr2

Apparently, It happens when the data value is equal to 0 too.

diegosps:
This one has a [simple fix](https://github.com/plouc/nivo/issues/952#issuecomment-655105135). But I'm not sure about side effects.
Will submit a PR when the calendar one gets closed (I guess I need to refork it to keep commit history clean).

wyze:
This appears to be fix with my latest pending PR, you can see a CodeSandbox with that version and your code here: https://codesandbox.io/s/nivo-forked-sivhw?file=/src/App.js

Issue #952, opened 2020-05-12
coloring diverging stacked bar chart


m-jakob93:
When only one data value != 0 is given in the data array for the stacked diverging bar, only the first color is chosen. The array of keys then does not always match the array of colors.

In the sandbox I reproduced the problem:

The keys are: ["pos", "neg"]
The colors are: ["green", "red"]
If the data looks like this [{pos: 1, neg: -1}] all is well, but if the data is [{pos: 0, neg: -1}], the "neg" is colored with the first color value.

[Codesandbox](https://codesandbox.io/s/fast-snow-cpnsj?file=/src/App.js)

OS: macOS Catalina 10.15.4
Browser: Chrome Version 81.0.4044.129

diegosps:
I believe this is the same bug as #986 and #1031 

Maybe close them as duplicate?

Looking the code, a simple fix is to remove all `if (barWidth > 0)` and `if (barHeight > 0)` from `grouped.js` and `stacked.js`

I tested it locally and it worked. Both the colors and legend items.

However, i didn't go further the code to see if this has any impact.

IgorStetsiuk:
Hey, guys! Do you have any updates on this? I've encountered the same issue as @m-jakob93 described, what could I possibly do except changing the source code of the lib? 

For now, I came up with the idea of checking, `data` and then to adjust `colors` and `keys` properties for that `datums` (data points) with that exact colors that I need to display.

I'll post updates here and my case and solution, ones its ready

wyze:
This is fixed in the latest version: https://codesandbox.io/s/wispy-feather-9fnwe

PR #1079, opened 2020-08-02
fix: empty bar values affects rendering (color, legend, possible others)

diegosps:
Removed the `if height/width > 0 ` that seemed to cause strange rendering issues.

This fixes: #1031 #952 #986 

I don't know why it was there and if this will have any side effect.


diegosps:
Sorry. I tried to remove the commits from the previous branch and made it worse lol

PR #1082, opened 2020-08-04
feat(generators): add network data generation

wyze:
Moves this from the website workspace so it can be shared with storybook.


wyze:
@tg970 Unfortunately I can't cut releases, only merge PRs. @plouc is the only one with the keys to release currently.

PR #1089, opened 2020-08-05
fix(radar): typing for custom functions

AlaaZorkane:
- Exported all custom elements.
- Changed `GridLabelCustomFunction` to `CustomGridLabel` (for consistency).
- Made `CustomGridLabel` and `CurstomDotSymbol` function components and typed their props according to docs.
- Made `legends` optional. 
- Fixed typo `IndexByCustomFunction` instead of `IndexByCustomFunctiono`


packages/radar/index.d.ts
```diff
@@ -12,16 +12,28 @@ import { OrdinalColorsInstruction, InheritedColorProp } from '@nivo/colors'
...
+        size: number
+        color: InheritedColorProp
+        borderWidth: number
+        borderColor: InheritedColorProp
+    }
+
+    type IndexByCustomFunction<D = any> = (datum: D) => string | number
+    export type CustomGridLabel = React.FC<GridLabelProps>
+    export type CustomDotSymbol = React.FC<DotSymbolProps>
+    export type CustomDotLabel = (...args: any[]) => React.ReactNode
```
AlaaZorkane, line 30: I think these should be strictly typed too, I never used them and I honestly don't have enough knowledge about them to type them.

packages/radar/index.d.ts
```diff
@@ -55,7 +67,7 @@ declare module '@nivo/radar' {
         isInteractive?: boolean
         tooltipFormat?: string | CustomFormatter
 
-        legends: LegendProps[]
+        legends?: LegendProps[]
```
AlaaZorkane, line 70: I couldn't find anything in the docs about whether the legends are optional or not, but there are some examples in the storybook with no legends (https://nivo.rocks/storybook/?path=/story/radar--custom-label-component) and I am already using them without legends in the project I am working on, actually typescript threw me an error because of this one lol.

wyze: Yes. They are optional.

wyze:
@AlaaZorkane I remember looking through this PR, I guess I didn't do a formal review or comment. Would you mind resolving the conflicts and I can get this merged?

packages/radar/index.d.ts
plouc: I agree, but I guess we can add this later, what you updated is already a good addition.

plouc (approved):
LGTM

wyze:
Thanks @AlaaZorkane!

PR #1097, opened 2020-08-10
feat(scales): Add support for symlog scale

OsQu:
Thank you for the great library!

Lately we needed to render some graphs using logarithmic scale, but found out that `d3-scale`'s logarithmic scale is not very well suited for data having a wider range. For those use cases one can use symlog scale which "offers a bi-symmetric log transformation, suitable ‘for wide-range data’. Its domain can span several orders of magnitude, with negative as well as positive values." `d3-scale` has also API for that.

Add support for [symlog scale](https://github.com/d3/d3-scale#symlog-scales). The code is heavily inspired by a PR adding the support for logarithmic scale: https://github.com/plouc/nivo/pull/378

Compared to logarithmic scale PR, no safe-guarding for zero or negative values is needed since the scale accepts both.

I added relevant stories to storybook but did not include links to website as I thought it's rather similar to logarithmic scale. Users browsing the storybook should find it easily though.

Here is a screenshot of a scatterplot using symlog scale:

image

Thank you in advance, and let me know if you have any suggestions for improvement.

### DIFF
--- a/packages/scales/src/symlogScale.js
+++ b/packages/scales/src/symlogScale.js
@@ -40,7 +40,7 @@ export const symlogScale = (
 }
 
 export const symLogScalePropTypes = {
-    type: PropTypes.oneOf(['symLog']).isRequired,
+    type: PropTypes.oneOf(['symlog']).isRequired,
     constant: PropTypes.number,
     min: PropTypes.oneOfType([PropTypes.oneOf(['auto']), PropTypes.number]),
     max: PropTypes.oneOfType([PropTypes.oneOf(['auto']), PropTypes.number]),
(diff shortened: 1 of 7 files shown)


packages/scales/src/symlogScale.js
```diff
@@ -0,0 +1,47 @@
...
+        .rangeRound(axis === 'x' ? [0, size] : [size, 0])
+        .nice()
+
+    scale.type = 'symlog'
+
+    return scale
+}
+
+export const symLogScalePropTypes = {
+    type: PropTypes.oneOf(['symLog']).isRequired,
```
wyze, line 43:
```suggestion
    type: PropTypes.oneOf(['symlog']).isRequired,
```

wyze (changes requested):
Looks great Oskari! One minor casing issue to avoid PropType warning while using.

wyze (approved):

OsQu:
Thank you Neil! 25 snapshot failures from CI, but looks like that the master branch is failing as well so probably unrelated to this PR.

wyze:
Yup. I have that fixed in #1128. Thanks for the contribution. We will get this out in the next release, which will hopefully be soon. :)

PR #1107, opened 2020-08-20
Passing through the calculated individual bar width and height to layers

Francismb:
**Issue:**
I have three keys in my bar chart(['currentYear', 'oneYearAgo', 'twoYearsAgo']), my custom layer renders a different key('goal'). If the user has no currentYear, oneYearAgo or twoYearsAgo data but has goal data, you cannot use the bars prop as there is no data in it. So I tried using the raw data prop but there was no barWidth prop to allow me to position my SVG layers accurately.

**Solution:**
Passing through the calculated bar height or width(depending on the orientation of the bar chart) to custom layers.

This is to allow you to use the data prop instead of the bars prop(Which could be empty depending on the data -> the issue in my case).

### DIFF
--- a/packages/bar/src/compute/grouped.js
+++ b/packages/bar/src/compute/grouped.js
@@ -113,7 +113,7 @@ export const generateVerticalGroupedBars = ({
         })
     }
 
-    return { xScale, yScale, bars }
+    return { xScale, yScale, bars, barWidth }
 }
 
 /**
@@ -190,7 +190,7 @@ export const generateHorizontalGroupedBars = ({
         })
     }
 
-    return { xScale, yScale, bars }
+    return { xScale, yScale, bars, barHeight }
 }
 
 /**


Francismb:
Not stale.

wyze:
Hi @Francismb, can you put together a CodeSandbox on how you are looking to use this? I'm not quite following your description.

Francismb:
I don't quite have time to whip it up right now at @wyze but basically... If the user has no data in the primary layer of the chart, then a custom layer does not receive the width of the bars through the component props.

wyze:
I'm not quite sure what you are doing, but this should be possible now, without this change. We no longer filter out data that is zero or missing, so you should get full access. I tried to recreate what I think you were describing in the PR description. https://codesandbox.io/s/heuristic-cray-19rg3?file=/src/App.js

Let me know if this solves the issue or what else you might be trying to accomplish.

Francismb:
Hey @wyze That actually looks good now. Thanks for the follow-up, I will close this pr now.

Issue #1101, opened 2020-08-14
`axisBottom.tickValues` does not respect string interval when crossing end of month boundaries


fillippeyton:
**Describe/explain the bug**
I'm running into a strange issue with using intervals in `tickValues` (every 3 days, every 7 days, etc). When the data moves past the end of the month, the tickValue interval is thrown off. As you can see in the screenshot, both 07/31 and 08/1 ticks are shown next to each other and does not hold up to the every 3 days interval. 

![image](https://user-images.githubusercontent.com/4128489/90264491-07305700-de06-11ea-8e0e-09f3aa1a224f.png)

**To Reproduce**
1) Use a line chart that uses a time series for the X Axis 
2) Use a dataset that crosses over the end of a month (i.e. 07-29-2020 through 08-08-2020)
3) Use a `tickValues` interval of "every 3 days", "every 7 days", etc

Reproduced in this code sandbox: https://codesandbox.io/s/brave-brown-9qd33

**Expected behavior**
I would expect the X-Axis to understand the end of the month and continue its interval of three days (or whatever the interval is). In my example, the X-Axis would look like this:

![image](https://user-images.githubusercontent.com/4128489/90265354-50cd7180-de07-11ea-9176-115e6f4b99aa.png)

To get around this for now, I was able to manually assign x-axis tick values: https://codesandbox.io/s/cocky-firefly-tekbg
I'll likely just create these tick values based on my data set and then format my dates via a function in `axisBottom.format`.

**Desktop (please complete the following information):**
 - OS: MacOS Catalina 10.15.6
 - Browser: Chrome
 - Version: 84.0

fillippeyton:
Looks like an issue based on the usage of `d3.ticks` with `d3.timeDay.every(n)`

http://jsfiddle.net/myqhpogs/

```
var x = d3.scaleTime()
    .domain([new Date('2020-08-28'), new Date('2020-09-13')]);

console.log(x.ticks(d3.timeDay.every(5)));

// result
[
  0: Mon Aug 31 2020 00:00:00 GMT-0700 (Pacific Daylight Time) {},
  1: Tue Sep 01 2020 00:00:00 GMT-0700 (Pacific Daylight Time) {},
  2: Sun Sep 06 2020 00:00:00 GMT-0700 (Pacific Daylight Time) {},
  3: Fri Sep 11 2020 00:00:00 GMT-0700 (Pacific Daylight Time) {},
]
```

fillippeyton:
[...]

PR #1132, opened 2020-09-17
feat: add possiblity to remove the first day of month in "every X days"

ItsJimi:
According @fillippeyton [issue#1101](https://github.com/plouc/nivo/issues/1101), I'm trying to resolve this usecase by adding the keywords `, no first day` (Idea for a better keywords?) to `ticksValues` key. I found the solution directly in the [D3 issue#2240](https://github.com/d3/d3/issues/2240#issuecomment-118582280).
### Example:
```json
tickValues: "every 5 days, no first day"
```

### DIFF
--- a/packages/axes/src/compute.js
+++ b/packages/axes/src/compute.js
@@ -74,7 +74,10 @@ const timeByType = {
 }
 
 const timeTypes = Object.keys(timeByType)
-const timeIntervalRegexp = new RegExp(`^every\\s*(\\d+)?\\s*(${timeTypes.join('|')})s?$`, 'i')
+const timeIntervalRegexp = new RegExp(
+    `^every\\s*(\\d+)?\\s*(${timeTypes.join('|')})s?(, no first day)?$`,
+    'i'
+)
 
 const isInteger = value =>
     typeof value === 'number' && isFinite(value) && Math.floor(value) === value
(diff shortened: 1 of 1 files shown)


fillippeyton:
How would this work when the first day _should_ show up? For instance, if the first day is 9/29 and you use a tick value of `tickValues: "every 2 days, no first day"`, would the returned dates be `9/29, 10/1, 10/3, etc`?

ItsJimi:
> How would this work when the first day _should_ show up? For instance, if the first day is 9/29 and you use a tick value of `tickValues: "every 2 days, no first day"`, would the returned dates be `9/29, 10/1, 10/3, etc`?

If the first should be displayed, it is displayed.
I think `, no first day` is not really relevant.

fillippeyton:
> > How would this work when the first day _should_ show up? For instance, if the first day is 9/29 and you use a tick value of `tickValues: "every 2 days, no first day"`, would the returned dates be `9/29, 10/1, 10/3, etc`?
> 
> If the first day should be displayed, it is displayed.
> I think `, no first day` is not really relevant.

Right, but if you have dynamic dates, you can't assume when you need `no first day` or not.

ItsJimi:
> Right, but if you have dynamic dates, you can't assume when you need `no first day` or not.

Maybe i don't understand what is your use case.

With `every 2 days, no first day`:
- If the start date is `9/29`, the result is `9/29, 10/1, 10/3, etc`.
- If the start date is `9/30`, the result is `9/30, 10/2, 10/4, etc`. (We keep 1 day between each dates)

With `every 2 days`:
- If the start date is `9/29`, the result is `9/29, 10/1, 10/3, etc`.
- If the start date is `9/30`, the result is `9/30, 10/1, 10/3, etc`. (We have don't have space between `9/30` and `10/1`)

plouc:
I think that maybe we could tackle this in a more generic way, several users also requested skipping ticks for other data types, exposing a filtering function for ticks could allow to this while keeping generic ticks generation, then users could filter out or even add ticks.

fillippeyton:
bump

wyze:
Going to go with the implementation in #1565.

PR #1135, opened 2020-09-23
fix(core): Add missing Theme types to match default theme object

spinloop:
Updates the Theme type to match the default theme object here:
https://github.com/plouc/nivo/blob/c443bc08b3abc8fa316e2ee78a6cc7dcded4c968/packages/core/src/theming/defaultTheme.js#L9

Otherwise passing the default theme object to the theme prop of ResponsiveScatterPlot results in a typescript error.

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -44,6 +44,9 @@ declare module '@nivo/core' {
             }>
         }>
         background: string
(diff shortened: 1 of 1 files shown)


wyze (approved):
Awesome, thanks for this!

PR #1143, opened 2020-09-30
refactor(pie): remove recompose and convert to hooks

DevelopmentByDavid:
Addresses #884 in the pie package by removing recompose.

A few notes/questions:
- I am not sure what to do with the `enhance.js` file.  It seems unnecessary since there are now hooks, but since I am unfamiliar with the project I left it there for now pending your advice.
- In `PieLayout.js` I combined two of the computed props.  Color is now always recomputed along with arcs. I don't think this combination changes anything since they seemed interdependent before, but I thought I should bring this change to your attention.
- The one test passes and stories _look_ normal, unless I missed something.
- All compose `withPropsOnChange` were converted to `React.useMemo`

Any other additions or changes I'd be happy to make, just let me know.

Thanks!


packages/pie/src/PieLayout.js
```diff
@@ -94,103 +22,145 @@ export const PieLayoutDefaultProps = {
...
+        })
+    ).isRequired,
+    width: PropTypes.number.isRequired,
+    height: PropTypes.number.isRequired,
+    fit: PropTypes.bool.isRequired,
+    sortByValue: PropTypes.bool.isRequired,
+    startAngle: PropTypes.number.isRequired,
+    endAngle: PropTypes.number.isRequired,
+    padAngle: PropTypes.number.isRequired,
+    // arcs: PropTypes.array.isRequired, // computed
```
wyze, line 146: What was the reason for commenting these prop types out?

wyze (changes requested):
Awesome work and thank you so much! A couple of questions and one change, but it looks great.

wyze:
Also, go ahead and delete `enhance` file since it is no longer used.

packages/pie/src/PieLegends.js
```diff
@@ -6,53 +6,43 @@
...
+            {...legend}
+            containerWidth={width}
+            containerHeight={height}
+            data={data}
+            theme={theme}
+        />
+    ))
+}
+
+PieLegends.defaultProps = {
```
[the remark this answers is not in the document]
DevelopmentByDavid: Good catch!

packages/pie/src/Pie.js
```diff
@@ -1,24 +1,19 @@
...
- * For the full copyright and license information, please view the LICENSE
- * file that was distributed with this source code.
- */
-import React, { Component } from 'react'
-import compose from 'recompose/compose'
-import defaultProps from 'recompose/defaultProps'
-import pure from 'recompose/pure'
-import withPropsOnChange from 'recompose/withPropsOnChange'
-import setDisplayName from 'recompose/setDisplayName'
+// /*
```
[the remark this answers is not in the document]
DevelopmentByDavid: Ah, this was because of my method for conversion. I commented out the original component (control + a, then control + /) and remade it piece by piece. Fixed it now.

packages/pie/src/PieLayout.js
DevelopmentByDavid: All of the props with the `// computed` comment, from my understanding, were computed using other props.  `Arcs`, for example, was always recomputed if any one of the `pie`, `data`, or `getColor` props changed.  

I commented them out because they no longer need to be pipe-lined in as props and can just be computed using `React.useMemo` in the body of the function.  However, your question made me realize something.  Is there some sort of override feature that would be removed as a result of removing these props?  Previously, I think it wasn't possible to override since the `// computed` props were re-computed whenever the props previously mentioned were changed, but I'm not 100% sure.  

Also, when I double checked, one of the props,`innerRadius`, I commented out was `// re-computed` so I un-commented that prop since it should actually be passed in by the user.

packages/pie/src/PieLegends.js
```diff
@@ -6,53 +6,43 @@
...
+            theme={theme}
+        />
+    ))
+}
+
+PieLegends.propTypes = {
+    width: PropTypes.number.isRequired,
+    height: PropTypes.number.isRequired,
+    arcs: PropTypes.arrayOf(arcPropType).isRequired,
+    data: PropTypes.arrayOf(PropTypes.object).isRequired,
```
wyze, line 45: We can remove this now since it is computed. It is throwing a warning in the console.

```suggestion
```

packages/pie/src/Pie.js
```diff
@@ -28,191 +23,183 @@ import PieSlicesLabels from './PieSlicesLabels'
...
+        onMouseEnter,
+        onMouseLeave,
+        tooltipFormat,
+        tooltip,
+
+        legends,
+        role,
+    } = props
+
+    const theme = usePartialTheme()
```
wyze, line 85: We can pass a `theme` prop to the component, so we need to override the default theme with those values. You can see this in the `custom tooltip` story. The background of the tooltip should be black and it is white.

wyze (changes requested):
Couple minor things and should be good!

packages/pie/src/Pie.js
DevelopmentByDavid: I fixed this.  I also noticed that in `props.js` that theme isn't a required or optional prop listed in the `propTypes` object.  I don't know if this was done intentionally, but I wanted to bring it up.

wyze: It looks like it is present on PieLegends, might as well make it available for the main components as well.

DevelopmentByDavid: Hmm, where do the prop types for the pie specific theme reside?  I looked in `nivo/packages/core/theming/propTypes.js` (since that is where PieLegends gets the propTypes declaration from), but it doesn't have the `tooltip` property that I see is being used inside of the stories.  It looks like `theme.tooltip` is only ever defined via prop types in `PieTooltip.js`.  I could of course do something like:
```js
import { themePropType } from '@nivo/core'
export const PiePropTypes = {
    // other prop types 
    theme: PropTypes.shape({ ...themePropType, tooltip: PropTypes.shape({}) })
}
```
However, I'm not sure if that encompasses all of the custom theming for the pie package.


Maybe this is starting to get outside of the scope of this PR.

wyze (approved):
Looks great, thanks for the work put in here @DevelopmentByDavid!

PR #1151, opened 2020-10-08
fix(core / swarmplot): Improve core and swarmplot typedefs

areese159:
When implementing custom layers I noticed that the Pattern helpers were not being exported in the type def which caused us to have to use `@ts-ignore` in our code base. I also noticed when working with swarmplots the type def also appeared to be slightly off. I propose two changes in this PR, the first being to export the pattern helpers in the type definition of @nivo/core and the second to give layers a real definition instead of `any[]` in `@nivo/swarmplot`. If anything is not correct please let me know and I will be more than happy to fix it.

### DIFF
--- a/packages/swarmplot/index.d.ts
+++ b/packages/swarmplot/index.d.ts
@@ -26,6 +26,22 @@ declare module '@nivo/swarmplot' {
         data: Datum
     }
 
+    export interface LayerProps<Datum> {
+        nodes: ComputedNode<Datum>[]
+        xScale: (input: number) => number
+        yScale: (input: number) => number
+        innerWidth: number
+        innerHeight: number
+        outerWidth: number
+        outerHeight: number
+        margin: number
+        getBorderColor: () => string
+        getBorderWidth: () => number
+        animate: boolean
+        motionStiffness: number
+        motionDamping: number
+    }
+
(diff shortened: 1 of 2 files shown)


packages/swarmplot/index.d.ts
```diff
@@ -40,6 +56,8 @@ declare module '@nivo/swarmplot' {
         event: React.MouseEvent<any>
     ) => void
 
+    export type Layers<Datum> = ((props: LayerProps<Datum>) => JSX.Element) | string
```
wyze, line 59: Can we make this consistent with other packages like `bar` or `line`. Change from string to a union of possible layers.

wyze (changes requested):
Looks great! Just want to tighten up the layer types a bit.

packages/swarmplot/index.d.ts
areese159: Of course, ill get those changes pushed up tomorrow.

areese159:
I think I have addressed your feedback, let me know if you want anything else changed.

areese159:
Sorry, I made a typo, I've amended the commit and pushed the change. This should be good for rereview.

wyze (approved):
Looks good, thank you!

Issue #1150, opened 2020-10-08
Parentheses in Sankey node ids break link colors


JonathanPicques:
**Describe/explain the bug**
Using parentheses in Sankey node ids break links colors (almost identical to https://github.com/plouc/nivo/issues/623)

**To Reproduce**
https://codesandbox.io/s/awesome-sunset-kkhyn

1. Remove parentheses line 20, 25 & 26: the colors work
2. Add back parentheses line 20, 25 & 26: the colors won't work anymore

**Expected behavior**
Custom colors should continue to render even when using parentheses in ids

**Screenshots**
Working without parentheses
![image](https://user-images.githubusercontent.com/1839757/95477091-42738000-0988-11eb-838a-651f6fd9605e.png)
Broken with parentheses
![image](https://user-images.githubusercontent.com/1839757/95476971-2839a200-0988-11eb-9a73-251a67cd7a31.png)

wyze:
Feel free to send a PR if you would like!

JonathanPicques:
@wyze would love too! is there somewhere a guide to start contributing? never used lerna so I'm a bit lost

wyze:
We have the https://github.com/plouc/nivo/blob/master/CONTRIBUTING.md you can follow, but it should be a few steps to get started. Clone/fork the repo, then run `make init` then grab a coffee, hah!, then `make storybook` and it will launch storybook in dev mode. Then feel free to make your changes and Storybook should reflect them automatically.

PR #1152, opened 2020-10-09
fix(sankey): Fix issue with gradient and parentheses in IDs

JonathanPicques:
Closes https://github.com/plouc/nivo/issues/1150


JonathanPicques:
It should work with ~everything (tested with ```' " ` @ / \```)

wyze:
Thanks for the quick fix!

PR #1161, opened 2020-10-15
Storybook 6.0.X Upgrade

jamesdh:
Upgrades to the latest version of Storybook, and includes the necessary changes to the new configuration formats [as described here](https://github.com/storybookjs/storybook/blob/master/MIGRATION.md#from-version-52x-to-53x).

Also includes some small changes to storybook names for the Bar and Funnel stories to fix a strange hierarchy/navigational quirk that came about from the upgrade.


package.json
```diff
@@ -37,6 +39,7 @@
     "babel-preset-react-app": "^9.1.2",
     "chalk": "^2.4.2",
     "clog-cli": "^1.0.0",
+    "core-js": "^3.0.1",
```
jamesdh, line 42: Needed to add this due to https://github.com/storybookjs/storybook/blob/master/MIGRATION.md#core-js-dependency-errors

wyze: Would yarn resolutions work for this instead of the explicit dev-dependency? https://classic.yarnpkg.com/en/docs/selective-version-resolutions

jamesdh: [Looks like it should be possible](https://github.com/storybookjs/storybook/issues/11255#issuecomment-673899817), although kind of ugly. 

I thought it probably didn't matter much since it's just a devDependency?

wyze:
@jamesdh Thanks for this. Are you able to resolve the conflicts? I did locally, but you don't allow maintainers to push to your fork. These changes will fix storybook for the release I just put out.

Thanks!

wyze:
I merged this in #1177 and got it deployed to https://nivo.rocks/storybook/. Thanks so much!

PR #1163, opened 2020-10-17
fix(bar): Fix BarItemProps types

### LINKED ISSUE #1155: Need for casting types over an over again in barComponent callback (TS)
**Describe/explain the bug**
I am working on adding tooltip functionality to my custom barComponent (in TypeScript).
I realized that `showTooltip`, `onMouseEnter`, `onMouseLeave` and others won’t accept the event parameter with type
 
`React.MouseEvent<SVGRectElement, MouseEvent>`  

because I think `BarMouseEventHandler` is a generic type with a default value `HTMLCanvasElement` but the generic type was not passed down at `onMouseEnter: BarMouseEventHandler` in `BarItemProps` - `BarItemProps` is not even generic so I can pass it in manually - right here:
 https://github.com/plouc/nivo/blob/52c1bc155b58c522c8dc3ca3d43385630b0c563c/packages/bar/index.d.ts#L54-L87

Also, this is `showTooltip` event is hardcoded: https://github.com/plouc/nivo/blob/52c1bc155b58c522c8dc3ca3d43385630b0c563c/packages/bar/index.d.ts#L83

And that causes to write this kind of code in TypeScript where I need to cast the types over and over again:
```
const handleMouseEnter = (
    e: React.MouseEvent<SVGRectElement, MouseEvent>,
  ) => {
    onMouseEnter(
      data as BarExtendedDatum,
      (e as unknown) as React.MouseEvent<HTMLCanvasElement, MouseEvent>,
    )
    if (tooltip) {
      showTooltip(
        tooltip(data as BarExtendedDatum),
        (e as unknown) as React.MouseEvent<HTMLCanvasElement, MouseEvent>,
      )
    }
  }
```

As you see I also need to type cast the data, since the callback functions are expecting (and what actually provided `data` prop is) `BarExtendedDatum`, however the provided `data` props is typed as:
```
var data: {
    id: string | number;
    value: number;
    indexValue: string | number;
}
```

**To Reproduce**
Check out the sandbox and try deleting types in `char-enhancers.tsx` to observe the errors.
https://codesandbox.io/s/horizontal-bar-chart-xuuy9?file=/src/chart-enhancers.tsx

eggei:
Fixes #1155 

**Changes:**

- Change `data` prop type to `BarExtendedDatum`
- Remove `color` field in `data` prop's type - it doesn't exist in `data` prop
- Create `BarTooltipDatum` to pass `color` type to tooltip props along with
  `BarExtendedDatum`
- Add `SVGRectElement` to `BarMouseEventHandler` generic type
- Update deprecated `React.StatelessComponent` to `React.FC`

### DIFF
--- a/packages/bar/index.d.ts
+++ b/packages/bar/index.d.ts
@@ -151,6 +151,7 @@ declare module '@nivo/bar' {
             onClick: BarMouseEventHandler<SVGRectElement>
             onMouseEnter: BarMouseEventHandler<SVGRectElement>
             onMouseLeave: BarMouseEventHandler<SVGRectElement>
+            role: string
         }>
 
     export class Bar extends React.Component<BarSvgProps & Dimensions> {}
--- a/packages/bar/index.d.ts
+++ b/packages/bar/index.d.ts
@@ -39,7 +39,6 @@ declare module '@nivo/bar' {
         value: number
         index: number
         indexValue: Value
-        color: string
         data: BarDatum
     }
 
@@ -51,19 +50,18 @@ declare module '@nivo/bar' {
 
     export type ValueFormatter = (value: number) => string | number
 
-    export type BarMouseEventHandler<T = HTMLCanvasElement> = (
+    type GraphicsContainer = HTMLCanvasElement | SVGRectElement
+
+    export type BarMouseEventHandler<T = GraphicsContainer> = (
         datum: BarExtendedDatum,
         event: React.MouseEvent<T>
     ) => void
 
-    export type TooltipProp = React.StatelessComponent<BarExtendedDatum>
+    export type BarTooltipDatum = BarExtendedDatum & { color: string }
+    export type TooltipProp = React.FC<BarTooltipDatum>
 
     export interface BarItemProps {
-        data: {
-            id: string | number
-            value: number
-            indexValue: string | number
-        }
+        data: BarExtendedDatum
         x: number
         y: number
         width: number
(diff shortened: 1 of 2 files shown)


eggei:
Hit @wyze - I created this PR after our conversation on the Discord channel.  I can't request reviews for some reason, so I thought I can mention you here to let you know about this. Thank you!

packages/swarmplot/index.d.ts
```diff
@@ -47,7 +47,7 @@ declare module '@nivo/swarmplot' {
         Axes = 'axes',
         Nodes = 'nodes',
         Mesh = 'mesh',
-        Annotations = 'annotations'
+        Annotations = 'annotations',
```
eggei, line 50: This got fixed by the linter checks, so I let it in with my comment.

wyze: Thanks, I think this is what is failing the build.

packages/bar/index.d.ts
```diff
@@ -153,7 +151,6 @@ declare module '@nivo/bar' {
             onClick: BarMouseEventHandler<SVGRectElement>
             onMouseEnter: BarMouseEventHandler<SVGRectElement>
             onMouseLeave: BarMouseEventHandler<SVGRectElement>
-            role: string
```
wyze, line 156: I think you need to revert this deletion, but everything else looks good.

wyze (changes requested):

packages/bar/index.d.ts
eggei: Hmm, this might be my first accidental deletion. + I committed it 😅 shameful.
I am sorry for that, thanks for reviewing - I will resolve this soon

wyze: No worries and no need to apologize! 😀

wyze (approved):

wyze:
Thanks!

PR #880, opened 2020-02-24
feat(sunburst): add mouse events and some labels

juliankrieger:
Because of @mjsarfatti 's absence, I have continued to implement what you recommended @plouc. 

Please review the changes and tell me what you think

-----

- Add slice labels at the first level
- Add onClick/onMouseEnter/onMouseLeave events

PR #1165, opened 2020-10-19
Feat enhanced sunburst

epentin:
This updates #880 

Adds support for onClick to sunburst


epentin:
If there is anything else I can do do to help get this pushed please let me know

PR #1168, opened 2020-10-20
build(lodash): add back babel lodash plugin

### LINKED ISSUE #1167: [v0.63.0] Radar chart bundles the entire lodash package
**Describe/explain the bug**
`lodash.js` is being bundled (most likely unnecessarily) while importing `@nivo/radar`

**To Reproduce**
I don't think I can showcase this issue using CodeSandbox, yet I have attached the screenshots of 2 treemaps generated by `source-map-explorer` for your reference.

1. Import `Radar` or `ResponsiveRadar` from  `@nivo/radar`.
2. Compare the bundle sizes and treemaps of version `0.62.0` and `0.63.0`.

**Expected behavior**
`lodash.js` should not be bundled.

**Screenshots**
`0.62.0`
image

`0.63.0`
image

wyze:
I see we had trouble with this in the past where an import like `import {..} from 'lodash'` sneaks in and includes the whole library. It appears we used to have this plugin a few years ago (https://github.com/plouc/nivo/commit/c172a40b963946178200be71da45162c22a2f0d5) but was lost when the multi package migration was started (https://github.com/plouc/nivo/commit/158a349d2ba8e9e017486e32fc89baa4e5c0c0a3#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519).

I've tested this and it works with both import syntaxes we use for lodash.

Close #1167


plouc:
Thank you @wyze! I'll probably create a release in the upcoming days as it makes a huge difference regarding bundle size.

PR #1172, opened 2020-10-21
Fix react-spring handling animating paths

### LINKED ISSUE #1170: Unhandled error after several re-renders (going from 0.62.0 to 0.63.0)
This issue didn't exist in 0.62, only noticed after upgrading to 0.63, and it goes away after downgrading:

image

Changing the data attribute several times seems to produce this issue, but the exception is deep down the stack.

Please let me know if this issue isn't an obvious fix and I will create a codesandbox to reproduce it.

### LINKED ISSUE #1171: Line chart documentation breaks when choosing non-linear curve style on SVG
When selecting a value for the curve parameter in the [Line Documentation](https://nivo.rocks/line/) the page fails to render. 

Console shows: TypeError: n[t] is undefied

1. Go to https://nivo.rocks/line/
2. Click on linear under Style for the curve interpolation
3. Select any other curve style

Desktop
 - OS: WIndows 10
 - Browser Firefox
 - Version 81.0.2

wyze:
Animating paths don't work well with `react-spring` as they did previous with `react-motion`. I've created a custom hook that can be used to animate the paths properly.

This might be charts that support animating `curve` prop that crash, others might work correctly. The when trying to animate a curve, `react-spring` receives different size paths and tries to split strings and crashes because the lengths don't match.

I've added a list of charts below that contain `animated.path` component and will check each individually and either push fixes or check the box to ensure it is working properly.

Close #1170 
Close #1171

------

TODO:

- [x] annotations
- [x] bump
- [x] funnel
- [x] line
- [x] parallel-coordinates
- [x] radar
- [x] sankey
- [x] stream

### DIFF
--- a/packages/annotations/src/AnnotationLink.js
+++ b/packages/annotations/src/AnnotationLink.js
@@ -8,23 +8,15 @@
  */
 import React, { memo } from 'react'
 import PropTypes from 'prop-types'
-import { useSpring, animated } from 'react-spring'
-import { useTheme, useMotionConfig } from '@nivo/core'
+import { animated } from 'react-spring'
+import { useAnimatedPath, useTheme } from '@nivo/core'
 
-const AnnotationLink = memo(({ points, isOutline }) => {
+const AnnotationLink = memo(({ isOutline, ...props }) => {
     const theme = useTheme()
+    const [point, ...points] = props.points
 
-    let path = `M${points[0][0]},${points[0][1]}`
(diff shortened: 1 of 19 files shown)


wyze:
I've verified each of the packages above no longer crash on the website/storybook, so this should be good to go.

I'd prefer to rebase and merge this instead of squash and merge to carry over the history to master.

PR #1175, opened 2020-10-22
fix(ci): switch to GitHub actions

wyze:
It appears our Travis builds stay queued for a while, lets try GitHub actions?


coooraph:
Hey, yes, why not, also it's nice that it's integrated.

PR #1185, opened 2020-10-25
fix(tooltip): Show tooltip on first tap for touch devices

### LINKED ISSUE #1184: Tooltips for chart types that support animated tooltips aren't shown on touch devices (phones, tablets, etc) on the first touch
**Describe/explain the bug**
When using a chart type that supports animated tooltip transitions (line, bump, sankey, etc), the tooltip does not show when any data point is tapped for the first time on a touch device. Subsequent taps will show the tooltip, but once you tap away so no tooltip is shown, you must again tap twice to show a tooltip. I believe this is only noticeable on touch devices because the event to show the tooltip only fires once while the event fires continuously on desktop while your mouse moves.

**To Reproduce**
1. Go to any chart on the docs site or storybook site that supports animated tooltip transititons, for example https://nivo.rocks/line/. Either visit the site on your phone or use the device emulator settings in Chrome dev tools to emulate a phone or a tablet.
2. Tap on any data point and notice that no tooltip is shown. Tap again and notice that it is shown.

**Expected behavior**
Tooltips should be shown on the first tap.

aschey:
This closes #1184 - tooltips not being shown the first time a data point is tapped on a touch device.

The issue seemed to stem from how the code was setting the transform values. On the first event, `hasDimension` was false so the transform value wasn't being set. This change sets the transform value the first time the event is fired so it displays correctly. I also got rid of the opacity property since I didn't see a scenario where it should be set to 0 after this change, but let me know if I missed anything and I can add that back.

I did go through every chart type in the storybook and all of the tooltips seemed to render and hide themselves appropriately.

### DIFF
--- a/packages/tooltip/src/components/TooltipWrapper.js
+++ b/packages/tooltip/src/components/TooltipWrapper.js
@@ -21,6 +21,8 @@ const tooltipStyle = {
     left: 0,
 }
 
+const translate = (x, y) => `translate(${x}px, ${y}px)`
+
 const TooltipWrapper = ({ position, anchor, children }) => {
     const theme = useTheme()
     const { animate, config: springConfig } = useMotionConfig()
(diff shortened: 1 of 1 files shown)


wyze:
Awesome, thank you!

PR #1197, opened 2020-10-29
fix(bar): include bars with zero height/width

### LINKED ISSUE #957: Tiny BarItem not being rendered.
**Describe/explain the bug**
Can we remove this line `if (barWidth > 0) {`? https://github.com/plouc/nivo/blob/f1bff166d0ca4a1ba12b9d54420e2983167f3854/packages/bar/src/compute/stacked.js#L197
Then we can control ourself in `barComponent` property if we want to have minimum bar width.
I have the issue where one bar has value 1 and another very high like 400. In smaller screen like mobile it becomes small so the bar isn't shown at all due to the if-statement above which makes the user think that the bar is 0 instead of 1.

**To Reproduce**
https://codesandbox.io/s/bar-custom-color-9jntn

**Expected behavior**
I should be able to see label of the small numbers also, not only large label numbers

### LINKED ISSUE #986: Values not rendered in bar chart will affect selected colors, legend and maybe other properties

### LINKED ISSUE #1031: Legends Disappear

wyze:
I know this is a hot topic and I know several other contributors made PRs adding options to change this behavior, but I think we should not be filtering out bars which would otherwise not be rendered.

I've put together a CodeSandbox comparing with 4 other charting libraries which all do not have this behavior: https://codesandbox.io/s/epic-sanne-hf3ef?file=/src/App.js

Also, I'm not sure I see the difference if the bar is not included versus not rendered. The bandwidth will remain the same since it is calculated over the entire data set, before we calculate the individual bars, so I don't see the harm in keeping this in the data.

@plouc would love to hear your thoughts on this.

Close #720 
Close #1061 
Close #1057 
Close #454
Close #591
Close #957
Close #1031
Close #986

### DIFF
--- a/packages/bar/src/compute/grouped.js
+++ b/packages/bar/src/compute/grouped.js
@@ -68,7 +68,7 @@ const generateVerticalGroupedBars = (
                 }
             })
         )
-    ).filter(bar => gt(bar.width, 0) && gt(bar.height, 0))
+    )
 
     return bars
 }
@@ -124,7 +124,7 @@ const generateHorizontalGroupedBars = (
                 }
             })
         )
-    ).filter(bar => gt(bar.width, 0))
+    )
 
     return bars
 }
--- a/packages/bar/src/compute/stacked.js
+++ b/packages/bar/src/compute/stacked.js
@@ -71,7 +71,7 @@ const generateVerticalStackedBars = (
                 }
             })
         )
-    ).filter(bar => bar.height > 0)
+    )
 
     return bars
 }
@@ -127,7 +127,7 @@ const generateHorizontalStackedBars = (
                 }
             })
         )
-    ).filter(bar => bar.width > 0)
+    )
 
     return bars
 }


wyze:
Wow, didn't realize CodeSandbox CI built for other links it finds in description and issues. Looks like I forgot some margin in the example link I posted in the description, so here is a forked version of `hopeful-bouman-2u69s` with margin added.

https://codesandbox.io/s/gallant-silence-zrwdi

You can see now, we are consistent with the other libraries.

plouc:
@wyze, really nice to have the comparison with other libs on codesandbox, that's really useful! I think your approach is good, legends will show up and we can see the label if we want.

plouc (approved):
LGTM

PR #1199, opened 2020-10-30
chore(ci): add website to codesandbox ci

wyze:
Now we have an additional sandbox built with all PRs that will build the website as well.


plouc:
@wyze, thank you for this! I didn't consider this and I'm quite impressed that codesandox is able to run Gatsby. Also the `ci.json` file is gonna be easier to maintain with `packages/*` 👍 .

wyze:
Yes, now only if storybook was as easy. Still on the hunt for a solution. :)

PR #1210, opened 2020-11-03
Fixes #1209 by passing child's node data to getChildColor

### LINKED ISSUE #1209: Sunburst children get parent's data in childColor function instead of their own
When trying to make sunburst children data dependent on their value, I came across an issue where parent's data node is passed to the function instead of child's (parent's data is accessible from the child's "ancestor" property), which prevents accessing child's node data:

```
<ResponsiveSunburst
	data={data}
	margin={{ top: 10, right: 0, bottom: 10, left: 0 }}
	borderWidth={1}
	borderColor="white"
	colors={['#61cdbb', 'rgb(232, 168, 56)']}
	childColor={(d) => {
		console.warn(d); // this will print parent's data, instead of child's
	}}
	animate
	motionStiffness={90}
	motionDamping={15}
	isInteractive
/>
```

Code in question - nivo/packages/sunburst/src/Sunburst.js, around line 195:

```
} else if (node.depth > 1) {
        node.data.color = getChildColor(node.parent.data)  // should be node.data
}
```

I'll be sending a simple PR shortly.

Thanks for an absolutely amazing library!

leobudima:
fix #1209

### DIFF
--- a/packages/sunburst/src/Sunburst.js
+++ b/packages/sunburst/src/Sunburst.js
@@ -193,7 +193,7 @@ const enhance = compose(
                 if (node.depth === 1 || childColor === 'noinherit') {
                     node.data.color = getColor(node.data)
                 } else if (node.depth > 1) {
-                    node.data.color = getChildColor(node.parent.data)
+                    node.data.color = getChildColor(node.data)
                 }
             })
 


wyze:
I don't think this does what you want. Here is the default sunburst after applying the change:

image

plouc:
@leobudima thank you for this PR, I'm closing it as we changed this package a lot, sorry for that, also as @wyze said, it's gonna somehow break the chart, if you want to set the color according to the node itself, you should pass a function to the `colors` property.

onionhammer:
@plouc 'childColors' should definitely receive the child as part of the function call; this could easily be added as a second argument to the function without breaking existing users

wyze:
@onionhammer You have full control over the colors by passing `inheritColorFromParent={false}` and a custom `colors` function. See https://codesandbox.io/s/compassionate-night-x9ys6?file=/src/SunburstChart.js

PR #1249, opened 2020-11-06
fix(pie): fix incomplete type for valueFormat

plouc:


wyze (approved):
Thanks, this will fix me up!

PR #1259, opened 2020-11-07
chore(lint): switch from tslint to eslint

wyze:
Upgrade to latest ESLint and install the typescript parser for ESLint. Using `eslint-config-react-app` so that is where all the plugins are coming from.

Removed `fmt-check` from CI as it is included in linting now.


.github/workflows/ci.yml
```diff
@@ -34,7 +34,5 @@ jobs:
       - run: yarn install --frozen-lockfile
       - run: make bootstrap
       - run: make packages-build
-      - run: make fmt-check
```
plouc, line 37: `eslint-config-prettier` purpose is:

> Turns off all rules that are unnecessary or might conflict with Prettier.

So I think we still need to run `fmt-check` which serves a different purpose.

Makefile
```diff
@@ -117,37 +117,7 @@ package-lint-%: ##@1 packages run eslint on package
 
 packages-lint: ##@1 packages run eslint on all packages
 	@echo "${YELLOW}Running eslint on all packages${RESET}"
-	@./node_modules/.bin/eslint "./packages/*/{src,tests}/**/*.js"
```
plouc, line 120: nice ;)

plouc (changes requested):

.github/workflows/ci.yml
wyze: Are you okay adding eslint-plugin-prettier? I think I meant to do that but missed it.

plouc: Oh, I wasn't aware of this plugin, I think it's good to be able to run everything with one command, I just hope we'll understand clearly when it's just about re-running prettier VS issues which have to be fixed manually. Let's add it and see how it goes.

plouc (approved):
LGTM

PR #1261, opened 2020-11-09
Update exports in Bar package to include the BarItem

anirudh-0:
#### Description

Add additional export of the default BarItem so that it can be used as a base BarComponent.

#### Use case

The need arose out of wanting to overlay text around the default BarComponent without having to rewrite the whole thing on my own.

Please consider this small change.

### DIFF
--- a/packages/bar/src/index.js
+++ b/packages/bar/src/index.js
@@ -7,6 +7,7 @@
  * file that was distributed with this source code.
  */
 export { default as Bar } from './Bar'
+export { default as BarItem } from './BarItem'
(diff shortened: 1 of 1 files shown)


wyze (approved):
This looks fine to me. @plouc, what do you think?

plouc (approved):
LGTM

plouc:
This looks good to me, the TypeScript definition is missing though, but we can tackle this when we migrate the package to TypeScript.

Thank you for your contribution @anirudh-0!

PR #1278, opened 2020-11-13
fix(changelog): update clog to allow setting a tag date

### LINKED ISSUE #1277: All release dates in CHANGELOG are the same
**Describe/explain the bug**
The dates for all releases in [CHANGELOG.md](https://github.com/plouc/nivo/blob/ea1ff453e30d9890a858b39e90026d7cf1d80b8a/CHANGELOG.md) appear as "2020-11-06" instead of the actual dates.

wyze:
Close #1277 

-----

If you check the referenced ticket, I point to the code where the date is always set to the current date. I pulled down the code and made the changes needed to support the date.

Looks like this also changed the order of Features/Bug Fixes sections and added colons to the component names.

I've updated this to also run on publish so it should get generated when a new version is released.

Alternatively, we could look at something else to generate these like https://github.com/conventional-changelog/conventional-changelog, but the output might not be exactly the same. Seems clog hasn't received updates in some time.


plouc:
Yes, this issue is annoying, I don't think we should include the tool directly into the repo though, I'd prefer to find another tool, even if the output is slightly different.

wyze:
Okay, that works for me. I will evaluate a few and see which ones work best and update the PR, maybe with a few samples.

wyze:
@plouc Should be good to go now. I'd like to squash this into a single commit on merge.

scripts/generate-changelog.js
```diff
@@ -1,21 +1,28 @@
 'use strict'
 
-const exec = require('child_process').exec
+const { exec } = require('child_process')
 
 const fetchTags = () =>
```
plouc, line 5: I think we could add some comments about how it works/what's happening here, it's a bit difficult to understand, for example explaining the format used for git log and how it's used in the regexp.

scripts/generate-changelog.js
```diff
@@ -26,8 +33,8 @@ const changes = ({ version, from, to }) =>
 const run = async () => {
```
plouc, line 33: Same here, I know it was already undocumented, but it's not straightforward, for example by defining the structure of `tags`, what do you think?

plouc (changes requested):

plouc (approved):
LGTM

PR #1279, opened 2020-11-14
@nivo/tooltip TypeScript migration

plouc:
Also changes a few things:
- Support for render props for tooltip management moved in a deprecated `LegacyContainer` component, we should use the hooks for this and we could potentially remove the dependency to `@nivo/tooltip` if we can get rid of it.
- `withContainer` HOC now use `Container` internally instead of duplicating its logic
- Added a more generic provider for the tooltip `TooltipProvider` which uses 2 different contexts internally for optimization reason and memoize context values for actions
- Added a `Tooltip` component which allows to handle the tooltip state in a generic way
- Improve `ConditionalWrapper`, cloning only the wrapper and using an element instead of a callback function

I think we shouldn't deprecate the `withContainer` HOC, as it allows us to access contexts/hooks in top level components such as `useTheme`/`useTooltip`/`useMotionConfig`.


packages/core/src/components/ConditionalWrapper.js
```diff
@@ -7,13 +7,14 @@ import PropTypes from 'prop-types'
 //     wrapper: (children: JSX.Element) => JSX.Element
 //   }
 
-const ConditionalWrapper = ({ children, condition, wrapper }) =>
-    condition ? cloneElement(wrapper(children)) : children
+export const ConditionalWrapper = ({ children, condition, wrapper }) => {
+    if (!condition) return children
+
+    return cloneElement(wrapper, {}, children)
```
wyze, line 13: The only downside to this is now you can't control the location of an element without a custom component, but also I don't know if we will be using this component that much.

Example:

```jsx
<ConditionalWrapper condition wrapper={<div><label>Example</label></div>}>
  <input type="text" />
</ConditionalWrapper>
```

In the above, the children prop will go to the div, when we want it to be a child of the label. Could make a custom component that receives the children prop to solve it. Like I said, not sure if we will be using this component extensively to worry about it.

packages/tooltip/src/Crosshair.tsx
```diff
@@ -52,15 +51,3 @@ const Crosshair = memo(({ width, height, type, x, y }) => {
...
 Crosshair.displayName = 'Crosshair'
-Crosshair.propTypes = {
-    width: PropTypes.number.isRequired,
-    height: PropTypes.number.isRequired,
-    x: PropTypes.number.isRequired,
-    y: PropTypes.number.isRequired,
-    type: crosshairPropTypes.type.isRequired,
-}
-Crosshair.defaultProps = {
-    type: 'cross',
```
wyze, line 63: Was this lost during the conversion? I don't see a default value for `type` above.

packages/tooltip/src/hooks.ts
```diff
@@ -0,0 +1,81 @@
...
+    )
+
+    const showTooltipFromEvent: TooltipActionsContextData['showTooltipFromEvent'] = useCallback(
+        (content: JSX.Element, event: MouseEvent, anchor: TooltipAnchor = 'top') => {
+            const bounds = container.current.getBoundingClientRect()
+            const x = event.clientX - bounds.left
+            const y = event.clientY - bounds.top
+
+            if (anchor === 'left' || anchor === 'right') {
+                if (x < bounds.width / 2) anchor = 'right'
```
wyze, line 33: I'd prefer if we didn't reassign props here, but make a new variable. Have the prop aliased to `_anchor` then an `anchor` variable in the function.

packages/tooltip/src/hooks.ts
```diff
@@ -0,0 +1,81 @@
...
+        }
+    }, [showTooltipAt, showTooltipFromEvent, hideTooltip])
+
+    return {
+        actions,
+        state,
+    }
+}
+
+export const useTooltip = () => {
```
wyze, line 65: Would this be better to rename to `useTooltipActions` ?

packages/tooltip/src/context.ts
```diff
@@ -0,0 +1,45 @@
...
+    hideTooltip: () => void
+}
+
+const defaultActions: TooltipActionsContextData = {
+    showTooltipAt: () => {},
+    showTooltipFromEvent: () => {},
+    hideTooltip: () => {},
+}
+
+export const TooltipActionsContext = createContext<TooltipActionsContextData>(defaultActions)
```
wyze, line 20: You could have not provided a default value for the contexts, since in the hooks you are checking for undefined values before returning the context value.

packages/core/src/components/ConditionalWrapper.js
plouc: I think it's fine to use a custom component if we want this kind of nesting, I think it's mostly useful for the `Container`, I also changed this because I struggled to get proper typings for it.

packages/tooltip/src/hooks.ts
plouc: I just kept the existing logic here, but I agree.

packages/tooltip/src/hooks.ts
plouc: I think `useTooltip` is enough, the distinction between `state` and `actions` is mostly internal.

packages/tooltip/src/context.ts
plouc: Right.

wyze:
I know you already merged this but I decided to throw my thoughts on here anyways. :)

Awesome idea of `LegacyContainer`, we can then track which components still need updated away from render props. Good work on reusing Container for the HOC, I almost did that but I thought there were some subtle differences so I didn't proceed. Definitely better to split contexts by state and actions, I do that in my work application.

As far as using `withContainer` vs. `Container`, I think the only down side is the display name, but this is probably fixable.

Storybook for Pie:

image

Storybook for Bullet:

image

PR #1282, opened 2020-11-16
Add the ability to round bar index scale

ericsoco:
I noticed that `ResponsiveBar` charts with many, narrow bars were varying in width. The `svg` container was the correct width, but the bars within it were offset from the left/right edges.

Screencap:
nivo-bar-nicing

After some digging, I narrowed it down to the use of `d3-scale.scaleBand.rangeRound()` within `Bar`'s `common.js::getIndexedScale()`. This diff puts this nicing behavior behind a `nice` prop, which defaults to `true` to maintain current functionality unless a developer opts out with `nice=false`.

The above test with `nice=false`:
nivo-bar-nicing-off

I didn't add anything to the Bar demo site; I can if you like, but this prop feels niche / edge-case enough that it may not need to be surfaced in the demo site.

### DIFF
--- a/packages/bar/index.d.ts
+++ b/packages/bar/index.d.ts
@@ -99,6 +99,7 @@ declare module '@nivo/bar' {
         padding: number
 
         valueScale: Scale
+        indexedScale: Scale
 
         axisBottom: AxisProps | null
         axisLeft: AxisProps | null
--- a/packages/bar/index.d.ts
+++ b/packages/bar/index.d.ts
@@ -18,7 +18,7 @@ import {
 import { AxisProps, GridValues } from '@nivo/axes'
 import { OrdinalColorScaleConfig, InheritedColorConfig } from '@nivo/colors'
 import { LegendProps } from '@nivo/legends'
-import { Scale } from '@nivo/scales'
+import { Scale, BandScale } from '@nivo/scales'
 
 declare module '@nivo/bar' {
     export type Value = string | number
@@ -99,7 +99,7 @@ declare module '@nivo/bar' {
         padding: number
 
         valueScale: Scale
-        indexScale: Scale
+        indexScale: BandScale
 
         axisBottom: AxisProps | null
         axisLeft: AxisProps | null
--- a/packages/bar/index.d.ts
+++ b/packages/bar/index.d.ts
@@ -99,7 +99,7 @@ declare module '@nivo/bar' {
         padding: number
 
         valueScale: Scale
-        indexedScale: Scale
+        indexScale: Scale
 
         axisBottom: AxisProps | null
         axisLeft: AxisProps | null
--- a/packages/bar/src/compute/common.js
+++ b/packages/bar/src/compute/common.js
@@ -11,17 +11,17 @@ import { scaleBand } from 'd3-scale'
 /**
  * Generates indexed scale.
  *
- * @param {Array.<Object>} data
- * @param {Function}       getIndex
- * @param {Array.<number>} range
- * @param {number}         padding
- * @Param {boolean}        nice
+ * @param {Array.<Object>}   data
+ * @param {Function}         getIndex
+ * @param {Array.<number>}   range
+ * @param {number}           padding
+ * @Param {indexedPropType}  indexedScale
  * @returns {Function}
  */
-export const getIndexedScale = (data, getIndex, range, padding, nice) => {
+export const getIndexedScale = (data, getIndex, range, padding, indexedScale) => {
     const scale = scaleBand().domain(data.map(getIndex)).padding(padding)
-    return nice ? scale.rangeRound(range) : scale.range(range)
-}    
+    return indexedScale.round ? scale.rangeRound(range) : scale.range(range)
+}   
 
 export const normalizeData = (data, keys) =>
     data.map(item => ({
--- a/packages/bar/src/compute/common.js
+++ b/packages/bar/src/compute/common.js
@@ -19,9 +19,7 @@ import { scaleBand } from 'd3-scale'
  * @returns {Function}
  */
 export const getIndexedScale = (data, getIndex, range, padding, nice) => {
-    const scale = scaleBand()
-        .domain(data.map(getIndex))
-        .padding(padding)
+    const scale = scaleBand().domain(data.map(getIndex)).padding(padding)
     return nice ? scale.rangeRound(range) : scale.range(range)
 }    
 
--- a/packages/bar/src/compute/common.js
+++ b/packages/bar/src/compute/common.js
@@ -11,17 +11,17 @@ import { scaleBand } from 'd3-scale'
 /**
  * Generates indexed scale.
  *
- * @param {Array.<Object>}   data
- * @param {Function}         getIndex
- * @param {Array.<number>}   range
- * @param {number}           padding
- * @Param {indexedPropType}  indexedScale
+ * @param {Array.<Object>} data
+ * @param {Function}       getIndex
+ * @param {Array.<number>} range
+ * @param {number}         padding
+ * @Param {scalePropType}  indexScale
  * @returns {Function}
  */
-export const getIndexedScale = (data, getIndex, range, padding, indexedScale) => {
+export const getIndexedScale = (data, getIndex, range, padding, indexScale) => {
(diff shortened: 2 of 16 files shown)


plouc:
@ericsoco seems interesting to have the option, however I'm not sure about the naming, having `nice` as a prop on the top level component is hard to figure out I think, maybe something like `roundIndexScale`? This should also be added to the TypeScript definitions, ideally I'd like to have everything documented, but the way the website works is not documented, so I can definitely understand if you don't want to :) It would also be nice to add a unit test for it.

wyze:
I agree about the prop name. Something like `indexScale={{ nice?: boolean, round?: boolean }}` to be more flexible to accept future options if needed.

ericsoco:
@plouc -- I can look into added to TS defs and a unit test; can also take a look at adding to the website if it's not too complex!

Also, I realize now that `nice` is slightly different semantically than `rangeRound` ([docs](https://github.com/d3/d3-scale)), so agreed on naming; can take @wyze's approach and refactor to `indexScale={round}`. I'll leave out "nice" for now, can add in the future if useful.

plouc:
Thank you @ericsoco! Yes @wyze approach is more flexible and align with the `valueScale` we already have.

packages/scales/index.d.ts
```diff
@@ -50,13 +50,19 @@ declare module '@nivo/scales' {
         max?: 'auto' | number
     }
 
+    export interface IndexedScale {
```
plouc, line 53: This should be a `BandScale` to match d3 types.

packages/scales/src/index.js
```diff
@@ -26,4 +27,5 @@ export const scalePropType = PropTypes.oneOfType([
     PropTypes.shape(timeScalePropTypes),
     PropTypes.shape(logScalePropTypes),
     PropTypes.shape(symLogScalePropTypes),
+    PropTypes.shape(indexedScalePropTypes),
```
plouc, line 30: Would have to be renamed to `bandScalePropTypes`.

packages/scales/src/indexedScale.js
```diff
@@ -0,0 +1,14 @@
...
+ * This file is part of the nivo project.
+ *
+ * Copyright 2016-present, Raphaël Benitte.
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+import PropTypes from 'prop-types'
+
+export const indexedScalePropTypes = {
```
plouc, line 11: This should be `bandScalePropTypes`.

packages/bar/index.d.ts
```diff
@@ -99,6 +99,7 @@ declare module '@nivo/bar' {
         padding: number
 
         valueScale: Scale
+        indexedScale: Scale
```
plouc, line 102: I think it would be better to call it `indexScale`, the scale type is a bandScale, and is used for indexes.

plouc (changes requested):

plouc:
@ericsoco Thank you for the adjustments, maybe what we suggested was confusing, the property should be `indexScale`, to be consistent with `valueScale` (we cannot really use x/y scale for names as the chart supports different layouts), but the scale type is a band scale.

ericsoco:
No worries, I can make these changes now (and fix the failing lint).

packages/bar/src/compute/common.js
```diff
@@ -15,10 +15,13 @@ import { scaleBand } from 'd3-scale'
  * @param {Function}       getIndex
  * @param {Array.<number>} range
  * @param {number}         padding
+ * @Param {scalePropType}  indexScale
  * @returns {Function}
  */
-export const getIndexedScale = (data, getIndex, range, padding) =>
-    scaleBand().rangeRound(range).domain(data.map(getIndex)).padding(padding)
+export const getIndexedScale = (data, getIndex, range, padding, indexScale) => {
```
plouc, line 21: nit: I think we could also rename this to `getIndexScale()`.

packages/bar/src/compute/stacked.js
```diff
@@ -149,12 +149,13 @@ export const generateStackedBars = ({
...
     padding = 0,
     valueScale,
+    indexScale,
     ...props
 }) => {
     const stackedData = stack().keys(keys).offset(stackOffsetDiverging)(normalizeData(data, keys))
 
     const [axis, range] = layout === 'vertical' ? ['y', [0, width]] : ['x', [height, 0]]
-    const indexedScale = getIndexedScale(data, props.getIndex, range, padding)
+    const indexedScale = getIndexedScale(data, props.getIndex, range, padding, indexScale)
```
plouc, line 158: Maybe `indexScale` and `indexScaleConfig`, as you did before?

packages/bar/src/props.js
```diff
@@ -34,6 +34,8 @@ export const BarPropTypes = {
     layout: PropTypes.oneOf(['horizontal', 'vertical']).isRequired,
     reverse: PropTypes.bool.isRequired,
     valueScale: scalePropType.isRequired,
+    indexScale: scalePropType.isRequired,
```
plouc, line 37: Actually, we should use `bandScalePropTypes` as other scale types aren't supported.

packages/bar/index.d.ts
```diff
@@ -99,6 +99,7 @@ declare module '@nivo/bar' {
         padding: number
 
         valueScale: Scale
+        indexScale: Scale
```
plouc, line 102: We should restrict to BandScale here, otherwise it might give the impression that other types are supported too.

plouc (changes requested):

wyze (changes requested):
Also need to update the canvas implementation to pass the value through to the generate bars methods or it will crash. I made the mistake with valueScale :O.

https://github.com/plouc/nivo/blob/484235ff44bf3cb516e906f8c2b101669b17fc8a/packages/bar/src/BarCanvas.js#L58

website/src/pages/bar/api.js
```diff
@@ -60,6 +60,7 @@ const BarApi = () => {
                     reverse: false,
 
                     valueScale: { type: 'linear' },
+                    indexScale: { type: 'band', round: false },
```
wyze, line 63:
```suggestion
                    indexScale: { type: 'band', round: true },
```

website/src/pages/bar/canvas.js
```diff
@@ -41,6 +41,7 @@ const initialProperties = {
     reverse: false,
 
     valueScale: { type: 'linear' },
+    indexScale: { type: 'band', round: false },
```
wyze, line 44:
```suggestion
    indexScale: { type: 'band', round: true },
```

website/src/pages/bar/index.js
```diff
@@ -39,6 +39,7 @@ const initialProperties = {
     reverse: false,
 
     valueScale: { type: 'linear' },
+    indexScale: { type: 'band', round: false },
```
wyze, line 42:
```suggestion
    indexScale: { type: 'band', round: true },
```

wyze (changes requested):
Code looks really good! https://dlqum.sse.codesandbox.io/bar/

Only minor change is I think we should change the website values to default to the default values of the component.

ericsoco:
@wyze oof good catch! Fixed.

wyze (approved):
Looks great! Thanks. https://z5c1z.sse.codesandbox.io/bar/

plouc (approved):
LGTM

plouc:
Thank you @ericsoco!

ericsoco:
Thank _you_ @plouc for the lovely library, and @wyze as well 😄

jamesdh:
@ericsoco I'm looking back at this issue and I see your bar chart uses time series data for its X-axis but I don't see any way to do so without converting their value to a string (which causes every bar to have a tick). Can any one of you guys clarify how that was done?

ericsoco:
I have string values for x-axis formatted as `YYYY-MM`, and I'm passing `xFormat: 'time:%Y-%m'` to `ResponsiveBar`.

I don't have a tick on every bar because I'm not using Nivo to render the x-axis labels / ticks, this is a hybrid component that functions as a slider (using Material-UI) with a bar chart overlaid.

I think what I just told you is probably not helpful for your use case...sorry!

PR #1287, opened 2020-11-16
chore(build): add version script to bump peer dependencies on new version

wyze:
This should run on publish automatically. I did run it to increment all the outstanding changes, maybe we can do it in a separate commit. It should only increment it when the core package version changes and out of sync with any package that has a peerDependency on core.


plouc (approved):
LGTM

PR #1291, opened 2020-11-17
fix(bar): fix stacked bars when key is missing

### LINKED ISSUE #1289: Bar keys need to be identical
**Describe/explain the bug**
Bars will not render unless keys are identical across all data members.

**To Reproduce**
https://codesandbox.io/s/nivo-basic-demo-forked-445rf?file=/src/index.js

Steps to fix the behavior:
1. change "donut" to "kebab" on US
2. remove "donut" from keys array
4. Voila, now it does render

**Expected behavior**
Chart should render whether or not the keys are identical, there's too many real use cases (a country where donuts are prohibited 🤣). Also I don't think that the answer will be to normalize the data by setting that missing key to 0, Sometime you want to select top 3 menu by country(so we need to trim the data).

Thanks

wyze:
Close #1289

### DIFF
--- a/packages/bar/src/Bar.js
+++ b/packages/bar/src/Bar.js
@@ -192,16 +192,18 @@ const Bar = props => {
                             key="bars"
                             willEnter={willEnter}
                             willLeave={willLeave}
-                            styles={result.bars.map(bar => ({
-                                key: bar.key,
-                                data: bar,
-                                style: {
-                                    x: spring(bar.x, springConfig),
-                                    y: spring(bar.y, springConfig),
-                                    width: spring(bar.width, springConfig),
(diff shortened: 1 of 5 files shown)


plouc (approved):
LGTM

joealden:
@wyze sorry to bug you about this, but do you have an idea of when this fix will be included in a release? My team recently updated to 0.65.1 from 0.63.0 without realising this problem broke one of our graphs, and we'd prefer to not to have to rollback to a previous version as we've started writing code that depends on some of the new features introduced recently :)

wyze:
@joealden I just published a new release, so you should be good to pull it in!

joealden:
@wyze I just saw that, thanks a lot!

PR #1292, opened 2020-11-17
fix(core): add missing properties back to theme type

p45mark:
Add missing `Theme` properties back that were inadvertently removed with `@nivo/line` version `0.64.0`

- background
- fontFamily
- fontSize
- textColor


wyze (approved):
Looks good, thanks!

PR #1293, opened 2020-11-17
chore(build): bump any nivo peer dep on version mismatch

wyze:
This should work for any peer + dev dependency that starts with '@nivo' and detects a version change. It uses devDependency as the true source since that is what lerna bumps when it is publishing.

Also added the current version of tooltip to core, so next time it runs you will see that get bumped.


plouc (approved):
LGTM

PR #1305, opened 2020-11-25
feat(axes): migrate package to typescript

### LINKED ISSUE #1217: Migrate @nivo/axes to TypeScript

### LINKED ISSUE #1334: Export default AxisTick
I wanted to do a custom render of the default AxisTick, basically just some conditional rendering, but seems @nivo/axes does not export this component.

Provide named export of this component.

Some cases can be probably solved by clever usage of axis prop on chart component, but there are not many examples, so I just wanted to "hack" around it.

wyze:
Close #1217
Close #1334


packages/axes/src/components/Axes.tsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react'
+import { Axis } from './Axis'
+import { AnyScale, AxisProp } from '../types'
+
+const positions = ['top', 'right', 'bottom', 'left'] as const
```
plouc, line 5: Positions are also defined in the canvas implementation, maybe we can move this to a types file.

packages/axes/src/components/Axes.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+export const Axes = <X extends number | string | Date, Y extends number | string | Date>({
```
plouc, line 7: I think we could create an alias for `number | string | Date`, it would be easier to maintain.

packages/axes/src/compute.ts
```diff
@@ -120,30 +125,39 @@ export const getScaleTicks = (scale, spec) => {
...
-export const computeCartesianTicks = ({
+export const computeCartesianTicks = <Value extends string | number | Date>({
     axis,
     scale,
     ticksPosition,
     tickValues,
-    tickSize,
-    tickPadding,
-    tickRotation,
+    tickSize = NaN,
```
plouc, line 133: Why do we need those? I'd suggest to add a comment if they're really required.

packages/axes/src/compute.ts
```diff
@@ -197,43 +211,60 @@ export const computeCartesianTicks = ({
...
-            key: `${v}`,
-            x1: 0,
-            x2: width,
-            y1: position(v),
-            y2: position(v),
-        }))
-    }
+    const position = 'bandwidth' in scale ? centerScale(scale) : scale
+
+    const lines: Line[] = values.map(value => {
```
plouc, line 249: nit: the `if` outside was to avoid running this check for each item, as it's stable.

packages/axes/src/types.ts
```diff
@@ -0,0 +1,122 @@
...
+    x: number
+    y: number
+}
+
+export type ScaleWithBandwidth =
+    | (ScaleBand<any> & { type: 'band' })
+    | (ScalePoint<any> & { type: 'point' })
+
+export type AnyScale =
+    | (ScaleLinear<any, any> & { type: 'linear' })
```
plouc, line 33: I think output is always a number, and input could be limited to `string | number | Date` as this is what's supported in this package.

packages/axes/src/types.ts
```diff
@@ -0,0 +1,122 @@
...
+    | (ScaleOrdinal<any, any> & { type: 'ordinal' })
```
plouc, line 34: I think we can also use better types here.

packages/axes/src/types.ts
```diff
@@ -0,0 +1,122 @@
...
+    | (ScaleTime<any, any> & { useUTC: boolean; type: 'time' })
```
plouc, line 35: Same for time scales.

packages/axes/src/types.ts
```diff
@@ -0,0 +1,122 @@
...
+    | (ScaleSymLog<any, any> & { type: 'symlog' })
```
plouc, line 36: Well, this is valid for all scales defined here :)

packages/axes/src/types.ts
```diff
@@ -0,0 +1,122 @@
...
+    // string is used for Date based scales,
+    // it can express a time interval,
+    // for example: every 2 weeks
+    | string
+    // override scale ticks with custom explicit values
+    | Value[]
+
+export type AxisLegendPosition = 'start' | 'middle' | 'end'
+
+export type ValueFormatter<Value extends string | number | Date> = (value: Value) => Value
```
plouc, line 54: I think the output should be `Value | string`

packages/axes/src/types.ts
```diff
@@ -0,0 +1,122 @@
...
+export type Point = {
+    x: number
+    y: number
+}
+
+export type ScaleWithBandwidth =
+    | (ScaleBand<any> & { type: 'band' })
+    | (ScalePoint<any> & { type: 'point' })
+
+export type AnyScale =
```
plouc, line 32: Maybe this type can be a generic where we pass the type of the value.

packages/axes/src/types.ts
```diff
@@ -0,0 +1,122 @@
...
+    // for example: every 2 weeks
+    | string
+    // override scale ticks with custom explicit values
+    | Value[]
+
+export type AxisLegendPosition = 'start' | 'middle' | 'end'
+
+export type ValueFormatter<Value extends string | number | Date> = (value: Value) => Value
+
+export interface AxisProp<Value extends string | number | Date> {
```
plouc, line 56: `AxisProp` and `AxisProps` might be confusing as they both hold props, maybe we could use a more descriptive name?

packages/axes/src/types.ts
```diff
@@ -0,0 +1,122 @@
...
+    legendPosition?: AxisLegendPosition
+    legendOffset?: number
+}
+
+export interface CanvasAxisProp<Value extends string | number | Date>
+    extends Omit<AxisProp<Value>, 'legend'> {
+    legend?: string
+}
+
+export interface AxisProps<Value extends number | string | Date = any> {
```
plouc, line 74: Could this extend the type already defined earlier?

packages/bullet/src/BulletItem.tsx
```diff
@@ -137,7 +137,6 @@ export const BulletItem = ({
     const axis = (
         <g transform={`translate(${axisX},${axisY})`}>
             <Axis
-                // @ts-ignore
```
plouc, line 140: Maybe we could defined an `AxisType` and cast?

plouc (changes requested):

packages/axes/src/compute.ts
wyze: I think this was to keep the types as optional, but I went ahead and removed it.

packages/axes/src/canvas.ts
```diff
@@ -50,42 +59,54 @@ export const renderAxisToCanvas = (
...
         ctx.beginPath()
         ctx.moveTo(0, 0)
         ctx.lineTo(axis === 'x' ? length : 0, axis === 'x' ? 0 : length)
         ctx.stroke()
     }
 
     ticks.forEach(tick => {
-        if (theme.axis.ticks.line.strokeWidth > 0) {
-            ctx.lineWidth = theme.axis.ticks.line.strokeWidth
+        if ((theme.axis.ticks.line.strokeWidth ?? 0) > 0) {
```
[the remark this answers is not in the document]
wyze: Just a few lines down it references the tick.

packages/axes/src/types.ts
wyze: I coudln't get this to work.

packages/axes/src/types.ts
wyze: I've updated the output, but the `Domain` which is the first argument did not like the `string | number | Date`.

packages/bullet/src/BulletItem.tsx
wyze: Not sure I understand what you mean here?

Issue #1314, opened 2020-12-01
v0.66.0: npm/yarn package has wrong peerDependencies


KENNYSOFT:
**Describe/explain the bug**

v0.66.0: npm/yarn package has wrong `peerDependencies`

**To Reproduce**

1. Make an empty directory, and walk into the directory
2. Run command:
    ```shell
    npm install --save react@16.14.0 prop-types@15.7.2 @nivo/core@0.66.0 @nivo/line@0.66.0 @nivo/tooltip@0.66.0
    ```
    or
    ```shell
    yarn add react@16.14.0 prop-types@15.7.2 @nivo/line@0.66.0 @nivo/core@0.66.0 @nivo/tooltip@0.66.0
    ```
3. See warnings about peer dependencies like following:
    > warning " > @nivo/line@0.66.0" has incorrect peer dependency "@nivo/core@0.65.0".
    > warning "@nivo/line > @nivo/axes@0.66.0" has incorrect peer dependency "@nivo/core@0.65.0".
    > warning "@nivo/line > @nivo/annotations@0.66.0" has incorrect peer dependency "@nivo/core@0.65.0".
    > warning "@nivo/line > @nivo/colors@0.66.0" has incorrect peer dependency "@nivo/core@0.65.0".
    > warning "@nivo/line > @nivo/voronoi@0.66.0" has incorrect peer dependency "@nivo/core@0.65.0".
    > warning "@nivo/line > @nivo/legends@0.66.0" has incorrect peer dependency "@nivo/core@0.65.0".
    > warning "@nivo/line > @nivo/tooltip@0.66.0" has incorrect peer dependency "@nivo/core@0.65.0".
    > warning " > @nivo/core@0.66.0" has unmet peer dependency "@nivo/tooltip@0.63.0".

**Expected behavior**

Not complaining about peer dependencies since it correctly set to 0.66.0 in 1f6cb2c933d1a319f223ea5509f12e17a952addc

**Screenshots**
npm:
![image](https://user-images.githubusercontent.com/6624567/100705169-19042d00-33ea-11eb-9647-b92301089b69.png)

yarn:
![image](https://user-images.githubusercontent.com/6624567/100705051-ebb77f00-33e9-11eb-8300-8ba402faeef2.png)

**Desktop:**
 - OS: Windows
 - Browser: -
 - Version: -

**Additional context**

[...]

PR #1317, opened 2020-12-02
chore(build): workaround lerna issues

wyze:
It was pointed out that after the last release, the peerDependencies still weren't updated correctly. If we go look at https://unpkg.com/browse/@nivo/colors@0.66.0/package.json we can see indeed, peer dependencies are still at 0.65.0 for @nivo/core. It also has the `gitHead` field that points to the commit, https://github.com/plouc/nivo/commit/1f6cb2c933d1a319f223ea5509f12e17a952addc, which _does_ include the correct peer dependencies.

So, we are going to run the script in 2 different lifecycle methods, `version` and `prepublishOnly`. Also, note that the generated changelog was not in that commit, so updated the script to stage it for commit.

Unrelated, but the build includes the `tsconfig.tsbuildinfo` from typescript for our typescript packages, so I added an exclude in the `files` field for that file.

Reference #1314

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,9 @@
   "scripts": {
     "changelog": "rm CHANGELOG.md && node scripts/generate-changelog.js",
     "make:init": "make init",
-    "version": "yarn changelog && node scripts/bump-peer-deps.js"
+    "prepublishOnly": "yarn release:prepare",
+    "release:prepare": "node scripts/bump-peer-deps.js && yarn changelog",
+    "version": "yarn release:prepare"
   },
   "workspaces": [
     "packages/*",
--- a/packages/bullet/package.json
+++ b/packages/bullet/package.json
@@ -25,7 +25,8 @@
   "files": [
     "README.md",
     "LICENSE.md",
-    "dist/"
+    "dist/",
+    "!dist/tsconfig.tsbuildinfo"
   ],
   "dependencies": {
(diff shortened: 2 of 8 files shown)


plouc (approved):
LGTM

plouc:
I wish there was an official way to do it... but anyway, thank you very much for the fix, it confuses users a lot.

wyze:
Yes I agree. I wish lerna supported this via a flag or some option.

andefred:
@wyze Not sure if this is relevant, but I've recently started working a lot with NX Workspace, and hear a lot of users who migrate from lerna to NX... maybe that's something to consider down the road?

Also, what's the timing for new releases considering the 0.66.0 in npm is broken?

Lastly, is there anything I can do to help? this library is looking very promising!

plouc:
@andefred, thank you for the suggestion, except the issue with the peer dependencies, I found Lerna quite good, I'm not really considering switching to a new tool at the moment.

I'll try to create a new release this week-end.

Sure ;) our current big project is to [migrate all packages to TypeScript](https://github.com/plouc/nivo/projects/1), we also want to get rid of `recompose` in favor of React hooks, and replace `react-motion` with `react-spring` for example.

plouc:
@andefred `0.67.0` published with proper versions.

KENNYSOFT:
Thanks for the fix.

Issue #1301, opened 2020-11-19
Area bump styles not updating - trapped in React.memo


LFDM:
### Observed problem

- Go to https://nivo.rocks/area-bump/
- Try to hover over an area. You will see that the legend labels change their style, but the areas themselves do not.
- When you click one of the areas, you see active and inactive styles. (as outlined below, this is a red herring)
- Once clicked, you can never go back to a state, where no area is selected

### Expected behavior

- According to https://github.com/plouc/nivo/blob/master/packages/bump/src/area-bump/hooks.js#L179 hovering over an area should trigger active/inactive styles. Clicking is unrelated to these style changes.

### What's going on

- This is in essence a memoization issue.
- `onMouseEnter` the current series is set

https://github.com/plouc/nivo/blob/master/packages/bump/src/area-bump/hooks.js#L191
```javascript
    const handleMouseEnter = useCallback(
        event => {
            showTooltipFromEvent(React.createElement(tooltip, { serie }), event)
            setCurrent(serie.id)
            onMouseEnter && onMouseEnter(serie, event)
        },
        [serie, onMouseEnter, showTooltipFromEvent, setCurrent]
    )
```
- The current series is used to compute new styles

https://github.com/plouc/nivo/blob/master/packages/bump/src/area-bump/hooks.js#L145
```javascript
    const getSerieStyle = useSerieStyle({
        fillOpacity,
        activeFillOpacity,
        inactiveFillOpacity,
        borderWidth,
        activeBorderWidth,
        inactiveBorderWidth,
        borderColor,
        borderOpacity,
        activeBorderOpacity,
        inactiveBorderOpacity,
        isInteractive,
        current,
    })

    const series = useMemo(
        () =>
            rawSeries.map(serie => {
                serie.color = getColor(serie)
                serie.style = getSerieStyle(serie)

                return serie
            }),
        [rawSeries, getColor, getSerieStyle]
    )
```

- This is where the problem begins. Specifically here (comments added by me)

https://github.com/plouc/nivo/blob/master/packages/bump/src/area-bump/hooks.js#L160
```javascript
    const series = useMemo(
        () =>
            rawSeries.map(serie => {
                serie.color = getColor(serie) // mutated in place!
                serie.style = getSerieStyle(serie) // mutated in place!
                
                // 
                return serie // returns the same referentially equal object, even if color or styles have changed
            }),
        [rawSeries, getColor, getSerieStyle]
    )
```
[...]

PR #1329, opened 2020-12-04
Fix for Area Bump chart problems in #1301

LFDM:
Please have a look at the original issue described in #1301 and at the commit message descriptions of the two first commits of this PR for more details.

The fix can be observed through storybook or the website. In both cases, hovering over the AreaBump chart will prior to this PR do nothing, and after this PR correctly trigger active/inactive styles on the chart.


plouc (approved):
LGTM

plouc:
@LFDM thank you very much for this fix!

LFDM:
:+1: Thanks for the quick response!

PR #1331, opened 2020-12-05
feat(arcs): introduce @nivo/arcs package

plouc:
The purpose of this PR was initially to add transitions to `@nivo/pie`, but when starting to implement it, I realized that this should also be useful for a few existing packages: `@nivo/pie`, `@nivo/chord` and `@nivo/sunburst`, which led me to think of introducing a `@nivo/arc` package.

This also allows us to move some logic specific to arcs in this new package like cursor detection for canvas charts, hooks for configuring and memoizing a d3 arc generator... while making the core package a bit lighter.

I haven't integrated it yet in the sunburst package as it might require a bit of work, and just replaced the cursor detection in the chord package, but ultimately, we should use more from this new package.

It could also help building new chart types such as a Rose chart.


ollwenjones:
sounds really cool 👍

plouc:
@wyze if you could have a look at this PR, that would be great, I'm sorry, it's huge 🙇🏻

wyze:
Of course @plouc, but only if you look at my axes PR next. :)

plouc:
Right, sorry also for that, in fact I already started, I'm gonna try to finish reviewing it in the next upcoming days.

packages/arcs/src/arc_labels/ArcLabel.tsx
```diff
@@ -0,0 +1,40 @@
+import React, { CSSProperties } from 'react'
```
wyze, line 1: Not a big fan of the underscore in the folder name, how about a dash (`-`)?

packages/arcs/src/arc_link_labels/ArcLinkLabel.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react'
```
wyze, line 1: Again, underscore replaced by dash in the folder name?

packages/arcs/src/centers.ts
```diff
@@ -0,0 +1,128 @@
...
+    computeExtraProps = () => ({} as ExtraProps),
+}: {
+    data: Datum[]
+    // define where the centers should be placed,
+    // 0.0: inner radius
+    // 0.5: center
+    // 1.0: outer radius
+    offset?: number
+    // arcs with a length below this (end angle - start angle in degrees)
+    // are gonna be excluded, this can be typically used to avoid having
```
wyze, line 106:
```suggestion
    // are going to be excluded, this can typically be used to avoid having
```

packages/arcs/src/index.ts
```diff
@@ -0,0 +1,13 @@
+export * from './arc_labels'
+export * from './arc_link_labels'
```
wyze, line 2:
```suggestion
export * from './arc-labels'
export * from './arc-link-labels'
```

If you decide to switch over from underscore to dashes, make this change as well.

packages/arcs/src/interpolateArc.ts
```diff
@@ -0,0 +1,27 @@
+import { to, SpringValue } from 'react-spring'
+import { ArcGenerator } from './types'
+
+/**
+ * Directly animating paths for arcs leads to sub-optimal results
+ * as the interpolation is gonna be linear while we deal with polar coordinates,
+ * this interpolator is gonna generate proper arc transitions.
```
wyze, line 7:
```suggestion
 * as the interpolation is going to be linear while we deal with polar coordinates,
 * this interpolator is going to generate proper arc transitions.
```

packages/arcs/src/useArcGenerator.ts
```diff
@@ -0,0 +1,29 @@
...
+import { arc as d3Arc } from 'd3-shape'
+import { ArcGenerator, Arc } from './types'
+
+/**
+ * Memoize a d3 arc generator.
+ *
+ * Please note that both inner/outer radius aren't static
+ * and should come from the arc itself, while it requires
+ * more props on the arcs, it provides more flexibility
+ * because it's not limited to pie then but can also works
```
wyze, line 11:
```suggestion
 * because it's not limited to pie then but can also work
```

packages/arcs/src/utils.ts
```diff
@@ -0,0 +1,35 @@
+import { useMemo } from 'react'
+import { radiansToDegrees } from '@nivo/core'
+import { DatumWithArc } from './types'
+
+/**
+ * Make sure an angle (expressed in radians)
+ * always fall in the range 0~2*PI.
```
wyze, line 7:
```suggestion
 * always falls in the range 0~2*PI.
```

packages/arcs/tests/boundingBox.test.ts
```diff
@@ -0,0 +1,52 @@
...
+            width: 100,
+            height: 100,
+        },
+    },
+]
+
+for (const boundingBoxCase of boundingBoxCases) {
+    const { args, expected } = boundingBoxCase
+
+    test(`computeArcBoundingBox() for position ${args[0]}, ${args[1]} with radius ${args[2]}, starting at ${args[3]}°, ending at ${args[4]}°`, () => {
```
wyze, line 45: There is also `test.each`, but this works too.

wyze (changes requested):
Looks great! Just minor grammar changes and question about directory names.

packages/arcs/src/arc_labels/ArcLabel.tsx
plouc: I prefer to use underscores :) I guess it's a matter of habits, using python lately, underscores in file names are one.
I don't think it makes a huge difference in the end, but I would prefer to stick with this.

packages/arcs/tests/boundingBox.test.ts
plouc: Thank you, I didn't know! I think we can keep it like this, apparently it accepts an array of arrays, I'm not sure it's worth changing it.

plouc:
> Looks great! Just minor grammar changes and question about directory names.

@wyze, thank you for the review, and the fixes! regarding directory name, I tend to use underscores in most of my projects, so I would prefer to use them.

wyze (approved):
Works for me then. Good to go!

Issue #1315, opened 2020-12-02
Add index to tooltipFormatter for Radar Graphs


petermaksymo:
**Is your feature request related to a problem? Please describe.**
I'd like to normalise my data for Radar graphs relative to each index. I've gone that far but for the tooltip it shows the normalised value and `tooltipFormatter` does not pass `index` so I don't think I get enough information to re-convert and display the original value.

**Describe the solution you'd like**
I'd like `tooltipFormatter` to pass the `index` as an additional parameter.

**Describe alternatives you've considered**
Any other suggestions to normalise the data for RadarGraphs while still showing the original value in the tooltip.

**Additional context**
The change should likely go [here](https://github.com/plouc/nivo/blob/master/packages/radar/src/RadarTooltipItem.js#L35), but I'm not sure if there are other places it's needed or if that would potentially break something else.

PR #1335, opened 2020-12-08
feat(radar) pass index to tooltipFormat

petermaksymo:
Related issue: #1315

Pass the `index` for `tooltipFormat` in Radar graph, this gives users much more information if they want to do a custom formatter for tooltips. For me, knowing which `index` the data is a part of allows me to get the true value for to display in tooltip when otherwise I would be showing a normalised value.


tunesmith:
I am interested in being able to format radar based off of index as well. Can this PR be re-opened?

plouc:
@tunesmith, rather than adding this extra argument, I think the data could be formatted before being passed to the component, other charts don't do that, and we would have to pass datum everywhere, which could affect memoization/performance.

PR #1349, opened 2020-12-15
fix(axes): rtl issue with x-axis

alikazemkhanloo:
This will fix RTL issue with the x-axis label. When using `tickRotate`, `text-align` is incorrect. Most RTL issues are solvable using a small trick like this one.

### DIFF
--- a/packages/axes/src/compute.js
+++ b/packages/axes/src/compute.js
@@ -138,6 +138,7 @@ export const computeCartesianTicks = ({
     const line = { lineX: 0, lineY: 0 }
     const text = { textX: 0, textY: 0 }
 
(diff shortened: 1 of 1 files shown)


wyze:
Hi @alikazemkhanloo, how can I test this out?

alikazemkhanloo:
Hi @wyze, you need to use `dir='rtl'` on html. Then use ` tickRotation: 45` for axisBottom config. you will see something like this without this PR.

Screen Shot 1399-09-26 at 12 50 39

Also, you can inspect the demo page and add `dir="rtl"` to see the issue I'm trying to solve.

wyze:
Looks good to me, thanks!

PR #1355, opened 2020-12-17
Fix type definition of PatternSquaresDef

netzwerg:
What the title says :) Taking this as an opportunity to thank you for this fantastic lib 👍🏼


plouc (approved):

plouc:
Good catch, thank you @netzwerg!

PR #1371, opened 2021-01-03
feat(bar): use custom label for legends

njenia:
allow for custom labels to show in the legend instead of the id or index of the series


thalelinh:
Can you write a small example of how someone would use this?

PR #1386, opened 2021-01-14
fix(core): fix invalid core types

callum-mellorreed-privitar:
Noticed some of the types in core were causing problems for anyone with `skipLibCheck: false` (the default value). Let me know if the reason for any change isn't clear, I've kept each kind of change in separate commit and a brief summary below.

- Cannot nest `declare` statements. Just removing the keyword should be fine
- Can't inline type alias in const declaration
- Accessor definition assumed all strings exist as properties on T
- referenced type `OpaqueInterpolation` doesn't exist, replaced per d3 types


mrmartineau:
@callum-mellorreed-privitar can you confirm if this PR fixes the issues mentioned in #1457?

Also, @plouc is there any reason why this PR might not be reviewed?

callum-mellorreed-privitar:
> @callum-mellorreed-privitar can you confirm if this PR fixes the issues mentioned in #1457?

Assuming the relevant files haven't changed/expanded since, yeah this is to fix the issues shown in #1457. IIRC I just chucked these in my own declaration file to override the nivo ones as this PR hasn't been acknowledged.

nostalgiaz:
@plouc any news related to this merge?

cpelican:
We could really use this fix as well! Any idea if it could be merged? Thanks!

PR #1404, opened 2021-01-28
fix(bar): improve grouped bar performance

gallotamas:
Bar charts are very slow to render with large data sets. It is clearly noticeable on the website example as well: https://nivo.rocks/bar/canvas/

These 2 changes significantly improve the performance:
- don't create new objects in every iteration of `reduce` in `filterNullValues` and `normalizeData` (even though the syntax is less elegant it is far more performant). See [this](https://www.richsnapp.com/article/2019/06-09-reduce-spread-anti-pattern) blog for more explanation.
- only calculate the cleaned data once as it doesn't depend on the keys


wyze (approved):
Great, thank you. There is a noticeable speedup in regenerating the data.

Issue #1425, opened 2021-02-17
HeatMap Label Formatting


arasovic:
I'm using svg heatmap component. I can edit tooltip values but I couldn't find any prop for edit label values.
Can I format label values, like in tooltip ? Do you have any solution?

![heatmap](https://user-images.githubusercontent.com/9802740/108261717-0d1c4180-7175-11eb-82be-cab142f3dc39.png)

second to H:M:S

wyze:
This is solved via #1428.

PR #1428, opened 2021-02-20
feat(heatmap): allow label formatting on heatmaps

robonyong:
Had a similar need to https://github.com/plouc/nivo/issues/1425, wanted to change the label that's shown in a heatmap cell.

I copied the way Bar handles a `label` prop for this one, please let me know if there's a different preferred way to do so now.

### DIFF
--- a/packages/heatmap/index.d.ts
+++ b/packages/heatmap/index.d.ts
@@ -15,6 +15,8 @@ declare module '@nivo/heatmap' {
 
     export type IndexByFunc = (datum: HeatMapDatum) => string | number
 
+    export type LabelFormatter = (datum: HeatMapDatum, key: string) => string | number
+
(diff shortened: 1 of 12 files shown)


wyze (changes requested):
First off, sorry for the late review!

We are trying to get rid of recompose so we'd like to avoid adding any new code containing that. That said we can take the recompose code you did and create a variable with useMemo hook!

```js
const getLabel = useMemo(() => getLabelGenerator(label, labelFormat), [label, labelFormat])
```

robonyong:
@wyze Makes sense! I removed the recompose stuff & switched to using `useMemo` inside the `useHeatmap` hook

wyze (approved):
Looks great, thanks!

Also, thanks for adding stories for the canvas implementation.

PR #1431, opened 2021-02-24
fix(legends): Add missing symbolBorderWidth to typings

bladedeyna:
Add missing property `symbolBorderWidth` to the nivo legends Typescript typings.


wyze:
Thank you @bladedeyna!

PR #1436, opened 2021-02-28
fix(voronoi): fix typo in export class name

swsoyee:
As mentioned in the title.


wyze (approved):
Oops! Thanks for this.

Issue #833, opened 2020-01-18
Network component needs documentation, and API could allow rendering custom nodes


rektdeckard:
Network component seem to be poorly documented on how to use with Layers, as well as interactivity. My immediate desire is to be able to show some information about a node on hover (or always), such as a label. I can't seem to find anything about this anywhere, and it seems to be a very common use case for this type of data visualization. There isn't much to learn from an unlabeled graph.

Ideally the API would allow for rendering any arbitrary component as a node, which would take the place of supplying `nodeColor`, `nodeBorderWidth`, `nodeBorderColor` and allow for more flexibility. For example:

```jsx harmony
const CustomNodesNetwork = data => {
  const renderNode = node => (
    <svg width="50" height="50" version="1.1" xmlns="http://www.w3.org/2000/svg">
      <g>
        <rect
          x={10}
          y={10}
          width={30}
          height={30}
          fill={node.color}
         />
        <text
          x={0}
          y={0}
          fill="black"
        >
          {node.label}
        </text>
      </g>
    </svg>
  );

  return (
    <ResponsiveNetwork
      height={700}
      nodes={data.nodes}
      links={data.links}
      renderNode={t => renderNode(t)}
      linkThickness={t => (2 * (2 - t.source.depth))}
    />
  );
}
```

Alternatively, a default label/hover behavior could easily render some text under certain conditions: 

```jsx harmony
    <ResponsiveNetwork
      ...
      nodeLabel={t => t.label}
      labelVisibility={visible: boolean | string}
    />
```

Thanks for your hard work on this excellent library, and sorry if this the library already has this capability and I'm just dumb! First time using anything D3.js related.

plouc:
No worries, it does not support this, and that's definitely an interesting suggestion, I'm currently working on migrating the whole codebase to TypeScript and trying to improve some packages in the process, I'll see what I can do about this then, do not expect this feature to be available soon though as I've currently just a limited amount of time to spend on side projects.

rektdeckard:
@plouc also, it doesn't seem to be responsive. I have to set a fixed `width` and `height` on the parent component to get it to render.

plouc:
Responsive means it will occupy the available space, while element usually have a width automatically defined by the width of the parent element, that’s not the case for height, nivo cannot guess that, so yes you have to define a height, which can be expressed as a percentage or inherited from a flex box, css grid.

iDreadful:
[...]

Issue #667, opened 2019-08-12
Tracking issue for removing `react-motion` dependency


Taelkir:
I'm getting the following console warning after upgrading to React 16.9.0
```
Warning: componentWillReceiveProps has been renamed, and is not recommended for use. See https://fb.me/react-async-component-lifecycle-hooks for details.

* Move data fetching code or side effects to componentDidUpdate.
* If you're updating state whenever props change, refactor your code to use memoization techniques or move it to static getDerivedStateFromProps. Learn more at: https://fb.me/react-derived-state
* Rename componentWillReceiveProps to UNSAFE_componentWillReceiveProps to suppress this warning in non-strict mode. In React 17.x, only the UNSAFE_ name will work. To rename all deprecated lifecycles to their new names, you can run `npx react-codemod rename-unsafe-lifecycles` in your project source folder.

Please update the following components: Motion, TransitionMotion
```
The above error is what I'm getting in my code, but I have recreated a similar warning in the console here: https://codesandbox.io/s/nivoline-line-layers-461-yogxd?fontsize=14

Only seems to happen to charts with custom tooltips - it doesn't appear for piecharts, but does if I render a choropleth or a line chart

Not sure if Motion/TransitionMotion are dependencies or part of Nivo.

----

Maintainer Edit: This is due to the `react-motion` package which is still used for some of the nivo packages, here is the list of packages still need to be migrated to `react-spring`:

- [x] Bar (#1619)
- [x] Bullet (#1252)
- [x] Chord (#1874)
- [x] Circle Packing (#1360)
- [x] Colors (#1885)
- [x] Network (#1767)
- [x] Scatterplot (#1726)
- [x] Sunburst (#1264)
- [x] Swarmplot (#1365)
- [x] Waffle (#2320)
- [x] Update `@nivo/core` to only support a `@react-spring` config.
- [x] Remove all references to `react-motion` in the project/doc.

ulysset:
Hi @Taelkir  it's part of react-motion, here's the corresponding issue: https://github.com/chenglou/react-motion/issues/569
Unfortunately, it hasn't been fixed yet.

Justin2997:
Maby we should have a fork of the fix in the organization repo, the projet last commit was in 2018.

loganpowell:
I was just about to submit this same issue. It seems React Motion has been abandoned.

gwillz:
[...]

PR #1444, opened 2021-03-10
feat(network) add nodeComponent to network

SheldonNunes:
The purpose of this PR is to introduce a new prop `nodeComponent` to @nivo/network. This `nodeComponent` allows for the overriding of the default Node component allowing for more customisation of networks. This has been a requested feature in a [previously stale issue](https://github.com/plouc/nivo/issues/833).

Also noticed there is no test file for `@nivo/network` so added a new one with some tests related to this change.

Adding in user svg:

Adding in a permanent text component below original Node:

EDIT: This change has also added support for custom node components for NodeCanvas.

### DIFF
--- a/packages/network/src/NetworkCanvas.js
+++ b/packages/network/src/NetworkCanvas.js
@@ -37,6 +37,7 @@ const NetworkCanvas = props => {
         nodeColor,
         nodeBorderWidth,
         nodeBorderColor,
+        renderNode,
 
         linkThickness,
         linkColor,
@@ -94,16 +95,11 @@ const NetworkCanvas = props => {
                 })
             } else if (layer === 'nodes') {
                 nodes.forEach(node => {
-                    ctx.fillStyle = getNodeColor(node)
-                    ctx.beginPath()
-                    ctx.arc(node.x, node.y, node.radius, 0, 2 * Math.PI)
-                    ctx.fill()
-
-                    if (nodeBorderWidth > 0) {
-                        ctx.strokeStyle = getBorderColor(node)
-                        ctx.lineWidth = nodeBorderWidth
-                        ctx.stroke()
-                    }
+                    renderNode(ctx, {
+                        node,
+                        getNodeColor,
+                        getBorderColor,
+                    })
                 })
             } else if (typeof layer === 'function') {
                 layer(ctx, {
@@ -122,6 +118,7 @@ const NetworkCanvas = props => {
         nodes,
         links,
         getNodeColor,
+        renderNode,
         nodeBorderWidth,
         getBorderColor,
         getLinkThickness,
--- a/packages/network/tests/Network.test.tsx
+++ b/packages/network/tests/Network.test.tsx
@@ -1,0 +1,67 @@
+import React from 'react'
+import { mount } from 'enzyme'
+import { Network } from '../src/index'
+
+const nodes = [
+    {
+        id: '0',
+        key: '0',
+        radius: 12,
+        depth: 0,
+        color: 'rgb(244, 117, 96)',
+    },
+    {
+        id: '1',
+        key: '1',
+        radius: 8,
+        depth: 1,
+        color: 'rgb(97, 205, 187)',
+    },
+]
+const links = [
+    {
+        source: '0',
+        target: '1',
+    },
+]
+
+describe('Network', () => {
+    describe('node', () => {
+        it('should render default nodes', () => {
+            const wrapper = mount(
+                <Network
+                    width={400}
+                    height={400}
+                    nodes={nodes}
+                    links={links}
+                    animate={false}
+                    nodeColor={function (e) {
+                        return e.color
+                    }}
+                />
+            )
+            const nodeElements = wrapper.find('Node')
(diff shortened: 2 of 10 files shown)


packages/network/src/Node.js
```diff
@@ -48,4 +48,4 @@ Node.propTypes = {
     handleNodeLeave: PropTypes.func.isRequired,
 }
 
-export default memo(Node)
+export default Node
```
plouc, line 51: Why did you remove the memoization?

packages/network/src/NetworkCanvas.js
```diff
@@ -60,6 +61,7 @@ const NetworkCanvas = props => {
 
     const theme = useTheme()
     const getNodeColor = useNodeColor(nodeColor)
+    const getNodeComponent = useNodeComponent(nodeComponent)
```
plouc, line 64: This isn't used currently, you can limit this to the SVG implementation, or add support for canvas, but you'll have to adjust the TS definitions and prop-types, also a custom node component for canvas would be a function receiving a 2D canvas context as the first argument and the _props_ as the second one. The swarmplot package use this approach if you're curious, and given you're willing to implement it of course :)

packages/network/tests/Network.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+    },
+]
+const links = [
+    {
+        source: '0',
+        target: '1',
+    },
+]
+
+describe('Network', () => {
```
plouc, line 28: Thank you!

website/src/data/components/network/props.js
```diff
@@ -137,7 +137,7 @@ const props = [
     {
         key: 'nodeBorderWidth',
         group: 'Nodes',
-        type: 'number | (node: Node) => number',
+        type: 'Function',
```
plouc, line 140: It can be either a number or a function, a function is always used internally for consistency, but you can use a static number as in the website demo for example.

website/src/data/components/network/props.js
```diff
@@ -150,6 +150,12 @@ const props = [
         defaultValue: NetworkDefaultProps.nodeBorderColor,
         controlType: 'inheritedColor',
     },
+    {
+        key: 'nodeComponent',
+        group: 'Nodes',
+        type: '(node: Node) => svg',
```
plouc, line 156: It's not correct if we consider canvas, you can add `flavors: ['svg'],` if you remove support for canvas, or adjust the type to use a union otherwise.

plouc (changes requested):

plouc:
@SheldonNunes, thank you for this, this could be really useful! and sorry for the late review.

packages/network/src/hooks.js
```diff
@@ -106,6 +106,12 @@ export const useNodeColor = color =>
         return () => color
     }, [color])
 
+export const useNodeComponent = nodeComponent =>
```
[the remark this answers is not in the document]
SheldonNunes: You are right. I had a misunderstanding of why these hooks were being used. I have removed in favour of passing in the component as you suggested

packages/network/src/Node.js
SheldonNunes: Not sure... I have put it back and there is no issue 👍

packages/network/src/NetworkCanvas.js
SheldonNunes: I have updated this PR with this suggestion. Let me know if there is any improvements for the the Canvas implementation 🙏

website/src/data/components/network/props.js
SheldonNunes: Similar to the swarmplot implementation I have added a `nodeComponent` for svg flavor and `renderNode` for canvas flavor.

plouc: Makes sense!

plouc:
@SheldonNunes, as I am currently converting `@nivo/network` to TypeScript: https://github.com/plouc/nivo/pull/1767, I went ahead and added this feature when converting, I wasn't sure you had the time to address the feedbacks, sorry.

plouc:
It's not yet exposed as a property in the PR, but already supported via `createElement`: https://github.com/plouc/nivo/pull/1767/files#diff-e3d2cb937239fcccb5e5a3df4d21fd3af0ce57b457285f18b3147d9ec35bd733R75, I will reuse what you did for canvas.

This package will now also use `react-spring`, because of https://github.com/plouc/nivo/issues/667, this changes the API a little bit, and impacts the internal implementation a lot.

SheldonNunes:
Ah ok! Shall I close off this PR then?

plouc:
@SheldonNunes, yes, again, sorry for that, but it is useful to add the feature in the other branch, so thank you!

SheldonNunes:
No worries. I had kind of forgotten about it until I got pinged it was reopened and had some extra time :) 

Glad this feature will make it into nivo though! :D

plouc:
@SheldonNunes, also adding support for custom links, could be useful if we want dashed lines, arcs instead of straight lines or event labels.

Issue #606, opened 2019-06-25
bullet chart range other than zero


syedali90:
I want to start range other than zero in bullet chart

my data starts from 1000-2000

so i want custom range

Is it possible ?

awatson1978:
Please reopen.

PR #1449, opened 2021-03-14
bugfix for an error that was preventing custom ranges in bullet graph

### LINKED ISSUE #1447: bullet chart range other than zero
Hello.  Would it be possible to support ranges that start at something other than 0?  For example, with the Bullet graph, we might be trying to graph biomarker reference ranges, such as blood pressure or human body temperature.  Body temperature should typically be around 37 Cel, and in a bullet graph one would generally want to display a range from 34C to 42C, with a healthy values between 36C and 40C.  Currently, we would have to display 0C to 42C, which creates a narrow band and is wasteful of space in the user interface and skews the visualization.  

Also, this request has been raised before in #606

Ideally include `rangeMin` and `rangeMax` props on the following components:
- Bullet
- Radar

There may be graphs that could benefit from custom ranges, but those are the two that are most relevant to bioinformatics and healthcare applications.

Using raw D3, or alternative D3 based charting libraries.  But Nivo is so much better than the alternatives, we'd love to see it included here.  We're trying to forward-port visualizations we've built with D3 to a React based approach using Nivo, and custo ranges are the one feature we're missing.

awatson1978:
Bugfix that was preventing the bullet graph from displaying custom ranges.  Fixes #1447 and #606 

Screen Shot 2021-03-12 at 7 51 03 PM

### DIFF
--- a/packages/bullet/src/hooks.ts
+++ b/packages/bullet/src/hooks.ts
@@ -18,7 +18,7 @@ export const useEnhancedData = (
 
                 const max = Math.max(...all)
 
-                const min = Math.min(...all, 0)
+                const min = Math.min(...all)
 
                 const scale = scaleLinear().domain([min, max])
 
--- a/packages/bullet/stories/bullet.stories.tsx
+++ b/packages/bullet/stories/bullet.stories.tsx
@@ -156,3 +156,34 @@ stories.add('custom title', () => (
         }))}
     />
 ))
+
+
+
+const temperatureData = [{
+    "id":"temperature",
+    "ranges":[30,45],
+    "measures":[34,36,40,42],
+    "markers":[37]
+}]
+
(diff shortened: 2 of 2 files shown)


awatson1978:
Well, drat.  8 tests failed because of my tiny change.  I'm now out of my league in knowing how to continue this bugfix.  But a usecase has been provided, the offending line of code has been identified, and the desired graph changes documented.  Going to need some help and guidance from here on out.

collingreen:
I know absolutely zero about this codebase but the failing tests appear to be snapshots, which might not be as bad as it seems. I expect if you put this behind an opt-in prompt you wont fail on all of these existing snapshots but I don't actually know how that works, how new snapshots get added, or anything really about that process.

Side note I'm surprised by most of these snapshots failing based on colors - I don't see why your changes would impact that at all. :/

PR #1466, opened 2021-04-05
fix(core): avoid occasional "ResizeObserver loop" error

ollwenjones:
This error is hard to reproduce, but it's a straightforward workaround to make sure the observer doesn't try to update when it can't within the given frame.

Some more context about the [error](https://github.com/WICG/resize-observer/issues/38), and reference for the [workaround](https://github.com/airbnb/visx/pull/335).

### DIFF
--- a/packages/core/src/hooks/useMeasure.js
+++ b/packages/core/src/hooks/useMeasure.js
@@ -3,20 +3,35 @@ import ResizeObserver from 'resize-observer-polyfill'
 
 export const useMeasure = () => {
     const measureRef = useRef(null)
+    const animationFrameId = useRef(null)
     const [bounds, setBounds] = useState({
         left: 0,
         top: 0,
         width: 0,
         height: 0,
     })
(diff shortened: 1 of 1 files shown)


packages/core/src/hooks/useMeasure.js
```diff
@@ -3,20 +3,33 @@ import ResizeObserver from 'resize-observer-polyfill'
...
+    )
 
     useEffect(() => {
         if (measureRef.current) {
             observer.observe(measureRef.current)
         }
 
-        return () => observer.disconnect()
+        return () => {
+            animationFrameId.current && cancelAnimationFrame(animationFrameId.current)
```
wyze, line 30: Lets change this to an if statement.

wyze (changes requested):

packages/core/src/hooks/useMeasure.js
ollwenjones: done. sorry it's been a while since I circled back.

wyze (approved):
Looks good now, thanks!

PR #1471, opened 2021-04-08
feat(line): use theme values for line slice tooltip

fpammer:
I noticed that the line chart slice tooltip did not apply the theme value for chip so I created this quick fix. For that I reused the Chip component from the tooltip package.

Additionally I added another theme value to theme the value part of the tooltip.

Small preview:
image

### DIFF
--- a/packages/line/src/SliceTooltip.js
+++ b/packages/line/src/SliceTooltip.js
@@ -20,7 +20,7 @@ const SliceTooltip = ({ slice, axis }) => {
             rows={slice.points.map(point => [
                 <Chip key="chip" color={point.serieColor} style={theme.tooltip.chip} />,
                 point.serieId,
-                <span key="value" style={{ ...theme.tooltip.tableCellValue }}>
+                <span key="value" style={theme.tooltip.tableCellValue}>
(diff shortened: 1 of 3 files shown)


packages/line/src/SliceTooltip.js
```diff
@@ -8,25 +8,21 @@
...
     const otherAxis = axis === 'x' ? 'y' : 'x'
 
     return (
         <TableTooltip
             rows={slice.points.map(point => [
-                <Chip key="chip" color={point.serieColor} />,
+                <Chip key="chip" color={point.serieColor} style={theme.tooltip.chip} />,
                 point.serieId,
-                <strong key="value">{point.data[`${otherAxis}Formatted`]}</strong>,
+                <span key="value" style={{ ...theme.tooltip.tableCellValue }}>
```
wyze, line 23:
```suggestion
                <span key="value" style={theme.tooltip.tableCellValue}>
```

Small change. We don't need to create a new object here.

wyze (changes requested):

wyze (approved):

wyze:
Thank you!

PR #1473, opened 2021-04-09
feat(scales): add nice argument to linear and time scales

MasterOdin:
This exposes being able to run the builtin d3 `nice` function over the linear and time scales, which smoothes out the starting/ending points of the range. In both cases, they accept either no argument, a numeric `tickCount`, and in the case of the time scale, a time interval.


wyze (approved):
Looks great, thank you for the feature addition.

Issue #212, opened 2018-05-30
Support calendars less than a year long


kristiehowboutdat:
I'd like to show a range of dates, ex. 3 months, without showing the entire calendar year. I see that you can use the `from` and `to` to specify the year that you want, but I'd like to only show the dates starting and ending on the `from` and `to` dates.

Perhaps a prop to opt in / out of this behavior?

Happy to help with a PR if you agree. THANK YOU for all of your hard work, this is an amazing library!

Related: https://github.com/plouc/nivo/issues/203

plouc:
Hi @kristiehowboutdat,

Thank you.
You're right, it could be a useful feature, but maybe the granularity should be month at least, because month shape is available for styling.
Otherwise it could be a different component, displaying weeks for examples.

kristiehowboutdat:
Having granularity of months would be awesome and should solve most use cases smaller than a year that involve a calendar specifically.

Proposal: Use a prop `granularity` that is either `"month"` or `"year"` with a default of `"year"` for existing compatibility. This is slightly more flexible than a boolean prop `showMonths` or something because it allows for future iteration on the string values without changing the API.

```jsx
// Show the entire calendar year of 2018
// default granularity="year"
<Calendar from="01-01-2018" to="03-31-2018" ... />

// Show the entire calendar year of 2018
<Calendar from="01-01-2018" to="03-31-2018" granularity="year" ... />

// Show only the (entire) months of Jan, Feb, March
<Calendar from="01-01-2018" to="03-31-2018" granularity="month" />

// Show only the (entire) months of Jan, Feb, March, April
<Calendar from="01-01-2018" to="04-01-2018" granularity="month" />
```

Is that the correct understanding? Where in the last example we would show the entire month of April, not just the first week of April?

plouc:
This could be a nice approach, and in fact keeping the same component might be useful to implement some kind of zooming on a specific month.

noodlespace:
+1

sivaraj-dev:
+1 any update @kristiehowboutdat

yturovtsev:
+1

plouc:
Please refrain from adding simple comments with `+1`, you can use the reactions for this :)

<img width="326" alt="capture d ecran 2019-02-14 a 10 51 25" src="https://user-images.githubusercontent.com/501642/52756981-8775d080-3046-11e9-9570-e0b0bccbeab8.png">

[...]

PR #1481, opened 2021-04-15
[issue_212] impement feature to show only the requested range

scherler:
As requested in https://github.com/plouc/nivo/issues/212 I implemented the ability to only show the requested dates.

I added three new stories to cover the edge cases of the issue
image
image
image

Any feedback is highly welcome.

Base usage with an `exact` flag
 ```
<Calendar exact {...commonProps} />
```

### DIFF
--- a/packages/calendar/src/compute.js
+++ b/packages/calendar/src/compute.js
@@ -83,7 +83,12 @@ const computeCellSize = ({
 
     return Math.min(hCellSize, vCellSize)
 }
-
+const getElapsedMonths = (from, to) => {
+    console.log('getElapsedMonths',from.getMonth(), to.getMonth(),  to.getMonth() - from.getMonth());
+    return (
+        to.getMonth() - from.getMonth() + 12 * (to.getFullYear() - from.getFullYear())
+    )
+}
 /**
  * Computes month path and bounding box.
  *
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

**Comment by scherler** [comment]
BTW I created a current version of the branch of the fork mentioned In the issue https://github.com/scherler/nivo/tree/diegosps-calendar-granularity

**Comment by scherler** [comment]
similar to #1173

**Comment by scherler** [comment]
image
this offset should be addressed in a follow up ticket

**Comment by scherler** [comment]
^^ I fixed that and other edge cases such as start date not being first of the month so we do not outLine the month path
image

File: packages/calendar/src/compute.js [hunk 1]
```diff
@@ -94,21 +126,29 @@ const monthPathAndBBox = ({
...
 
+    const startDate = calculateReturnDate(fromDate, timeYear(date), exact)
+    const elapsedMonths = getElapsedMonths(startDate, date)
+    const lastDate = calculateReturnDate(fromDate, timeYear(t1), exact)
+    // if the toDate is not the last of the months, we do not need to paint the path
+    const lastDayOfMonth = new Date(toDate.getFullYear(), toDate.getMonth() + 1, 0)
     // ranges
-    const firstWeek = timeWeek.count(timeYear(date), date)
-    const lastWeek = timeWeek.count(timeYear(t1), t1)
+    let firstWeek = timeWeek.count(startDate, date)
```
romainverduci, line 142: Why changing from `const` to `let` ? I don't see it being updated anywhere

romainverduci (commented):
Overall lgtm given my understanding of the feature.
A couple of comments / questions though.
Kudos for the added stories that helped me to understand what the change is doing better 🎉

packages/calendar/src/compute.js
scherler: I was changing while i was developing and forgot to revert it

scherler:
@romainverduci thanks for your review and I fixed all the issue you brought up.

scherler:
#1503 provides a cleaner implementation

PR #1499, opened 2021-04-30
fix(line): update type for custom line canvas layers

### LINKED ISSUE #1413: Custom line canvas layers have different props then custom svg line layers
A custom canvas line layer has these arguments:

https://github.com/plouc/nivo/blob/ce7c755b23cfc55ead9bac9920a55ef6e764bd99/packages/line/src/LineCanvas.js#L122-L135

But uses this type definition:

https://github.com/plouc/nivo/blob/ce7c755b23cfc55ead9bac9920a55ef6e764bd99/packages/line/index.d.ts#L68-L76

Rather then inheriting its layer type definition form line props a specific type definition needs to be added.

https://github.com/plouc/nivo/blob/ce7c755b23cfc55ead9bac9920a55ef6e764bd99/packages/line/index.d.ts#L216-L218

jguddas:
this resolves #1413

### DIFF
--- a/packages/line/index.d.ts
+++ b/packages/line/index.d.ts
@@ -65,10 +65,19 @@ declare module '@nivo/line' {
         | 'mesh'
         | 'legends'
 
+    export interface LineGenerator {
+        (data: Datum[]): string
+        x:any
+        y:any
+        defined:any
+        curve:any
+        context:any
+    }
+
     export interface CustomLayerProps extends Omit<LineSvgProps, 'xScale' | 'yScale'> {
         innerHeight: number
         innerWidth: number
-        lineGenerator: (data: Datum[]) => string
+        lineGenerator: LineGenerator
         points: Point[]
         series: ComputedSerie[]
         xScale: ScaleFunc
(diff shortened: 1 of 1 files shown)


packages/line/index.d.ts
```diff
@@ -213,8 +213,17 @@ declare module '@nivo/line' {
     export class Line extends React.Component<LineSvgProps & Dimensions> {}
     export class ResponsiveLine extends React.Component<LineSvgProps> {}
 
+    export interface CustomCanvasLayerProps extends CustomLayerProps {
+        ctx: CanvasRenderingContext2D
+        lineGenerator: any
```
wyze, line 218: Is there a reason you are overriding this to `any`?

jguddas: @wyze I did a bit more testing and added:

```typescript
    export interface LineGenerator {
        (data: Datum[]): string
        x:any
        y:any
        defined:any
        curve:any
        context:any
    }
```

I don't know what the correct typings for `x`, `y`, `defined`, `curve` and `context` are I just know that they where missing.

wyze: No worries, just pushed a fix. :)

wyze:
Thanks @jguddas!

jguddas:
Awesome, thanks for the help and the quick me merge @wyze

PR #1503, opened 2021-05-03
feat(calendar): add TimeRange component and storybook

scherler:
Re-write of  #1481,  adding new logic instead of changing the old one.  

image

Forked calendar preview package:

```
"@cloudbees/nivo-calendar": "0.69.4",
```

PR #1504, opened 2021-05-03
[add_typescript_to_calendar] enable type generation

scherler:


scherler:
@wyze can you comment on what is missing to make it into this repository?

wyze:
Is there a reason this is needed? We have only had tsconfig files for when we convert the package to Typescript.

scherler:
It triggers the generation of the types https://github.com/plouc/nivo/blob/93943b88e601e0e612b187a0bfbb6e6c3748597d/Makefile#L164 so when you have a type based project you can use the lib with type support.

wyze:
This was included with #1503.

PR #1512, opened 2021-05-09
feat(legends): migrate package to typescript

### LINKED ISSUE #1221: Migrate @nivo/legends to TypeScript

### LINKED ISSUE #1321: Legend `itemTextColor` no longer works 0.63+
It seems like you can no longer set the legend label colors on 0.63.0+

### 0.62.0 - legend label colors work (are white)
https://codesandbox.io/s/nivopie-typescript-forked-cclvv

### 0.63.0 - legend label colors do not work
https://codesandbox.io/s/nivopie-typescript-forked-xpk7d

wyze:
Close #1221
Close #1321


packages/legends/src/types.ts
```diff
@@ -0,0 +1,143 @@
...
+    itemTextColor?: string
+    itemBackground?: string
+    itemOpacity?: number
+    itemsSpacing?: number
+
+    effects?: EffectProps[]
+}
+
+export type LegendProps = {
+    data: Datum[]
```
jamesdh, line 80: Is requiring a Datum[] here right? I can't find anything in the [Legends guide](https://nivo.rocks/guides/legends) or in the storybooks demonstrating the use of this or explaining what it's for. This and the required `padding` above both broke every one of my  charts.

packages/legends/src/types.ts
```diff
@@ -0,0 +1,143 @@
...
+export type Datum = {
+    id: string | number
+    label: string | number
+    color?: string
+    fill?: string
+}
+
+type CommonLegendProps = {
+    direction: LegendDirection
+    padding: number | Record<'top' | 'right' | 'bottom' | 'left', number>
```
jamesdh, line 65: Similarly, I never used the padding prop before and can't find any for of documentation describing how/what it is used for. Should this really be required?

jamesdh (commented):
Ultimately I just added a default...

```javascript
data: [], 
padding: 0
```
to each of my legend definitions. This got things building again and worked for everything but my line charts which still render but no longer show any legends.

wyze:
> This got things building again and worked for everything but my line charts which still render but no longer show any legends.

https://github.com/plouc/nivo/blob/master/packages/line/src/Line.js#L193

You have to exclude `data` from legend prop or it will use it. Looking at the previous props, they should have been not required. Sorry about that.

jamesdh:
@wyze cool, I wasn't sure if you'd see this post-merge review :) 

> You have to exclude data from legend prop or it will use it. 

Yea, unfortunately that triggers a typescript error since it's required, and so removing it causes the build to fail compilation.

Issue #1519, opened 2021-05-14
GridLines & Axis rerendering on tooltip render


cameronevans:
**Describe/explain the bug**
GridLines & Axis components are rerendering every time the tooltip is rendered. This only happens with line 0.69.0 or later as the memoization was removed in the Typescript conversion:
GridLines: https://github.com/plouc/nivo/commit/73f98032b6faed65ea89b57ef9d5e1043c1a1acb#diff-38c896d16b0cfcea875ed07eb4f5dcef2e385be26c535bf48e943e7ce91604c2L77
Axis: https://github.com/plouc/nivo/commit/73f98032b6faed65ea89b57ef9d5e1043c1a1acb#diff-2005c198987521875392c5d373d9f345238d806c1b41e98312e202369c453fa3L191

**To Reproduce**
1. Go to https://codesandbox.io/s/charming-butterfly-4i46p
2. In the preview, click on "Open in new window"
3. Open React developer tools
4. Turn on "Highlight when components render"
5. GridLines and Axis components are highlighted on mouse over

**Expected behavior**
GridLines & Axis components should not rerender on mouse over

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Firefox
 - Version: 88.0.1
 
**Additional context**
The issue isn't shown by the CSB iframe preview's dev tools, the preview  must be in a new window so the browser dev tools can be used.

PR #1521, opened 2021-05-14
fix(axes): restore memoization

cameronevans:
fixes: #1519
restore memoization that was removed in https://github.com/plouc/nivo/commit/73f98032b6faed65ea89b57ef9d5e1043c1a1acb


wyze:
Thanks for the PR! It looks like we might be losing some type information. We'll need to get that fixed before we can merge this in.

wyze:
Thanks for this!

PR #1529, opened 2021-05-19
fix(swarmplot): fix swarmplot node onMouseMove event

Streeterxs:
ref https://github.com/plouc/nivo/issues/1530


wyze:
Good catch, thanks!

Issue #1337, opened 2020-12-10
Customize Bump yScale


danilofuchs:
**Is your feature request related to a problem? Please describe.**
I'm building a chart of rankings over time.
This is an annual ranking spanning 7 years. Each year, 20 competitors are chosen by an organisation.
In total, 67 distinct people have won a place in any ranking.

The way I modelled the data was the following: Each person is assigned a serie. If in a specific year this person is not featured in the rankings, I assign `null` to the data point on that year. Otherwise, we position them accordingly.

When plugging this data into `Bump`, it creates a chart with a y axis from 1 to 67, even though only 1 to 20 is populated with information.

**Describe the solution you'd like**
I would like to be able to modify `yScale` to limit axis to go from 1 to 20

**Describe alternatives you've considered**
- I could limit myself into using only 20 series and mixing data from multiple people, but that would severely hinder dynamic interaction and tooltips.
- I could build `Bump` myself based on this code: https://github.com/plouc/nivo/blob/master/packages/bump/src/bump/Bump.js, but that would be a large undertake

**Additional context**
![image](https://user-images.githubusercontent.com/17657014/101705900-c0c2d000-3a66-11eb-81c5-51ebb05c1180.png)

danilofuchs:
@wyze I currently have a deadline on finishing this project, and am able to contribute with Nivo on this issue. I just need some guidance.

What other charts/PRs have similar features?

wyze:
Seems like we might be able to do xScale/yScale like we do for line charts.

PR #1540, opened 2021-05-23
feat(bump): support more series than ranks

mtdavis:
If a bump chart has more series than ranks, you run into y-scale issues as described in [this closed issue](https://github.com/plouc/nivo/issues/1337). This PR fixes that by scaling the y-axis on the number of rank values, rather than the number of series.


wyze:
Thanks for this!

PR #1544, opened 2021-05-26
fix(packages): remove `global` from default props

targos:
This change removes the use of the `global` global variable, which
doesn't exist in browsers and is automatically polyfilled by Webpack.
Other bundlers such as Vite do not do that.
Check for the presence of `window` instead, which should work in more
environments.

### DIFF
--- a/packages/bar/src/props.js
+++ b/packages/bar/src/props.js
@@ -152,8 +152,7 @@ export const BarDefaultProps = {
 
     annotations: [],
 
-    pixelRatio:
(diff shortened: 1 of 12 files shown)


wyze (changes requested):
Looks like lint is failing on CI.

targos:
> Looks like lint is failing on CI.

Should be fixed

wyze (approved):
Looks good, thanks!

Issue #196, opened 2018-05-21
Legend Series Toggle


westonnh:
Can we have the legend series toggle on and off the data in the chart?

plouc:
@westonnh, unfortunately legends aren't interactives for now, but it can be implemented using your own legends, that's what I did on https://stateofjs.com/2017/connections/

plouc:
@westonnh, there's a work in progress on https://github.com/plouc/nivo/pull/224 which will add the ability to implement this feature.

serchavalos:
> [...] but it can be implemented using your own legends, that's what I did on https://stateofjs.com/2017/connections/

@plouc I guess you are referring to this part, right?

plouc:
@serchavalos, yes, that's the approach I've used for state of js

aiisahik:
I just figured out how to do this without having to build my own interactive legend component. The key is hidden inside of this storybook: https://nivo.rocks/storybook/?path=/story/waffle--legend-toggle




```
    <ResponsiveLine
                    data={ data }
                    colors={ this.state.colors || { scheme: 'red_yellow_green' } }
                    legends={[
                        {
                            onClick: (d) => {
                                const colors = data.map(series => (series.id === id ? '#ff3344' : '#DDD'));
                                this.setState({
                                    colors: colors,
                                });
                            },
                        },
                    ]}
    />
```

Once you have this, you can also implement a way for the user to deselect the series highlighting.

onionhammer:
@aiisahik That kinda works as a poor-mans version of this feature, but it doesn't work for stacked `ResponsiveLine`s,

fpezzola:
Is there a way to implements the `hiddenIds` in the ResponsiveLine chart?

sattuchin:
This feature would be very useful. Please prioritize when possible. Thank you!

shiftrtech:
Its this not possible yet? so much need it feature for multiple line charts

yuyoux:
Is it supported now?

fpezzola:
Can you add `hiddenIds` to Responsive Line?

peersky:
Really essential feature. Bump!

marcelojrfarias:
Bump! ;)

rafacarrenho:
Bump!

wyze:
[...]

PR #1556, opened 2021-05-29
feat(bar): add ability to toggle serie via legend item

wyze:
Reference #196

I couldn't figure out if there was a good way to get this to work when using the `legend.dataFrom: 'indexes'`, so that isn't supported.


shehi:
Just to confirm: this feature isn't supported in bar-canvas, correct? Coz right now, `toggleSerie` field of `legends` parameter has no effect on bar-canvas.

wyze:
Correct.

gluckzhang:
Hi @wyze , thanks for providing this nice feature! May I ask if it is possible to display the only data series by clicking the series on the legend? The current design is to hide one series when clicking it on the legend. In my case, I have many lines in one chart and I would like to highlight one of them. If I click the same item again, it could draw all the lines back.

Thank you and look forward to your answer!

PR #1562, opened 2021-06-02
feat(stream): add ability to customize legend label

### LINKED ISSUE #323: Customize legend labels in Stream component
Hello, I've read the docs for Stream and Legend but did not find anything regarding how to customize legend labels. 

I've a situation like this:
```javascript
const data = [
  {
    JD: 10500,
    PC: 6700,
	date: '2018-09'
  },
  {
    JD: 7860,
    PC: 5430,
    date: '2018-08'
  }
];

const sellers = [
  {
    id: 'JD',
    name: 'John Dorian'
  },
  {
    id: 'PC',
    name: 'Percival Cox'
  }
];

const keys = sellers.map(s => s.id);
```

There is a way to display the `name` of each seller inside the legend instead of their id apart from mapping the data array and change it to use seller's name as keys instead of their id?

wyze:
Close #323

### DIFF
--- a/packages/stream/index.d.ts
+++ b/packages/stream/index.d.ts
@@ -16,6 +16,7 @@ import {
     StackOffset,
     AreaCurve,
     SvgDefsAndFill,
+    PropertyAccessor,
 } from '@nivo/core'
 import { OrdinalColorScaleConfig, InheritedColorConfig } from '@nivo/colors'
 import { LegendProps } from '@nivo/legends'
@@ -55,6 +56,8 @@ declare module '@nivo/stream' {
         offsetType: StackOffset
         curve: AreaCurve
 
+        legendLabel: PropertyAccessor<T, string>
+
         margin: Box
 
         axisTop: AxisProps | null
@@ -68,15 +71,15 @@ declare module '@nivo/stream' {
         fillOpacity: number
 
         borderWidth: number
-        borderColor: InheritedColorConfig
+        borderColor: InheritedColorConfig<T>
 
         enableDots: boolean
         renderDot: StreamDotsItem
         dotPosition: 'start' | 'center' | 'end'
         dotSize: DatumToNumber | number
-        dotColor: InheritedColorConfig
+        dotColor: InheritedColorConfig<T>
         dotBorderWidth: DatumToNumber | number
-        dotBorderColor: InheritedColorConfig
+        dotBorderColor: InheritedColorConfig<T>
 
         isInteractive: boolean
         tooltipLabel: TooltipLabel<T>
--- a/packages/stream/index.d.ts
+++ b/packages/stream/index.d.ts
@@ -45,6 +45,30 @@ declare module '@nivo/stream' {
         y2: number
     }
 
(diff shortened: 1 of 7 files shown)


wyze:
Here is an example of intended functionality: https://codesandbox.io/s/nivo-forked-ove5e?file=/src/charts/Stream.tsx

website/src/data/components/stream/props.js
```diff
@@ -233,6 +233,18 @@ const props = [
...
         group: 'Dots',
     },
+    {
+        key: 'legendLabel',
+        help: 'Legend label accessor',
+        description: `
+            Define how to access the legend label of each datum,
+            by default, nivo will look for the \`id\` property.
+        `,
+        type: 'string | (datum: RawDatum): string',
```
plouc, line 243: I don't think the type of `datum` is correct as [we just pass the `id`](https://github.com/plouc/nivo/pull/1562/files#diff-44b3698944ddca2af8ad0c22a159233c92c88739e99d018b7a9b9f09f345dfb6R107) if I understand correctly, maybe we could use an inlined type such as `(datum: { id: string }): string`?

packages/stream/index.d.ts
```diff
@@ -55,6 +56,8 @@ declare module '@nivo/stream' {
         offsetType: StackOffset
         curve: AreaCurve
 
+        legendLabel: PropertyAccessor<T, string>
```
plouc, line 59: I don't think we can use `T` here.

packages/stream/index.d.ts
```diff
@@ -68,15 +71,15 @@ declare module '@nivo/stream' {
         fillOpacity: number
 
         borderWidth: number
-        borderColor: InheritedColorConfig
+        borderColor: InheritedColorConfig<T>
```
plouc, line 74: The structure you'll get when defining `borderColor` is:
```ts
interface ComputedDatum {
    id: number
    layer: {
        index: number
        x: number
        value: number
        y1: number
        y2: number
    }[] 
    path: string
    color: string
}
```

packages/stream/index.d.ts
```diff
@@ -68,15 +71,15 @@ declare module '@nivo/stream' {
...
 
         enableDots: boolean
         renderDot: StreamDotsItem
         dotPosition: 'start' | 'center' | 'end'
         dotSize: DatumToNumber | number
-        dotColor: InheritedColorConfig
+        dotColor: InheritedColorConfig<T>
```
plouc, line 80: I think custom accessors for the dots will receive the following structure:

```ts
interface DotDatum {
    key: string
    id: number
    color: string
    index: number
    x: number
    value: number
    y1: number
    y2: number
}
```

Apparently `Datum`, in the definitions file.

It's a combination of [what's computed for the layer points](https://github.com/plouc/nivo/blob/master/packages/stream/src/hooks.js#L91), and some extra properties passed in https://github.com/plouc/nivo/blob/master/packages/stream/src/StreamDots.js#L35.

plouc (changes requested):

plouc (approved):
LGTM

PR #1566, opened 2021-06-02
fix(pie): use legend.data property if supplied

### LINKED ISSUE #1439: Legends in Pie chart ignore data property
I found this problem while trying to setup legends on multiple lines (as in [this issue](https://github.com/plouc/nivo/issues/433))

A practical example can be found [here](https://codesandbox.io/s/elated-pike-lp7h8).

Line chart will render legends correctly with the passed custom data.
On the other hand, Pie will render legends with data passed to the chart, not to legends.

Is there a problem with my implementation? Is it not intended for Pie chart to render legends with custom data?

wyze:
Close #1439

### DIFF
--- a/packages/pie/src/PieLegends.tsx
+++ b/packages/pie/src/PieLegends.tsx
@@ -23,7 +23,7 @@ const PieLegends = <RawDatum,>({
                     {...legend}
                     containerWidth={width}
                     containerHeight={height}
-                    data={dataWithArc}
+                    data={legend.data ?? dataWithArc}
                 />
             ))}
         </>
--- a/packages/pie/tests/Pie.test.tsx
+++ b/packages/pie/tests/Pie.test.tsx
@@ -546,6 +546,41 @@ describe('Pie', () => {
                 )
             })
         })
+
+        it('should use legend.data if provided', () => {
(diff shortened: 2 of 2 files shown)


packages/pie/src/PieLegends.tsx
```diff
@@ -23,7 +23,7 @@ const PieLegends = <RawDatum,>({
                     {...legend}
                     containerWidth={width}
                     containerHeight={height}
-                    data={dataWithArc}
+                    data={legend.data ?? dataWithArc}
```
plouc, line 26: As [we already have some unit tests for the legends](https://github.com/plouc/nivo/blob/master/packages/pie/tests/Pie.test.tsx#L518), I think we could add a test case for this, what do you think?

plouc (changes requested):

packages/pie/src/PieLegends.tsx
wyze: Oh, yes definitely will add a test. Good catch.

plouc: Thanks!

plouc (approved):
LGTM

PR #1568, opened 2021-06-03
feat(network): add onClick support for node

### LINKED ISSUE #1351: New Feature - Network Chart - onClick functionality
A large client loves our Network visualization and wants to use it as a central piece of their UX - to make it truly impactful/useful they're requesting functionality that can only be performed with onClick (or, less so, a stationary/persistent tooltip). This is only a problem for us in that if we don't have it we may have to look for another visualization package or go custom d3 - we'd prefer to stick with Nivo and are willing to donate to get this functionality.

We would like onClick interactive functionality added to NetworkChart (Responsive / SVG) - such that we can capture the series data (in this case, in nodes data) and display it in a persistent layout (does not disappear onMouseOut).

We've tried providing the "deeper interactive" related data in a list next to the chart but this in essence requires duplication of effort - the user can see the information in the tooltip and then has to go find the record in the list in order to interact with that information.

**Additional context**
I see this offered in other charts so I'm hoping it's possible to achieve here. While we are a startup and are on a skeleton budget we see this as having the potential to move the needle on an account that could close our round of funding... so we're willing to lay out some donation if it makes a difference in getting this added. Thank you!

wyze:
Close #1351 

Additionally this brings missing tooltip functionality to the canvas flavor.

### DIFF
--- a/packages/network/index.d.ts
+++ b/packages/network/index.d.ts
@@ -75,6 +75,7 @@ declare module '@nivo/network' {
         theme?: Theme
 
         isInteractive?: boolean
+        onClick?: (node: ComputedNode, event: React.MouseEvent<SVGCircleElement>) => void
     }
 
     export interface NetworkSvgProps extends NetworkProps, MotionProps {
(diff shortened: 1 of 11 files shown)


plouc (approved):
LGTM, that's a great addition!

bwhthd:
Any indication on when this will be tagged/released?We're looking to implement this as well.

wyze:
I'll shoot for a release next week when I am back from vacation.

PR #1569, opened 2021-06-03
chore(core): revert ResizeObserver changes

### LINKED ISSUE #1561: Tooltip initial render incorrect on rapid mouse movement in bar graphs
The current Tooltip implementation incorrectly renders the tooltip initially before rendering it in the correct spot, causing an after-effect that is aesthetically unpleasing and is causing spacing issues in my implementation. Usually renders it to the right lower quadrant of my mouse when I move it between columns at a somewhat quick speed. 

To Reproduce this, go to any of the bar graphs on the storybook page and go back and forth between the columns. 

These tooltips should render in the correct spot, or have a delay that allows them to wait for mouse movement to stop.

This has been tested in chrome 90.0.4430.212

I quick fix would find a way to make the tooltip a onClick response.

wyze:
Reverts "fix(core): avoid occasional "ResizeObserver loop" error (#1466)"

This reverts commit 7f88e10da2b1eae178474c9f98e6d652d4a13f6e.

Close #1561 

I'm not sure we had a report about the ResizeObserver loop, but now that the linked issue was created the reverted code is the offender. If you try the latest released version of nivo you can easily reproduce the jumpy tooltip. 

@ollwenjones Sorry I had to revert your PR, but if you want to try another attempt at fixing the ResizeObserver loop error you were receiving. I'll be more than happy to review it and ensure our tooltip doesn't regress this time.

### DIFF
--- a/packages/core/src/hooks/useMeasure.js
+++ b/packages/core/src/hooks/useMeasure.js
@@ -3,35 +3,20 @@ import ResizeObserver from 'resize-observer-polyfill'
 
 export const useMeasure = () => {
     const measureRef = useRef(null)
-    const animationFrameId = useRef(null)
     const [bounds, setBounds] = useState({
         left: 0,
         top: 0,
         width: 0,
         height: 0,
     })
-    const [observer] = useState(
-        () =>
-            new ResizeObserver(([entry]) => {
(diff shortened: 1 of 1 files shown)


plouc (approved):
LGTM

PR #1570, opened 2021-06-04
fix(legends): update proptypes to avoid generated type error

### LINKED ISSUE #1567: Legends (LegendPropShape) -- No Exported Member (error TS2694)
**Describe/explain the bug**
The version used to produce this bug is 0.70.1. The NPM package csstype does not contain a member called AnimationIterationCountProperty. This causes issues when running the tsc. This is affecting the LegendPropShape type (props.d.ts)

`data: PropTypes.Requireable<(PropTypes.InferProps<{
        id: PropTypes.Validator<import("csstype").AnimationIterationCountProperty>;
        label: PropTypes.Validator<import("csstype").AnimationIterationCountProperty>;
        color: PropTypes.Requireable<string>;
        fill: PropTypes.Requireable<string>;
    }> | null)[]>;`

**To Reproduce**
Use the LegendPropShape type and attempt to run tsc --emitDeclarationOnly. This will result in a tsc error.
https://codesandbox.io/s/missing-cssimport-r1cxp?file=/package.json

1. Use legends in a plot, such as a bar plot
2. Execute tsc --emitDeclarationOnly

**Expected behavior**
Running tsc --emitDeclarationOnly does not cause this error.

 - OS: macOS
 - Browser N/A (this is a type bug, not runtime bug)
 - Version N/A (this is a type bug, not runtime bug)

**Additional context**
N/A

wyze:
Close #1567

Since we are actively removing PropTypes, I think this is an acceptable compromise. Most current legends don't support passing a custom `data` property to the legends config anyways.

### DIFF
--- a/packages/legends/src/props.ts
+++ b/packages/legends/src/props.ts
@@ -15,14 +15,7 @@ import PropTypes from 'prop-types'
  * ```
  */
 export const LegendPropShape = {
-    data: PropTypes.arrayOf(
-        PropTypes.shape({
-            id: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
-            label: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
-            color: PropTypes.string,
-            fill: PropTypes.string,
-        })
-    ),
+    data: PropTypes.arrayOf(PropTypes.object),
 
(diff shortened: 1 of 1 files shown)


plouc:
I agree, can be quite confusing.

PR #1571, opened 2021-06-04
fix(bar): fix inconsistency with custom tooltip

### LINKED ISSUE #1477: Custom tooltip inconsistency in Bar and Calendar
See tooltips here:
https://codesandbox.io/s/nivo-custom-tooltip-9p3is?file=/src/App.tsx

In above sandbox both bar and calendar use BasicTooltip

Calendar tooltip looks as expected.

In the bar chart the tooltip is nested inside a panel.
bar-tooltip

wyze:
Close #1477

### DIFF
--- a/packages/bar/src/BarItem.js
+++ b/packages/bar/src/BarItem.js
@@ -7,9 +7,8 @@
  * file that was distributed with this source code.
  */
 import { compose, pure, withPropsOnChange } from '@nivo/recompose'
-import React from 'react'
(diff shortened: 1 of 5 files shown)


plouc:
Maybe we could add a unit test for this as this package already has a test suite?

wyze:
> Maybe we could add a unit test for this as this package already has a test suite?

Good call. Done!

plouc (approved):
LGTM, Thank you!

PR #1574, opened 2021-06-05
feat(scatterplot): add ability to specify nodeId accessor

### LINKED ISSUE #1572: Scatterplot doesn't seem to support id / keys
Hi,

When using scatter plot data sources, I can't seem to find a way to pass through unique ids for SVG keys. That means that when I filter items, nivo animations loses track of items. e.g. if I'm plotting A, B, C, D and remove item B, rather than item B fading to transparant before removal, B appears to move to replace the item that was at location D. Not sure if this is because of a React or D3 keying issue but something isn't working under the hood.

I've tried providing both a unique 'id' and 'key' field in my data items, but that doesn't seem to be respected.

Perhaps there is a pattern I'm missing?

thanks!

wyze:
Close #1572

### DIFF
--- a/packages/scatterplot/src/ScatterPlot.js
+++ b/packages/scatterplot/src/ScatterPlot.js
@@ -41,6 +41,7 @@ const ScatterPlot = props => {
         colors,
         blendMode,
 
+        nodeId,
         nodeSize,
         renderNode,
 
@@ -87,6 +88,7 @@ const ScatterPlot = props => {
         yFormat,
         width: innerWidth,
         height: innerHeight,
+        nodeId,
         nodeSize,
         colors,
     })
--- a/website/src/data/components/scatterplot/props.js
+++ b/website/src/data/components/scatterplot/props.js
@@ -99,7 +99,7 @@ const props = [
 
             If you use a time scale, you must provide a time format
             as values are converted to Date objects.
-            
+
             Under the hood, nivo uses [d3-format](https://github.com/d3/d3-format),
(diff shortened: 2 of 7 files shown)


website/src/data/components/scatterplot/props.js
```diff
@@ -165,12 +165,24 @@ const props = [
...
             which will receive the raw value and should return the formatted one.
         `,
     },
+    {
+        key: 'nodeId',
+        group: 'Base',
+        defaultValue: defaults.nodeId,
+        type: 'string | (datum: NodeIdDatum): string',
+        help: `ID accessor for the node.`,
+        description: `
```
plouc, line 180: I think it's worth mentioning that the node ID impacts the animations, nodes are gonna transition from a previous state at the same index, thus, it can be useful to manipulate this property if we want to establish some continuity between nodes at different index.

packages/scatterplot/src/ScatterPlot.js
```diff
@@ -41,6 +41,7 @@ const ScatterPlot = props => {
         colors,
         blendMode,
 
+        nodeId,
```
plouc, line 44: Maybe we can add a unit test for this, what do you think?

plouc (changes requested):

plouc (approved):
LGTM

MilesParker:
Thanks Neil and Raphaël!

plouc:
Well, @wyze took care of all the hard work 😅

PR #1583, opened 2021-06-09
chore(deps): upgrade lodash to v4.17.21

### LINKED ISSUE #1581: Vulnerable version of lodash is used now
Hi, thanks for such an amazing library. 

I use the recent version of @nive/bar and the audit I run reported that it has a dependency to the vulnerable version of lodash. The patched version (>=4.17.21) should be used instead. It looks like it affects at least @nivo/bar, @nivo/annotations, @nivo/core, @nivo/colors, @nivo/scales, @nivo/axes and probably others I don't use right now.

https://www.npmjs.com/advisories/1673

Could you please update lodash to the advised version?

Thanks

wyze:
Close #1581


plouc (approved):
LGTM

Issue #1624, opened 2021-06-24
Docs: setting `isInteractive` breaks Bar chart


publicJorn:
**Describe/explain the bug**

When using the docs for bar chart when hovering the chart while `isInteractive: true` it breaks.

Long error message from React docs:  
`Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: null.`

Interestingly, when copying the generated code from the `code` tab and using it locally, there is no error. The tooltip works.

**To Reproduce**

- Go to https://nivo.rocks/bar/
- Make sure `isInteractive` is `true`
- Hover over the bar chart
- See the screen go blank and error messages in console
- Refresh the window
- Make sure `isInteractive` is `false`
- Hover over the bar chart
- Nothing happens (expected)

**Expected behavior**

It should not err when `isInteractive: true`

**Screenshots**

![Screenshot 2021-06-24 at 10 29 00](https://user-images.githubusercontent.com/1209869/123229911-07de5300-d4d7-11eb-9946-6e3952db9518.png)

**Desktop (please complete the following information):**
 - OS: osx
 - Browser: chrome, firefox
 - Version: latest

**Additional info**  
It does not break when checking the `custom tooltip example`

PR #1625, opened 2021-06-24
Fix missing tooltip breaking interactivity on bar chart doc

publicJorn:
Fix for: https://github.com/plouc/nivo/issues/1624

### DIFF
--- a/website/src/pages/bar/index.js
+++ b/website/src/pages/bar/index.js
@@ -146,7 +146,6 @@ const initialProperties = {
 
     isInteractive: true,
     'custom tooltip example': false,
-    tooltip: null,
 
     animate: true,
     motionStiffness: 90,


plouc:
@publicJorn, thank you for spotting this issue and proposing a fix! I think this now breaks the custom tooltip example.

wyze:
Like plouc mentioned, this does break the custom tooltip example. I went ahead and created the fix in #1627 and deployed it. The website code can be quite confusing on it's setup when trying to make changes.

jorn-cloudnatix:
Ah, the proper fix was a bit deeper down the rabbit hole. Thanks for the quick response guys!

publicJorn:
Ah, the proper fix was a bit deeper down the rabbit hole. Thanks for the quick response guys!

PR #1634, opened 2021-06-25
feat(calendar): expose weekdayTicks prop on TimeRange component

brammitch:
On the new TimeRange component, the weekday tick marks are set to Monday/Wednesday/Friday. This PR gives the user an option to change them, or remove them altogether by passing an empty array.

Updates were made to the Storybook showing both alternative options.


packages/calendar/src/types.ts
```diff
@@ -209,6 +209,7 @@ export type TimeRangeSvgProps = Dimensions & { data: CalendarDatum[] } & Partial
                 square: boolean
                 role: string
                 weekdayLegendOffset: number
+                weekdayTicks: number[]
```
wyze, line 212:
```suggestion
                weekdayTicks: Array<0 | 1 | 2 | 3 | 4 | 5 | 6>
```

Let's update the type to be more strict.

website/src/data/components/time-range/props.js
```diff
@@ -228,6 +214,24 @@ const props = [
             max: 100,
         },
     },
+    {
+        key: 'weekdayTicks',
+        help: 'define weekday tickmarks to show',
+        type: 'number[]',
```
wyze, line 220: Update this type here as well.

wyze (changes requested):
Looks good, just what a more strict type.

Also, can you merge master or rebase. I merged in a PR containing the changes you have in `.storybook` folder.

website/src/data/components/time-range/props.js
```diff
@@ -29,20 +29,6 @@ const props = [
...
-        type: 'string | Date',
-        required: true,
-    },
-    {
-        key: 'to',
-        group: 'Base',
-        help: 'end date',
-        type: 'string | Date',
-        required: true,
-    },
```
wyze, line 45: Thanks for removing my copy/paste error. 👍

brammitch: No problem! 👍

website/src/data/components/time-range/props.js
brammitch: Made changes as suggested

packages/calendar/src/types.ts
brammitch: Made changes as suggested

wyze (approved):
Thanks!

PR #821, opened 2020-01-12
WIP: Typescript migration

plouc:
## Migrate the whole code base to TypeScript  🎢💥 🔥 

⚠️ We decided to tackle this incrementally, you can track the progress here: https://github.com/plouc/nivo/projects/1

### Infrastructure

- [x] update rollup config to support TS
- [ ] update rollup config to generate minified code (terser?)
- [ ] make storybook using TS
- [ ] update TS example to CRA
- [x] Remove nivo custom babel config to use babel-env only
- [ ] Get rid of circular dependencies between packages

#### Core packages

- [x] annotations
- [x] axes
- [x] colors
- [x] core
- [x] generators
- [x] legends
- [x] tooltip

### Charts

- [x] bar
- [x] bullet
- [ ] bump
- [x] calendar
- [ ] chord
- [x] circle-packing
- [x] funnel
- [ ] geo
- [ ] heatmap
- [ ] line
- [ ] network
- [ ] parallel-coordinates
- [x] pie
- [x] radar
- [x] sankey
- [x] scales
- [x] scatterplot
- [x] stream
- [x] sunburst
- [x] swarmplot
- [ ] treemap
- [x] voronoi
- [ ] waffle

PR #1636, opened 2021-06-29
fix(chord): fix missing typescript properties

flluch:
- legends was missing
- ribbonBorderColor was missing the type

refs #821


wyze:
Thanks!

PR #1707, opened 2021-08-04
Fix: ResponsiveSwarmPlotCanvas does not work with valueScale type time

### LINKED ISSUE #1705: ResponsiveSwarmPlotCanvas does not work with valueScale type "time"
**Describe/explain the bug**
I am unable to use `ResponsiveSwarmPlotCanvas` for plotting a dataset with `valueScale` type set to `time`. The Axis just shows some strange time stamps which are neither related to the timestamp in the dataset or anything else. 

**To Reproduce**
I have implemented a simple example with `ResponsiveSwarmPlotCanvas` where the X-Axis is picking up some random time stamps.
https://codesandbox.io/s/affectionate-mahavira-j5nx0?file=/src/App.js

However the same example works perfectly fine when I switch to `ResponsiveSwarmPlot`
https://codesandbox.io/s/frosty-williamson-5uhbe?file=/src/App.js

Both examples are exactly the same, the only difference is the component being used. 

1. Go to the sandbox: https://codesandbox.io/s/affectionate-mahavira-j5nx0?file=/src/App.js
2. The X-Axis has some random time stamp values

**Expected behavior**
`ResponsiveSwarmPlotCanvas` handles `valueScale` type `time` same as `ResponsiveSwarmPlot`.

**Screenshots**
X-Axis is using strange time stamp values in case of `ResponsiveSwarmPlotCanvas`
image

The same example works perfectly fine when we replace `ResponsiveSwarmPlotCanvas` with `ResponsiveSwarmPlot`
image

 - Windows 10
 - Chrome Version 92.0.4515.107
 - "@nivo/core": "0.73.0",
    "@nivo/swarmplot": "0.73.0",

**Additional context**
I need to switch from `ResponsiveSwarmPlot` to `ResponsiveSwarmPlotCanvas` in order to get better performance when rendering a rather large number of nodes.

Patil2099:
Fixes #1705


therightstuff:
Are approvals by non-maintainers welcome? The fix looks correct to me 👍

packages/swarmplot/src/SwarmPlotCanvas.tsx
```diff
@@ -100,7 +101,7 @@ export const InnerSwarmPlotCanvas = <RawDatum,>({
         id,
         value,
         valueFormat,
-        valueScale: defaultProps.valueScale,
+        valueScale: valueScale || defaultProps.valueScale,
```
plouc, line 104: I'd suggest to handle this the same way other defaults are managed, this can be confusing if we change this at some point (`valueScale = defaultProps.valueScale`).

plouc (changes requested):

packages/swarmplot/src/SwarmPlotCanvas.tsx
```diff
@@ -54,6 +54,7 @@ export const InnerSwarmPlotCanvas = <RawDatum,>({
     id = defaultProps.id,
     value = defaultProps.value,
     valueFormat,
+    valueScale,
```
therightstuff, line 57:
```suggestion
    valueScale = defaultProps.valueScale,
```

packages/swarmplot/src/SwarmPlotCanvas.tsx
therightstuff:
```suggestion
        valueScale,
```

plouc (approved):

PR #1712, opened 2021-08-09
Bar renderToStatic or renderToString ends up in empty chart

LiamMartens:
can be fixed with initial: null


n-moser:
This is really a bug. Also the bar chart sample of the API returns an empty chart.

OleksandrRakovets:
Why is it closed? The issue still persists on the latest version. Please merge.

plouc:
@OleksandrRakovets, this cannot be merged because the repository has been deleted.

PR #1719, opened 2021-08-13
Improve SVG bar chart accessibility

plouc:
- Add support for various aria attributes to the SVG bar chart
- Add support for various aria attributes to the core SVG wrapper
- Mark SVG defs as hidden for screen readers
- Make bar items focusable and trigger tooltip when focused
- Use a single React hook to handle both bar SVG and Canvas implementations
- Improve the accessibility of the website component tabs used to switch between chart/code/data
- Allow markdown for components' help content
- Do not use `target="_blank"` for links in markdown as it's generally considered not to be the best practice
- Update `caniuse-lite`


bradlocking:
Hey Raphaël - Thanks for adding accessibility support to the bar chart component! We're currently working with a few other charts (Pie, Sunburst) and was wondering your thoughts on the complexity of getting broader accessibility support?

Issue #192, opened 2018-05-19
Add support for custom tooltip


plouc:
Allowing custom tooltips has been requested several times:

https://github.com/plouc/nivo/issues/117 | https://github.com/plouc/nivo/issues/186 | https://github.com/plouc/nivo/issues/120

This issue is used to track down the ongoing implementation of custom tooltips on all nivo packages/components:

- [x] `@nivo/bar` *Responsive***Bar** `>=0.36.0`
- [x] `@nivo/bar` *Responsive***BarCanvas** `>=0.36.0`
- [x] `@nivo/bullet` *Responsive***Bullet** - https://github.com/plouc/nivo/pull/1694
- [x] `@nivo/calendar` *Responsive***Calendar** - https://github.com/plouc/nivo/pull/223
- [x] `@nivo/chord` *Responsive***Chord**
- [x] `@nivo/chord` *Responsive***ChordCanvas**
- [x] `@nivo/circle-packing` *Responsive***Bubble** - https://github.com/plouc/nivo/pull/200
- [x] `@nivo/circle-packing` *Responsive***BubbleHtml** - https://github.com/plouc/nivo/pull/200
- [x] `@nivo/circle-packing` *Responsive***BubbleCanvas** - https://github.com/plouc/nivo/pull/200
- [x] `@nivo/heatmap` *Responsive***HeatMap**
- [x] `@nivo/heatmap` *Responsive***HeatMapCanvas**
- [x] `@nivo/line` *Responsive***Line** - https://github.com/plouc/nivo/pull/258
- [x] `@nivo/pie` *Responsive***Pie** - https://github.com/plouc/nivo/pull/208
- [x] `@nivo/pie` *Responsive***PieCanvas**
- [x] `@nivo/radar` *Responsive***Radar** - https://github.com/plouc/nivo/pull/1761
- [x] `@nivo/sankey` *Responsive***Sankey**
- [x] `@nivo/scatterplot` *Responsive***ScatterPlot** - https://github.com/plouc/nivo/pull/255
- [x] `@nivo/scatterplot` *Responsive***ScatterPlotCanvas**
- [x] `@nivo/stream` *Responsive***Stream** - https://github.com/plouc/nivo/pull/1721
- [x] `@nivo/sunburst` *Responsive***Sunburst**
- [x] `@nivo/treemap` *Responsive***TreeMap** - https://github.com/plouc/nivo/pull/200
[...]

PR #1721, opened 2021-08-15
Migrate stream package to TypeScript

plouc:
Along with the TypeScript migration, I also made the following changes:
- Replaced `tooltipFormat` with `valueFormat`, and add a `formattedValue` property on each computed datum for potential reuse
- Replaced `legendLabel` and `tooltipLabel` with a single `label` property, I don't think we need 2 different properties for this
- Added support for layers
- Added basic unit tests
- Added support for aria attributes on the SVG container
- Added support for custom tooltip and stack tooltip


wyze (changes requested):
Looks good. Build is erroring because of formatting in the test file.

plouc:
> Looks good. Build is erroring because of formatting in the test file.

@wyze, thx for the review, I fixed the formatting issue and added support for custom tooltips to complete another item from https://github.com/plouc/nivo/issues/192.

PR #1722, opened 2021-08-18
feat(timerange): add from, to, and emptyColor props

vgeary96:
Updated the `<TimeRange/>` component to support custom date ranges irregardless of the data provided via optional `from` and `to` props. Also added support for gaps in the data and an `emptyColor` prop for those gaps.


wyze (approved):
Looks great, thanks!

Issue #859, opened 2020-02-07
y-axis - auto scale with holes in data


axlider:
Hi,

I use "yScale={{ type: 'linear', stacked: false, min: 'auto', max: 'auto' }}" to auto scale y axis.
Unfortunately when I combine this with "holes in data",  auto-scaling does not work anymore and the hole y-axis appears.
Unfortunately the null values on y axis are responsible for this.
Shouldn't only '0' values influence auto-scaling and not null values?

I think the problem is line 126 in 'packages/scales/src/compute.js':
` min = Math.min(...all)`  
because `Math.min([null])` resolves to '0'.

I think 'all' should be filtered before calculating y.
Someting like:

```
const validAll = all.filter(val =>  val !== null);
min = Math.min(validAll);

```
Thanks a lot!

jhonabreul:
Hey! Has anybody check this out? Please, I have the same issue

renato:
I'm facing the same issue, I used a similar workaround as the OP but I believe "auto" should account for null values.

PR #1729, opened 2021-08-23
Fix autoscale of linechart when data have holes.

PierrVanRoy:
The yScale.auto option is not properly computed when the data have holes and is positive. 

This PR filters those hole from the values of the chart so tat the min is not always 0 in that case.

Before:
image

After:
image

Reported also by: https://github.com/plouc/nivo/issues/859

### DIFF
--- a/packages/scales/src/compute.ts
+++ b/packages/scales/src/compute.ts
@@ -195,7 +195,11 @@ export const generateSeriesAxis = <Axis extends ScaleAxis, Value extends ScaleVa
 
     switch (scaleSpec.type) {
         case 'linear': {
-            const all = sortBy(uniq(values as number[]).filter(v => v !== null), v => v)
+            const all = sortBy(
+                // filer null values to deal with holes in linechart
+                uniq(values as number[]).filter(v => v !== null),
+                v => v
+            )
 
             return { all, min: Math.min(...all), max: Math.max(...all) }
         }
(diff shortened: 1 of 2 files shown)


packages/scales/src/compute.ts
```diff
@@ -195,7 +195,7 @@ export const generateSeriesAxis = <Axis extends ScaleAxis, Value extends ScaleVa
 
     switch (scaleSpec.type) {
         case 'linear': {
-            const all = sortBy(uniq(values as number[]), v => v)
+            const all = sortBy(uniq(values as number[]).filter(v => v !== null), v => v)
```
plouc, line 198: I think we should add a unit tests for this, also maybe a comment explaining why we filter out null values, what do you think?

plouc (approved):

plouc:
@PierrVanRoy, thank you for this fix, nice UI btw :) Is that a public project?

plouc (approved):
LGTM

PierrVanRoy:
@plouc , we are using Nivo quite a lot at [Pigment](https://www.gopigment.com/) but sadly I have no public access to share with you :-( .
BTW, thanks a lot for this project. It's a real pleasure to use it!

PierrVanRoy:
hello @plouc,

Do you have any visibility on when this fix will be include a nivo release? 🙏 
Just want to know if we should patch on our side or if we ca wait for an official release. 

Thanks 

Pierre

plouc:
@PierrVanRoy, unfortunately, I'm pretty busy with work and personal matters at the moment, so it's unlikely that I'll release the fix soon.

Issue #1728, opened 2021-08-23
Bump d3-color


foreseeti-adam:
Hi,

d3-color version<3.0.0 that is used in some of your packages is vulnerable to [ReDoS](https://snyk.io/vuln/SNYK-JS-D3COLOR-1076592), hence it would be a good idea to bump this ([if possible](https://github.com/d3/d3-color/releases/tag/v3.0.0)) in your next release.

Thank you for providing a great visualization tool!

plouc:
Hi @foreseeti-adam, thank you for reporting this issue, as it's a major version change, we would have to investigate a little bit before migrating. If the only change is the adoption of modules, shouldn't be an issue though as we already use ESM.

huornlmj:
Bump,  d3-color@2.0.0 has a Regular Express Denial of Service vulnerability via the rgb() and hrc() functions since August 2020.  Popular composition analysis tools are thus flagging nivo as introducing a vulnerable 3rd party component.

dnoji:
The d3-color package is still be flagged by [security tools](https://security.snyk.io/vuln/SNYK-JS-D3COLOR-1076592). Can this be upgraded to 3.1.0 in the next release?

marcospassos:
Hey guys, any news on this?

B0lemix:
still being flagged like vulnerable to ReDoS and no fix available

IshfaqAhmedProg:
@plouc any fix for this?

dnoji:
> The d3-color package is still being flagged by [security tools](https://security.snyk.io/vuln/SNYK-JS-D3COLOR-1076592). Can this be upgraded to 3.1.0 in the next release?

I think it might have been fixed in this [release](https://github.com/plouc/nivo/releases/tag/v0.82.0) by #2142

jvu1:
The vulnerability still exists for me because two of nivo's dependencies: `d3-interpolate@2.0.1` and `d3-scale-chromatic@2.0.0` still use version 2.0.0 of `d3-color`.

Anybody still getting the flagged after updating to `v0.82.0` or greater?

RikuRuokonen:
This seems still be issue?

It seems that  `/d3-scale-chromatic@2.0.0` still uses v.2.0.0 of `d3-color`.

Subsequently,  `/@nivo/core@0.83.0` (latest one) seems to use said version in `d3-scale-chromatic`.

adrian-meditect:
[...]

PR #1743, opened 2021-08-28
Upgrade d3 color

plouc:
fix https://github.com/plouc/nivo/issues/1728

Also upgrade the Node.js version used on CI to `14` (active), as `12` is now in maintenance mode and not compatible with the new `d3-color` version, I couldn't enable `16` at the moment because of an incompatible package used for the website (`sharp`), I think we can enable it when upgrading gatsby and its plugins.


plouc:
Had too many issues with `d3-color` upgrade, closing this for now.

PR #1754, opened 2021-09-01
Modifying Radar's Axis to be drawn by Axis.domain.line of the Theme

GoodGoodJM:
It seems that Radar's Axis is not currently being drawn by referring to the theme incorrectly.
Modify to draw axis using `theme.axis.domain.line`.

### DIFF
--- a/packages/radar/src/RadarGrid.js
+++ b/packages/radar/src/RadarGrid.js
@@ -34,7 +34,7 @@ const RadarGrid = memo(({ indices, levels, shape, radius, angleStep, label, labe
                         y1={0}
                         x2={position.x}
                         y2={position.y}
-                        {...theme.grid}
+                        {...theme.axis.domain.line}
                     />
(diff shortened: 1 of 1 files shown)


plouc:
@GoodGoodJM, that's expected, this is the equivalent of the cartesian grid, not an axis.

plouc:
But there's a bug in the current implementation, should be `theme.grid.line`.

GoodGoodJM:
ok but, if use theme.gird.line, it change all line's styles of radar. If we does not want axis, all line will remove..

plouc:
@GoodGoodJM, yes, both are part of the grid, it's the same behavior as other charts, using axis is not correct. We could add another option to disable those.

plouc (approved):
LGTM

plouc:
Thank you for this fix @GoodGoodJM!

PR #1756, opened 2021-09-02
Fix: Avoid duplicate key error with time scales at millisecond precision

protyze:
When using the Line-Chart as a timeline with Date-Objects on the X-Axis, the AxisTicks and GridLines are using a string representation of the x-Values' Date-Object as the react element's "key". This causes duplicate keys when the timelines scale is below 1 second intervals (since a date-objects string representation does only include seconds, for example: "Thu Sep 16 2021 09:40:59 GMT+0100 (Western European Summer Time)").

This is a small PR to use the timestamp (.valueOf()) as the react element's key for AxisTicks and GridLines, so that the Line Chart works correctly with milliseconds precision.

### DIFF
--- a/packages/axes/src/compute.ts
+++ b/packages/axes/src/compute.ts
@@ -230,8 +230,8 @@ export const computeCartesianTicks = <Value extends AxisValue>({
         }
     }
 
-    const ticks = values.map(value => ({
-        key: typeof value === 'number' || typeof value === 'string' ? value : `${value}`,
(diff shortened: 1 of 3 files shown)


netzwerg:
This is currently blocking some of our UI work – how can we support you in moving this PR forward?

netzwerg:
bump

plouc:
Good catch! And sorry for the late review.

Issue #1785, opened 2021-09-28
Error using legend label customizing in Radar


minemanemo:
**Describe/explain the bug**
- We can't customize legend label in Radar component ([go source](https://github.com/plouc/nivo/blob/master/packages/radar/src/hooks.ts#L80))
- Type document and issue said "you can customize legend label!!!" ([go type docs 1](https://github.com/plouc/nivo/blob/master/packages/radar/src/types.ts#L126), [go type docs 2](https://github.com/plouc/nivo/blob/master/packages/legends/src/types.ts#L67), [go type docs 3](https://github.com/plouc/nivo/blob/master/packages/legends/src/types.ts#L58), [go issue](https://github.com/plouc/nivo/issues/791#issuecomment-564522268))

**To Reproduce**

1. Go to [codesandbox](https://codesandbox.io/s/nivo-radar-bug-report-ftrmw)
2. Open `src/App.tsx` and see 31 line
3. legends props's data property is not working...
4. See browser

**Expected behavior**
- expected results(Legends): `🟠 My consumption` `🔵 Average` `🟢 Top 11%`
- actual result(Legends): `🟠 myc` `🔵 avg` `🟢 top`

**Screenshots**
- Please check codesandbox

**Desktop (please complete the following information):**
 - OS: macOS Big Sur 11.5.2
 - Browser: chrome
 - Version: 93.0.4577.82

**Additional context**
I'll fork and pull request

plouc:
@minemanemo it's just supported for a few chart types.

PR #1786, opened 2021-09-28
fix(radar): improve radar's legend data to be customizable

minemanemo:
Improve Radar component's legend data to be customizable.

[Related issues](https://github.com/plouc/nivo/issues/1785)

### DIFF
--- a/packages/radar/src/Radar.tsx
+++ b/packages/radar/src/Radar.tsx
@@ -68,7 +68,7 @@ const InnerRadar = <D extends Record<string, unknown>>({
         centerY,
         angleStep,
         curveFactory,
-        customLegends,
+        boundLegends,
         customLayerProps,
     } = useRadar<D>({
         data,
@@ -173,7 +173,7 @@ const InnerRadar = <D extends Record<string, unknown>>({
     if (layers.includes('legends')) {
         layerById.legends = (
             <Fragment key="legends">
-                {customLegends.map((legend, i) => (
+                {boundLegends.map((legend, i) => (
                     <BoxLegendSvg
                         key={i}
                         {...legend}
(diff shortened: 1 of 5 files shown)


packages/radar/src/hooks.ts
```diff
@@ -83,6 +85,11 @@ export const useRadar = <D extends Record<string, unknown>>({
         color: colorByKey[key],
     }))
 
+    const customLegends = legends.map(legend => ({
```
plouc, line 88: I think this should be memoized, it could lead to unnecessary re-render otherwise.

packages/radar/src/hooks.ts
```diff
@@ -83,6 +85,11 @@ export const useRadar = <D extends Record<string, unknown>>({
...
+        ...legend,
+        data: legend?.data?.map(d => ({ legendData, ...d })) || legendData,
```
plouc, line 90: Why passing `legendData` (in `.map(d => ({ legendData, ...d }))`)? it seems a bit strange especially if passed for each datum.

packages/radar/src/Radar.tsx
```diff
@@ -172,13 +173,12 @@ const InnerRadar = <D extends Record<string, unknown>>({
     if (layers.includes('legends')) {
         layerById.legends = (
             <Fragment key="legends">
-                {legends.map((legend, i) => (
+                {customLegends.map((legend, i) => (
```
plouc, line 176: I think this new feature should have a unit test.

packages/radar/src/hooks.ts
plouc, line 88: What about `boundLegends` instead? those legends aren't necessarily custom, just bound to the data.

plouc (changes requested):

packages/radar/src/hooks.ts
minemanemo: Misunderstanding caused a wrong legend data. 
Added unit test case.

packages/radar/src/hooks.ts
minemanemo: that's good idea. :)

PR #1793, opened 2021-10-06
fix(bar): avoid to redefine react forwardRef type

adriengibrat:
Avoid to redefine react type ;)

The previous definition was problematic when using proptypes in a typescript file: 

Any typescript codebase importing `@nivo/bar` suddendly can't attach `propTypes` to the returned value of `forwardRef` because they are now just a function and not a `ForwardRefExoticComponent` anymore... unexpected & very disturbing.

This resulted in breaking build with typescript errors!

The symptomatic error is ts(2339):
> Property 'propTypes' does not exist on type '(props: ...) => ReactElement<...> | null'.


plouc:
Thank you @adriengibrat!

Issue #1649, opened 2021-06-30
Add to Sunburst props "innerRadius" similar to Pie props.innerRadius


derrickbeining:
**Is your feature request related to a problem? Please describe.**
I need to be able to customize the size of the center hole in `Sunburst`. I noticed that the `Pie` component is able to do this with the `innerRadius` prop. Would it be possible to expose a similar customization to `Sunburst`?

**Describe the solution you'd like**
Add `innerRadius: number` prop to `SunburstProps` to expose control over the radius of the center hole in a `Sunburst`.

**Describe alternatives you've considered**
Retiring to the countryside.

**Additional context**
Screen capture of `Pie` chart doing what I want `Sunburst` to be able to do:

https://user-images.githubusercontent.com/25019711/124044076-8ad14300-d9da-11eb-9c6d-7c45b2eef8c0.mov

adamberro:
Hi, I am looking for the same functionality. I would be happy to spend some time on this to PR a change. Can you provide any suggestions/guidance for how to approach this?

jsw:
Bump. I'd also be interested in retrieving the inner radius as a property in `SunburstCustomLayerProps` so that I can render a circle in the center.

petrvecera:
There seems to be PR implementing this feature 👍  https://github.com/plouc/nivo/pull/2219

PR #1794, opened 2021-10-07
feat(sunburst): add innerRadiusRatio, renderRootNode prop to control size of inner circle

adamberro:
* Adds a prop similar to the pie chart's `innerRadius` prop to change the size of the sunburst chart's inner circle.
* Adds a prop to render the root node in the center of the chart.

Fixes https://github.com/plouc/nivo/issues/1649

Please let me know if this approach works or if it would be more desirable to give more flexibility such as the option to configure the type of d3 scale, or if there are any other issues or anything else is missing.


adamberro:
Hi @plouc, anything we can do to get the review process started? Thank you!

vicaub:
I really like this feature and need it on my project too!
But probably it would look better if all layer had the same radius length

packages/sunburst/src/hooks.ts
```diff
@@ -68,6 +75,12 @@ export const useSunburst = <RawDatum>({
         // are going to be computed first
         const sortedNodes = sortBy(descendants, 'depth')
 
+        const innerRadiusOffset = radius * Math.min(innerRadiusRatio, 1)
+
+        const maxDepth = Math.max(...sortedNodes.map(n => n.depth))
+
+        const scale = d3ScaleRadial().domain([0, maxDepth]).range([innerRadiusOffset, radius])
```
vicaub, line 82: using d3 scaleLinear instead of scaleRadial would render better in my opinion

adamberro:
Hi @plouc, just checking in on whether we might be able to merge this. I'm happy to make any changes as needed. Thanks!

packages/arcs/src/centers.ts
```diff
@@ -45,6 +45,11 @@ export const useArcCentersTransition = <Datum extends DatumWithArc, ExtraProps =
     mode: ArcTransitionMode = 'innerRadius',
     extra?: TransitionExtra<Datum, ExtraProps>
 ) => {
+    // center label of root node
```
adamberro, line 48: This centers the root most label. Not sure if there is a more proper way to handle this, but I don't think it was accounted for originally

Screen Shot 2021-11-26 at 9 09 56 AM
.

packages/sunburst/src/hooks.ts
adamberro: scaleLinear looks quite different from the original implementation: 
Screen Shot 2021-11-26 at 9 57 09 AM

Maybe we could update the sunburst props to include setters for the arc inner/outer radii. I'm not sure if this reaches a happy medium between customizability and ease of use, but my initial thought for the api would be something like: 
```
type GetInnerRadius = (currentNode: HierarchyRectangularNode<RawDatum>, nodes: HierarchyRectangularNode<RawDatum>[]) => number
```
@plouc Thoughts on this approach?

plouc:
@adamberro, sorry for the late reply, I'll try to run this branch locally and review the PR in the upcoming weeks.

lers92:
Also looking forward to this new feature 😄

tkonopka:
bump

rnsv:
bump

TotomiEcio:
Hi! I hope this MR is able to merge. I would like this changes. Thank you

plouc:
Superseded by https://github.com/plouc/nivo/pull/2696.

Issue #1771, opened 2021-09-09
[ResponsiveLine] Typescript error "orient does not exist"


Paul6552:
**Describe/explain the bug**
Typescript is complaining that the property "orient" does not exist on axisBottom or axisLeft.

**To Reproduce**
Go to [Nivo Webpage Line](https://nivo.rocks/line/)

Take the generated code and paste it into your development area. (e.g. Intellij)

**Expected behavior**
"No typescript error"

**Screenshots**
![Bildschirmfoto vom 2021-09-09 09-29-15](https://user-images.githubusercontent.com/25934591/132642149-97452992-1861-4215-9530-d9e69037bd02.png)

**context**
"@nivo/line": "^0.73.0",
"@nivo/core": "^0.73.0",
Browser: Firefox latest or Chrome latest

DoneDeal0:
@Paul6552 , @plouc I sent a pr to fix this issue:  [#1797](https://github.com/plouc/nivo/pull/1797)

plouc:
@NathaEEUD, you're right, it shouldn't be exposed, and if we were to support overriding it, we would have to create different types for horizontal & vertical axes anyway, for example `orient='left'` would not make sense for an horizontal axis.

alb7h:
Is this a documentation problem then? The default [example](https://nivo.rocks/line) shows this in the example code.

plouc:
This has been removed.

PR #1797, opened 2021-10-08
fix: add orient prop to axisprops

DoneDeal0:
fix issue [#1771](https://github.com/plouc/nivo/issues/1771).


MarcCelani-at:
Can we reopen this? We are running into this as well.

PR #1807, opened 2021-10-18
feat(radar): add support for patterns and gradients

adamberro:
Hi, we currently use this chart but need to support patterns with it. 
This PR adds defs and fill props to radar chart to support patterns and gradients.

Screen Shot 2021-10-18 at 2 53 18 PM


plouc:
Thank you for your contribution @adamberro! It will be included in the upcoming release.

PR #1812, opened 2021-10-27
feat(defs): allow gradientTransform to be passed to gradient definitions

### LINKED ISSUE #1098: Gradient rotation
In the line chart, there are defs and fill properties, that allows defining and assigning gradients for chart series. Unfortunately, there is only horizontal gradient allowed. It would be a great if you could extend the LinearGradientDef with rotation in degrees (ex. 90, 45 etc.)

I would need to create a chart like this: 
image

Green series should have a vertical gradient

benjamintd:
Fixes #1098

This PR allows defining gradient transforms as part of the `defs` for linear gradients. 
This fixes the (now stale, but still relevant) associated feature request #1098. 

Example usage: 

```tsx
 <Line
        {...commonProperties}
        enableArea={true}
        yScale={{
            type: 'linear',
            stacked: true,
        }}
        curve={select('curve', curveOptions, 'linear')}
        defs={[
            linearGradientDef(
                'gradientA',
                [
                    { offset: 0, color: 'inherit' },
                    { offset: 100, color: 'inherit', opacity: 0 },
                ],
                { gradientTransform: 'rotate(90 0.5 0.5)' }
            ),
        ]}
        fill={[{ match: '*', id: 'gradientA' }]}
    />
```

The transform attribute follows the SVG spec documented here: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/transform 

I added a test and modified the gradient example to show how to use that new property, which for instance allows horizontal gradients on lines: 
image

### DIFF
--- a/packages/core/index.d.ts
+++ b/packages/core/index.d.ts
@@ -255,6 +255,7 @@ declare module '@nivo/core' {
             color: string
             opacity?: number
         }[]
+        gradientTransform?: string
     }
 
     export type PatternDotsDef = {
--- a/packages/core/src/components/defs/gradients/LinearGradient.js
+++ b/packages/core/src/components/defs/gradients/LinearGradient.js
@@ -8,8 +8,8 @@
  */
 import PropTypes from 'prop-types'
 
-export const LinearGradient = ({ id, colors }) => (
-    <linearGradient id={id} x1={0} x2={0} y1={0} y2={1}>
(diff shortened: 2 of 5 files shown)


plouc (approved):
LGTM

plouc:
@benjamintd thank you for your contribution! That's a nice addition to the lib.

PR #1814, opened 2021-10-27
fix(swarmplot): use the borderWidth property - fixes #1813

### LINKED ISSUE #1813: borderWidth is not being used in swarm plot
The border width property of a swarm plot is not being used in the component.

**To Reproduce**
Go to https://nivo.rocks/swarmplot/ and move the `borderWidth` input. Nothing happens on the graph.

**Expected behavior**
The border width of the circles is affected by the `borderWidth` setting.

**Screenshots**
image

**Cause** 

The width is hardcoded to 0 here: https://github.com/plouc/nivo/blob/399215b6e1d00274a0a8278fa64a6152f7079520/packages/swarmplot/src/SwarmPlot.tsx#L144

~I suspect this has to do with the use of `d3-force` which needs to know about the width of the elements.~
EDIT: the d3-force positions take the borderWidth into account, so passing down the prop seems to suffice to display the borders. The canvas rendering do not use the `borderWidth` property, however.

I'll try to submit a PR soon to propose a solution.

benjamintd:
This PR fixes #1813 by passing down the borderWidth property which was omitted in the Swarmplot component.

The result can be seen in the story at `/story/swarmplot--extra-layers` which defined a borderWidth which was not being used. It is now displaying as intended: 
image

### DIFF
--- a/packages/swarmplot/src/SwarmPlot.tsx
+++ b/packages/swarmplot/src/SwarmPlot.tsx
@@ -36,6 +36,7 @@ const InnerSwarmPlot = <RawDatum,>({
     colors = defaultProps.colors as OrdinalColorScaleConfig<Omit<ComputedDatum<RawDatum>, 'color'>>,
     colorBy = defaultProps.colorBy,
     borderColor = defaultProps.borderColor as InheritedColorConfig<ComputedDatum<RawDatum>>,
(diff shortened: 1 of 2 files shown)


plouc:
Thank you @benjamintd!

PR #1830, opened 2021-11-09
fix(radial-bar): remove unnecessary console log

benjamintd:
Sorry for the tiny pull request, but I stumbled into this log in a package source, which is probably an unwanted development artifact.


plouc:
@benjamintd, don't be sorry, on the contrary, having a trailing `console.log` in a production package is far from ideal 😅 Thank you!

PR #1836, opened 2021-11-14
fix(core): export linearGradientDef in typings

victoransel1:
Export missing linearGradientDef function type.


plouc:
Thank you @victoransel1!

PR #1849, opened 2021-12-11
chore: update @react-spring/web to 9.3.1

keenahn:
In `@react-spring/web` 9.2.6, a bug occurs on the `postinstall` script and the build fails due to weird characters. This means that `yarn install` fails for yarn 3.

9.3.1 fixes this: https://github.com/pmndrs/react-spring/releases/tag/v9.3.1

Since that version is pinned in `nivo`, no `nivo` packages could be installed using yarn 3. This simple change fixes that.

### DIFF
--- a/packages/annotations/package.json
+++ b/packages/annotations/package.json
@@ -29,7 +29,7 @@
   ],
   "dependencies": {
     "@nivo/colors": "0.74.0",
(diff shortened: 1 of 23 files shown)


plouc:
@keenahn, thank you!

PR #1930, opened 2022-02-18
fix: remove required for labelYOffset in DotsItem

ycmjason:
Currently this is causing error when `labelYOffset` is not provided. But since `labelYOffset` has a default value of `-12`, we should remove `required`.


plouc:
Thank you @ycmjason!

PR #1957, opened 2022-03-23
RadialBar: Utilize defs/fill, modeled after Bar implementation

bdefore:
I wanted to use the patterned fill seen in https://nivo.rocks/marimekko/ for my use of ResponsiveRadialBar but noticed that `defs` and `fill` props were not expected in the RadialBarSvgProps type nor the implementation to pass them along to SvgWrapper in RadialBar the way that Bar does. This PR adds that implementation in a similar way.

Intermediate commits show I was initially pursuing this as Bar does by having `bindKeys` do a targetKey of `data.fill` (https://github.com/plouc/nivo/blob/master/packages/bar/src/Bar.tsx#L268) but it felt more appropriate to instead avoid polluting incoming data and instead target `fill` at the root ComputedBar. I'd be fine with either approach if one is preferred.

Note that the test suite fails on packages/scatterplot/tests/ScatterPlot.test.tsx but I believe these failures predate this proposal.

image

### DIFF
--- a/packages/arcs/src/ArcShape.tsx
+++ b/packages/arcs/src/ArcShape.tsx
@@ -58,7 +58,8 @@ export const ArcShape = <Datum extends DatumWithArcAndColor>({
         <animated.path
             d={style.path}
             opacity={style.opacity}
-            fill={datum.fill || style.fill || style.color}
+            fill={datum.fill || style.color}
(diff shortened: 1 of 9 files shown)


plouc:
@bdefore, thank you for this addition, it makes sense to do it the way you did, and it works with the `arcs` package!

PR #1963, opened 2022-03-30
fix(line): make copy of points data before reversing

benurb:
Hello 👋 

this is a fix for basically a very similar problem to #1644. If `enablePoints` is set to `true` the order of the points in any custom layer changes between the first and any subsequent render, because the points array is modified in place.

Best,
Ben


plouc:
Good catch, thank you for the fix!

PR #1980, opened 2022-04-27
Allow BarCanvas axis to be nullable

squidsoup:
Types do not currently allow `axisBottom`, `axisLeft`, `axisRight`, or `axisTop` to be set to `null`, however this is required to disable axis and is found in the nivo documentation.


plouc (approved):
LGTM

plouc:
Right, good catch and thank you for the fix!

PR #1982, opened 2022-04-29
fix makefile rule for linting an individual package

tkonopka:
The makefile has a rule for linting an individual package, e.g. `make package-lint-calendar`.

The current rule does not find the intended files in the package, and prints an error.

```
No files matching the pattern "./packages/calendar/{src,tests}" were found.
Please check for typing mistakes in the pattern.
```

The pull request adjusts the rule to lint all files in `src` and `tests` directories.


plouc:
Thank you @tkonopka!

Issue #1898, opened 2022-01-20
Radar Chart Rotation


paluchs:
I have a Radar chart with 8 axes. I would like to change the orientation (rotation) of the axes such that I don't have one pointing north, but rather one point northnortheast and one northnorthwest. As far as I know, there is no propertiy to change this.

It would be super elegant if there was an axisRotation prop to define a custom angle from where to start drawing the first axis.
E.g.: axisRotation={22.5} to achieve what I am trying to do above.

Alternatively one could also define the angles between all the axis as a list starting from 0°. In the case explained above: [22.5, 45, 45, 45, 45, 45, 45, 22.5]
This:
<img width="168" alt="Bildschirmfoto 2022-01-20 um 10 59 44" src="https://user-images.githubusercontent.com/34346600/150316558-0b65f149-61f4-4f8d-a17a-6fc60308923b.png">

To this:
<img width="157" alt="Bildschirmfoto 2022-01-20 um 11 02 19" src="https://user-images.githubusercontent.com/34346600/150317004-eef0b0bf-0e8f-40e0-a585-dc8603c29f5f.png">

plouc:
This is a good idea, would be nice to have this option! I'm quite busy atm though 😅 But if you want to/can implement this, that would be great!

tkonopka:
I can have a look and submit a PR

tkonopka:
Please see draft PR #1985.

It implements a prop called `angle`. Is that an acceptable name? It's shorter than 'axisRotation'. An alternative might be `startAngle` because that is already implemented in the Pie component.

Also, a similar feature would be relevant to other circle-based components, e.g. Chord. It would be nice to have a unified api. Would you like that in the same PR, or separately?

plouc:
@tkonopka, sorry I reviewed the PR first, `startAngle` for the pie chart is a bit different as you also have an end, maybe simply `rotation`?

I agree this could be interesting for other charts, but I think `start/endAngle` would be good for _non closed_ charts while `rotation` could be used for _closed_ ones, chord could allow to define a start and end angle in fact.

PR #1985, opened 2022-04-30
feat(radar): orientation of radar chart

tkonopka:
Addresses #1898

Provides a prop `angle` that sets the global orientation for the radar chart.

Example of a default radar chart (left, `angle=0`) and a rotated chart (right, `angle=36`):

2022-04-30-radar-angle

### DIFF
--- a/packages/line/src/Points.js
+++ b/packages/line/src/Points.js
@@ -18,19 +18,22 @@ const Points = ({ points, symbol, size, borderWidth, enableLabel, label, labelYO
      * We reverse the `points` array so that points from the lower lines in stacked lines
      * graph are drawn on top. See https://github.com/plouc/nivo/issues/1051.
      */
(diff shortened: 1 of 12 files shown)


plouc:
I think we could name the property `rotation` instead what do you think? Also it would be nice to ad a test for this new feature if possible.

tkonopka:
Thanks for the feedback @plouc. Just added some changes:

- renamed the new prop from 'angle' to 'rotation'
- added a simple unit test for the rotation feature
- prettified some code in the 'line' package (unrelated to this PR, but required for the CI)

plouc (approved):
LGTM

plouc:
Thank you @tkonopka! I created [a task](https://github.com/plouc/nivo/issues/1995) to add it to the interactive demo.

PR #2008, opened 2022-05-17
Bump all React 17 deps to 18

benrhere:
This is a pretty simple PR which bumps all react 17 dependencies to 18. I'm not sure about other steps that are required for achieving React 18 support, or sure about testing these changes, but hopefully this is a step in the right direction.


package.json
```diff
@@ -83,8 +83,8 @@
   "resolutions": {
     "@types/react": "^17.0.11",
     "@types/eslint": "7.29.0",
-    "react": "17.0.2",
-    "react-dom": "17.0.2"
+    "react": "18.0.0",
```
benrhere, line 86: Should this have a ^ in front of it? (ditto next line)

website/package.json
```diff
@@ -48,8 +48,8 @@
     "gatsby-source-filesystem": "^4.4.0",
     "gatsby-transformer-sharp": "^4.4.0",
     "prism-react-renderer": "^1.2.1",
-    "react": "17.0.2",
-    "react-dom": "17.0.2",
+    "react": "18.0.0",
```
benrhere, line 51: Same as above: Does this need ^ ?

benrhere:
It appears there may be some more work to do here. When running npm update, I get:
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR!
npm ERR! While resolving: nivo@undefined
npm ERR! Found: react@18.1.0
npm ERR! node_modules/react
npm ERR!   react@"^18.0.0" from the root project
npm ERR!
npm ERR! Could not resolve dependency:
npm ERR! peer react@"^16.8.0 || ^17.0.0" from @storybook/addons@6.4.22
npm ERR! node_modules/@storybook/addons
npm ERR!   @storybook/addons@"^6.4.9" from the root project
npm ERR!   peer @storybook/addons@"^6.4.0" from @storybook/addon-knobs@6.4.0
npm ERR!   node_modules/@storybook/addon-knobs
npm ERR!     @storybook/addon-knobs@"^6.4.0" from the root project
npm ERR!
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.

There seems to be a problem with @storybook/addon-knobs, which doesn't support react 18. But that package actually appears to have been deprecated: https://github.com/storybookjs/storybook/discussions/15060

plouc:
@benrhere, yes, I'm afraid there's more work to do before being able to switch to React 18, also migrating everything to `react-spring` and removing `recompose` for good.

tkonopka:
Among the blocking factors, the storybook examples will need updating as @benrhere suggests to avoid deprectated addons. The work is voluminous because most nivo packages use them, but it could be done.

How about creating an issue listing the affected storybook files, and inviting contributions? Volunteers could update and test individual packages (I can help too). And then this PR could at least update the storybook dependencies. @plouc what do you think?

zzzbra:
Is this still the primary PR for getting the React 18 deps into Nivo? Did anyone create an issue listing the affected Storybook examples? I would be happy to help with this.

package.json
shehi: Definitely please. This shouldn't be a question, but rather a statement.

shehi:
This ticket needs more love, as we have entered 2023 now. Happy New Year btw!

hunt3r:
Being pinned to react 17 is causing some issues for my project as mantine 5.x and remix 1.11.x are moving onto 18 and some plugins for Mantine 5.x I need are on 18. Nivo is a backbone of my solution as well, I'd love to have this on 18 too. Is anyone actively pursuing this at this point?

plouc:
It is now possible to use React 18, and it's now the default for the project (see https://github.com/plouc/nivo/pull/2313).

Issue #1962, opened 2022-03-28
Custom scatterplot legend ignoring data value


MasterOdin:
**Describe/explain the bug**

When passing in a custom multi-line legend to the `ScatterPlotCanvas`, the legend data is not being used properly, causing all ids to appear in each row, causing an overflow. I am using `@nivo/scatterplot@0.79.1`.

**To Reproduce**

Using the following legend data:
```
const legends = [
  {
    anchor: "bottom",
    data: [
      {
        color: "#3182EB",
        id: "population_1",
        label: "population_1"
      },
      {
        color: "#3182EB",
        id: "population_2",
        label: "population_2"
      },
      {
        color: "#3182EB",
        id: "population_3",
        label: "population_3"
      },
      {
        color: "#3182EB",
        id: "population_4",
        label: "population_4"
      },
      {
        color: "#3182EB",
        id: "population_5",
        label: "population_5"
      },
      {
        color: "#3182EB",
        id: "population_6",
        label: "population_6"
      },
      {
        color: "#3182EB",
        id: "population_7",
        label: "population_7"
      },
      {
        color: "#3182EB",
        id: "population_8",
        label: "population_8"
      }
    ],
    direction: "row",
    itemDirection: "right-to-left",
    itemHeight: 21,
    itemOpacity: 1,
    itemTextColor: "#000",
    itemWidth: 131,
    justify: false,
    symbolSize: 16,
    translateX: 0,
    translateY: 82
  },
  {
    anchor: "bottom",
    data: [
      {
        color: "#3182EB",
        id: "population_9",
        label: "population_9"
      },
      {
        color: "#3182EB",
        id: "population_10",
        label: "population_10"
      },
      {
        color: "#3182EB",
        id: "population_11",
        label: "population_11"
      },
      {
        color: "#3182EB",
        id: "population_12",
        label: "population_12"
      },
      {
        color: "#3182EB",
        id: "population_13",
        label: "population_13"
      }
    ],
    direction: "row",
    itemDirection: "right-to-left",
    itemHeight: 21,
    itemOpacity: 1,
    itemTextColor: "#000",
    itemWidth: 131,
    justify: false,
    symbolSize: 16,
    translateX: 0,
    translateY: 103
  }
];
```

Getting the following result in my app:
[...]

PR #2016, opened 2022-05-26
fix(scatterplot) support multiple legends

tkonopka:
Addresses #1962

Also provides custom legend labels and toggling visibility of series via the legend

## Background

Issue #1962 describes a scatterplot with multiple series. An attempt to create a legend that spans two lines doesn't work. This is because the chart ignores the manually-specified items under the `legends` prop.

## Change summary

This PR adds support for multi-line legends to @nivo/scatterplot. The fix applies to svg and canvas components (although legends in canvas lack some features, which is an existing limitation)

While editing the legend-handling code, I also added some new features. 

 - support for custom legend labels (new prop `legendLabel`) (svg and canvas)
 - support for toggling series by clicking on legend labels, similar to feature available in @nivo/bar (new prop `initialHiddenIds`) (svg only)

The custom labels and toggling are extras. These can be rolled back or saved for later if you prefer.

Other changes include:

 - some miscellaneous edits to fix linting warnings (e.g. missing dependencies in hooks)
 - unit tests for multi-line legends
 - updated storybook examples for multi-line legends and visibility toggling
 - some re-arrangement in calculations to reduce memory use and avoid redundant calculations when visibility changes (will also be relevant for supporting opacity)

## Example

The screenshot below is based on the dataset in the scatterplot storybook. The dataset has six series. Note the legend uses two lines, the labels are different than the series ids (not shown), two of the series are inactive/invisible. Manual legends can also contains unrelated items if needed.

2022-05-26-multiline-legend

### DIFF
--- a/packages/scatterplot/src/ScatterPlot.tsx
+++ b/packages/scatterplot/src/ScatterPlot.tsx
@@ -130,7 +130,6 @@ const InnerScatterPlot = <RawDatum extends ScatterPlotDatum>({
             />
         )
     }
-    //console.log("nodes: "+JSON.stringify(nodes))
 
     if (layers.includes('nodes')) {
         layerById.nodes = (
--- a/packages/scatterplot/src/ScatterPlot.tsx
+++ b/packages/scatterplot/src/ScatterPlot.tsx
@@ -1,13 +1,13 @@
 import { createElement, Fragment, ReactNode, useMemo } from 'react'
 import { SvgWrapper, Container, useDimensions, CartesianMarkers } from '@nivo/core'
 import { Axes, Grid } from '@nivo/axes'
-import { BoxLegendSvg } from '@nivo/legends'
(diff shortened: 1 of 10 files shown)


schiersj:
Any ideas when this will go to release?

ncolboc:
Interesting PR!
Do you know when it will be merged?

yairEO:
I am a against including any legend logic to Nivo. Nivo should be focused on high-level abstraction for data visualization and legend is highly specific to the design and should be implemented regardless of Nivo. For example in the App I work on, we sometimes use vertical legend, and sometimes radio legend, and sometimes toggable legend items (to show/hide specific ones on the chart).

Legend will add more needless complexity to Nivo itself for no good reason because some very basic default ability and anyway it is not easy to control it's position relative to the chart using CSS queries and any other customization.. many reasons not to include legend in Nivo.

Issue #942, opened 2020-05-04
Sunburst doesn't handle small values well


Deliaz:
It seems that **sunburst** doesn't handle well very small numbers. 
It looks like some coordinate is overfloated and it shows a small-valued beam as a fully filled layer, which covers the others beams.
Please check the data structure in my code example below. 

**Screenshots**
<img src="https://user-images.githubusercontent.com/4865430/80960360-5d5a2b00-8e11-11ea-990f-e2040630bc66.png" width="400">

Code example: https://codesandbox.io/s/nivo-sunburst-f7i02?file=/src/data.json:8237-8284

 - OS: MacOS
 - Browser Chrome 81
 - Version @nivo/sunburst": "^0.61.0

Deliaz:
The issue still happens on the latest version `0.63.1`
https://codesandbox.io/s/nivo-sunburst-forked-4hu4o?file=/src/data.json

wyze:
Indeed it does, thanks @Deliaz and sorry for the noise.

PR #2022, opened 2022-05-27
fix(arcs) skip generation of very small arcs

tkonopka:
Addresses #942

## Background

Issue #942 reports unexpected behavior in the sunburst chart. When the data contains an element that should produce an extremely small arc, that arc can instead "flip around" and dominate the chart.

A similar effect can occur with the pie chart.

```js
const pieData = [
  {
    id: "A",
    value: 30
  },
  {
    id: "B",
    value: 20
  },
  {
    id: "C",
    value: 1e-9  // very small value compared to values in 'A' and 'B'
  }
];
```
Visualizing the above data in a donut chart produces:

Slice C should be tiny or invisible, but instead the yellow color (third color in the nivo color scheme) dominates the visualization. 

https://codesandbox.io/s/stoic-cray-nfyf1n?file=/src/App.js

## Fix

The cause of this effect lies in `d3-shape`. The bug only affects arcs with rounded corners. I raised an [issue](https://github.com/d3/d3-shape/issues/199) to let them know. 

The proposed fix for nivo involves setting a nonzero angle threshold in the `@nivo/arcs` package. The arc renderer can now skip rendering arcs that would be too small to see. Fixing the arcs package automatically fixes issues in `@nivo/pie` and `@nivo/sunburst`.

The new threshold (now set to 1e-7 radians) is set via a prop `skipAngle` in `ArcsLayer`. The name matches a similar prop in `ArcLabelsLayer` and `ArcLinkLabelsLayer`. However, this prop is now 'hidden' inside the arcs package - there are no changes to the API for the Pie and Sunburst components. The justification for 'hiding' this threshold is that it should not have any impact on expected behavior of the chart - tiny slices are expected to be invisible, and this threshold ensures this is the case despite the bug in d3-shape.

The PR also includes some unit tests for the pie and sunburst packages.

### DIFF
--- a/packages/arcs/src/ArcsLayer.tsx
+++ b/packages/arcs/src/ArcsLayer.tsx
@@ -5,6 +5,7 @@ import { DatumWithArcAndColor, ArcGenerator } from './types'
 import { useArcsTransition } from './useArcsTransition'
 import { ArcTransitionMode } from './arcTransitionMode'
 import { ArcMouseHandler, ArcShape, ArcShapeProps } from './ArcShape'
+import { useFilteredDataBySkipAngle } from './utils'
 
 export type ArcComponent<Datum extends DatumWithArcAndColor> = (
     props: ArcShapeProps<Datum>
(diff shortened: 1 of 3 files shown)


tkonopka:
The underlying bug has been fixed in d3-shape v3.2, so this patch is no longer needed.

Issue #2033, opened 2022-06-06
Tooltip positioning on a scaled <div style={{ transform: 'scale(2)' }} />


andresgutgon:
# What?
When a chart is inside a div that is scaled tooltip positioning is not correct.
``` javascript
<div style={{ transform: 'scale(2)' }}>
  <MyFancyNivoChart />
</div>
```

Look at the end how the cursors doesn't correspond with the `Mesh` that I put in debug mode 👀 

https://user-images.githubusercontent.com/49499/172188802-c9752fc4-635c-4cc5-96b6-4a6e89a1821b.mp4

### Describe the solution you'd like
I made a PR with it https://github.com/plouc/nivo/pull/2034

### Describe alternatives you've considered
The thing is that event position is not calculated ok inside an scaled div. The alternative is to find the scaling factor to calculate position

### Additional context
This is the [canvas lib I'm using](https://reactflow.dev/docs/examples/overview/) as you can see the nodes can be zoom-in/zoom-out

## Thanks
Thanks a lot for Nivo. It's awesome and super well documented. 🙌

andresgutgon:
Closing because the PR was merged 🎉 Thank @plouc

plouc:
@andresgutgon, no problem, there's no real release cycle, it depends pretty much on the time I can find to work on the project. I've released this fix in `0.80.0`, but I missed something, it fixes `getRelativeCursor` for SVG based implementation but [all canvas implementations are broken](https://github.com/plouc/nivo/issues/2122) due to this, as it uses `getBBox`, which is only available for SVG elements, I'm trying to fix that.

andresgutgon:
Shit : / Sorry for that. We don't use any canvas chart and I missed that

andresgutgon:
Oh! I just saw you fixed it https://github.com/plouc/nivo/pull/2125 🙌

PR #2034, opened 2022-06-07
Fix tooltip positioning when charts are rendered inside an scaled div

andresgutgon:
# What?
In a normal situation mouse enter / mouse leave events capture the position ok. But when the chart is inside a scaled
element with a CSS transform like: `transform: scale(2);` tooltip are not positioned ok. Comparing original width `box.width` agains scaled width give us the scaling factor to calculate ok mouse position.

## The issue
This PR [fix this issue](https://github.com/plouc/nivo/issues/2033)

## The problem
The problem as explained in the issue that when the charts are inside an scaled div like this:
```javascript
<div style={{ transform: 'scale(2)' }}>
  <MyFancyNivoChart />
</div>
```
The tooltip will be misplaced as you can see in the video in the issue I linked

### The solution
I found [the solution here](https://github.com/mapbox/mapbox-gl-js/pull/10096)

### DIFF
--- a/packages/core/src/lib/interactivity/index.js
+++ b/packages/core/src/lib/interactivity/index.js
@@ -12,6 +12,14 @@ export * from './detect'
 export const getRelativeCursor = (el, event) => {
     const { clientX, clientY } = event
     const bounds = el.getBoundingClientRect()
+    const box = el.getBBox()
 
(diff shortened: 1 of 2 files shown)


geclos (approved):

plouc (approved):
LGTM

PR #2076, opened 2022-07-13
Align d3-shape version on bar with other packages

pauldijou:
I've been facing hoisting issues while using nivo on a monorepo app with Lerna / Yarn workspaces, leading to runtime errors due to an incorrect React context, probably because the `@nivo/core` module was duplicated at the root of the app and inside each project.

Wondering if that could be because `@nivo/bar` is not using the same minor version for `d3-shape` as other packages.


plouc:
Good catch @pauldijou!

Issue #1401, opened 2021-01-27
Allow onMouseMove and onMouseLeave Events For Line Charts when Slice Tooltip is Enabled


pnewhook:
**Is your feature request related to a problem? Please describe.**
I'm synchronizing multiple charts on the same page using the technique described here: https://nivo.rocks/storybook/?path=/story/scatterplot--synchronizing-charts However, adding custom onMouseMove and onMouseLeave handles precludes using the out of the box slice tooltip. Each chart has two series, one for current values and one for target values. It would be nice to see both values in the tooltip, as it behaves when `enableSlices = 'x'`

This scenario is similar requested in this comment, https://github.com/plouc/nivo/issues/47#issuecomment-656187650, but I felt it deserved it's own feature request given the number of upvotes that comment has.

**Describe the solution you'd like**
I would like to be able to supply `onMouseMove` and `onMouseLeave` handlers when `enableSlices="x"`

**Describe alternatives you've considered**
I'm exploring adding a custom tooltip but it is messier than the default  x slice tooltip.

**Additional context**

MatissJanis:
bump

Porqumn:
bump

fspin:
bump

p-abasian:
bump

PR #2128, opened 2022-09-18
Add mouse events to line slices

sigmike:
On the Line chart the mouse events (`onMouseEnter`, `onMouseMove`, `onMouseLeave` and `onClick`) work fine when the mesh is enabled but not when the slices are enabled. This PR fixes that.

Related issue: #1401.

### DIFF
--- a/packages/line/src/Line.js
+++ b/packages/line/src/Line.js
@@ -220,6 +220,10 @@ const Line = props => {
                 tooltip={sliceTooltip}
                 current={currentSlice}
                 setCurrent={setCurrentSlice}
+                onMouseEnter={onMouseEnter}
+                onMouseMove={onMouseMove}
+                onMouseLeave={onMouseLeave}
+                onClick={onClick}
             />
         )
     }
--- a/packages/line/src/Slices.js
+++ b/packages/line/src/Slices.js
@@ -10,7 +10,19 @@ import { memo } from 'react'
 import PropTypes from 'prop-types'
 import SlicesItem from './SlicesItem'
 
-const Slices = ({ slices, axis, debug, height, tooltip, current, setCurrent }) => {
+const Slices = ({
+    slices,
+    axis,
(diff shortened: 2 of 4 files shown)


azhard4int:
@plouc Can we get this merged, we need mouse event feature for line slices at Usermaven.com

Stormtv:
This would be very helpful

Stormtv:
@plouc & @wyze any chance to get this merged?

yo-johnnyd-hc:
Oh yes please! I need this as well.

zikowang:
Nice work! We need this feature as well.

fspin:
I think this is also related: https://github.com/plouc/nivo/pull/2187
It will be great indeed if one of these PR can be merged! 🙏

un-neil:
@plouc & @wyze please can we get this merged?

plouc (approved):
LGTM, however I think this should be added to the docs, but this can be done in a separate PR, as adding doc is currently undocumented 😅.

cbix:
@sigmike @plouc I noticed this passes the slice to the handler while normally they'd receive a point, was this on purpose? It's particularly confusing with TypeScript. While I'm aware the line chart still lacks full TS support I see two possible improvements here:
1. change the event handler so it optionally receives a slice or a point, something like
```ts
export type PointMouseHandler = (point: Point, slice: null, event: React.MouseEvent) => void
export type SliceMouseHandler = (point: null, slice: Slice, event: React.MouseEvent) => void
export interface LineProps {
  // ...
  onMouseEnter?: PointMouseHandler | SliceMouseHandler
  // ...
}
```
2. add extra event handlers for slices (`onSliceMouseEnter` etc.)

damianr13:
Is there any fix in version v.0.80.0? I am unable to upgrade due to the following issue: https://github.com/plouc/nivo/issues/2310

PR #2139, opened 2022-09-28
Minor docs fix for specifying color

Josh-a-e:


plouc:
Good catch, thank you!

PR #2204, opened 2022-12-19
Node 18 support

piwysocki:


ivanrosvimi:
Up!

MiguelPraw:
UP

api/package.json
```diff
@@ -16,7 +16,7 @@
   },
   "engineStrict": true,
   "engines": {
-    "node": "16.x"
```
dubzzz, line 19: Why not: >=16.x

Changing to 18.x would make users of node 16 not being able to import the library anymore. >=16.x seems to be a better compromise.

plouc:
Node 18 is now supported (and used), fixed in https://github.com/plouc/nivo/pull/2313.

PR #2222, opened 2023-01-23
Check if ResizeObserver is defined instead of checking window object

andriuss:
**ResizeObserver** is not implemented in all environments, for example, [jsdom](https://github.com/jsdom/jsdom/issues/3368).
This forces nivo library consumers to mock or add polyfils, for example [jest.setup.js](https://github.com/plouc/nivo/blob/1867b606111ad1253cadb7e3cbcab2663c3275c5/packages/jest.setup.js).

Checking for global ResizeObserver removes the need to have it on the global window object.

### DIFF
--- a/packages/core/src/hooks/useMeasure.js
+++ b/packages/core/src/hooks/useMeasure.js
@@ -11,10 +11,9 @@ export const useMeasure = () => {
     })
 
     const [observer] = useState(() => {
-        // Check if window is defined (so if in the browser or in node.js).
-        const isBrowser = typeof window !== 'undefined'
-        if (!isBrowser) return null
-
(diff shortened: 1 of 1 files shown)


plouc (approved):
LGTM

plouc:
@andriuss, I think you forgot to run prettier.

andriuss:
> @andriuss, I think you forgot to run prettier.

Thanks! I did run **packages-lint-fix** now.

andriuss:
The build looks good now?

joaopedromatias:
Hey @andriuss, even with this PR we are still going to need to mock the `Resize Observer` on `jsdom` env right?

joaopedromatias:
:shipit:

PR #2239, opened 2023-02-09
Allow custom choice of ticksPosition

deanrobertcook:
The `ticksPosition` value is ignored in the `Axes` class. This commit first checks to see if one is set on the axis properties and prefers that, or else defaults to the old behavior.


deanrobertcook:
bump

PR #2240, opened 2023-02-10
bugfix: added string to theme.fontSize to prevent TypeScript warning

GrantMLambert:
Currently using strings to set the font size will raise a TypeScript warning.


plouc:
Font Size should be a number as nivo assumes it is a number in various places (SVG, canvas...).

PR #2280, opened 2023-03-29
fix: Relax constraints on `@react-spring/web`

dubzzz:
Having a constraint on `@react-spring/web` blocking it to a very precise version (here: `9.4.5`) can push users of nivo into pulling it several times in their node_modules or bundles if they don't really check for duplicated packages. Indeed it's rather simple to have two distinct libraries requesting it at version `9.x` at the same time, and if one of them (not nivo) request something not being `9.4.5` they we would pull the library twice.

It seems that given `@react-spring/web` seems to follow semver, there is not real risk to say nivo supports `^9.4.5` instead of `9.4.5` only.

### DIFF
--- a/packages/annotations/package.json
+++ b/packages/annotations/package.json
@@ -1,45 +1,45 @@
 {
-  "name": "@nivo/annotations",
-  "version": "0.80.0",
-  "license": "MIT",
-  "author": {
-    "name": "Raphaël Benitte",
-    "url": "https://github.com/plouc"
-  },
-  "repository": {
-    "type": "git",
-    "url": "https://github.com/plouc/nivo.git",
-    "directory": "packages/annotations"
-  },
-  "keywords": [
-    "nivo",
-    "dataviz",
-    "react",
-    "charts",
-    "annotations"
-  ],
-  "main": "./dist/nivo-annotations.cjs.js",
-  "module": "./dist/nivo-annotations.es.js",
-  "typings": "./dist/types/index.d.ts",
-  "files": [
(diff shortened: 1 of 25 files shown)


dubzzz:
Seems that I have some pieces of the code to adapt if I want this Pr to make it and at least compile:

```txt
packages/treemap/src/defaults.ts:69:14 - error TS4023: Exported variable 'htmlDefaultProps' has or is using name 'AnimationConfig' from external module "/tmp/9a4cf066/node_modules/@react-spring/core/dist/index" but cannot be named.

69 export const htmlDefaultProps = {
```

dubzzz:
This PR is waiting for a status on https://github.com/pmndrs/react-spring/pull/2128. If the other PR gets merged then nivo could probably start declaring this import with a `^`. In the meantime, maybe I should change from `9.4.5` to the version that will have the patch in it for TS users 🤔

packages/annotations/package.json
```diff
@@ -29,7 +29,7 @@
   ],
   "dependencies": {
     "@nivo/colors": "0.80.0",
-    "@react-spring/web": "9.4.5",
+    "@react-spring/web": "9.4.5 || ^9.7.2",
```
dubzzz, line 32: The versions >9.4.5 and <9.7.2 are not valid from a TS point of view, thus I dropped them from the range of acceptable versions

dubzzz:
@plouc Any opinion?

plouc (approved):
LGTM

PR #2282, opened 2023-04-05
Bugfix/website/bar add legends fix

Nanak360:
Clicking the Add Legend button on the bar page crashes the entire website as shown bellow
Tab-1680499869883-website

Bellow is the error message that we see when clicking the Add Legend button on dev env.
Tab-1680499836483-with error

Idealy a new legend group should be added. This PR has the fix for that.
Tab-1680499765293-proper

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 it's built on top of d3.
 
 Several libraries already exist for React d3 integration,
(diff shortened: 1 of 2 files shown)


Nanak360:
#2288 refers this issue

PR #2299, opened 2023-04-24
Add firstWeekday attr to calendar/TimeRange

yzalvov:
So that Monday could be first.

Screenshot 2023-04-24 at 18 59 25

### DIFF
--- a/packages/calendar/src/compute/timeRange.ts
+++ b/packages/calendar/src/compute/timeRange.ts
@@ -10,7 +10,7 @@ import {
     timeSunday,
 } from 'd3-time'
 import { timeFormat } from 'd3-time-format'
-import { DateOrString } from '../types'
+import { DateOrString, Weekday } from '../types'
 import { isDate } from 'lodash'
 
 // Interfaces
@@ -48,7 +48,7 @@ interface ComputeCellPositions
     }[]
     colorScale: (value: number) => string
     emptyColor: string
-    firstDayOfWeek: number
+    firstWeekday: Weekday
 }
 
 interface ComputeWeekdays
@@ -57,7 +57,7 @@ interface ComputeWeekdays
         ComputeBaseDimensionProps {
     ticks?: number[]
     arrayOfWeekdays?: string[]
-    firstDayOfWeek: number
+    firstWeekday: Weekday
 }
 
 interface Day {
@@ -147,32 +147,24 @@ export const computeCellSize = ({
     }
 }
 
-export const getDayIndex = (date: Date, firstDayOfWeek = 0) => {
+export const getDayIndex = (date: Date, firstWeekday: Weekday) => {
     const days = [0, 1, 2, 3, 4, 5, 6]
     const day = date.getDay()
-    const offsetDay = day - firstDayOfWeek
+    const offsetDay = day - firstWeekday
     const [dayIndex] = days.slice(offsetDay)
     return dayIndex
 }
 
-const getTimeInterval = (firstDayOfWeek = 0) => {
-    switch (firstDayOfWeek) {
-        case 0:
-        default:
-            return timeSunday
-        case 1:
-            return timeMonday
-        case 2:
-            return timeTuesday
-        case 3:
-            return timeWednesday
-        case 4:
-            return timeThursday
-        case 5:
-            return timeFriday
-        case 6:
-            return timeSaturday
-    }
+const getTimeInterval = (firstWeekday: Weekday) => {
+    return [
+        timeSunday,
+        timeMonday,
+        timeTuesday,
+        timeWednesday,
+        timeThursday,
+        timeFriday,
+        timeSaturday,
+    ][firstWeekday]
 }
 
 function shiftArray<T>(arr: T[], x: number): T[] {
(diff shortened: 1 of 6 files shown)


packages/calendar/src/types.ts
```diff
@@ -212,6 +212,7 @@ export type TimeRangeSvgProps = Dimensions & { data: CalendarDatum[] } & Partial
                 role: string
                 weekdayLegendOffset: number
                 weekdayTicks: Array<0 | 1 | 2 | 3 | 4 | 5 | 6>
+                firstDayOfWeek: number
```
plouc, line 215: I think it should be a union as we don't support any number.

packages/calendar/src/compute/timeRange.ts
```diff
@@ -244,15 +300,11 @@ export const computeWeekdays = ({
...
-        'Tuesday',
-        'Wednesday',
-        'Thursday',
-        'Friday',
-        'Saturday',
-    ],
+    firstDayOfWeek,
+    arrayOfWeekdays = shiftArray(
+        ['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'],
+        firstDayOfWeek || 0
```
plouc, line 306: Why the `|| 0` as we have a default?

packages/calendar/src/compute/timeRange.ts
```diff
@@ -135,26 +147,68 @@ export const computeCellSize = ({
...
 function computeGrid({
     startDate,
     date,
     direction,
+    firstDayOfWeek,
 }: {
     startDate: Date
     date: Date
     direction: 'horizontal' | 'vertical'
+    firstDayOfWeek: number
```
plouc, line 198: I think we should restrict with an union.

packages/calendar/src/types.ts
plouc: Maybe it would be even better to use `'monday' | 'tuesday' | ...`, we don't really know if `0` stands for Monday or Sunday, I feel it would provide a clearer API.

packages/calendar/src/compute/timeRange.ts
```diff
@@ -135,26 +147,68 @@ export const computeCellSize = ({
...
 
+export const getDayIndex = (date: Date, firstDayOfWeek = 0) => {
+    const days = [0, 1, 2, 3, 4, 5, 6]
+    const day = date.getDay()
+    const offsetDay = day - firstDayOfWeek
+    const [dayIndex] = days.slice(offsetDay)
+    return dayIndex
+}
+
+const getTimeInterval = (firstDayOfWeek = 0) => {
```
plouc, line 158: Rather than a switch, we could use a map IMHO.

packages/calendar/src/props.ts
```diff
@@ -52,4 +52,5 @@ export const timeRangeDefaultProps = {
     dayRadius: 0,
     square: true,
     weekdayLegendOffset: 75,
+    firstDayOfWeek: 0,
```
plouc, line 55: It could be nice to add a story for this or, ideally, to add it to the interactive doc.

plouc (changes requested):

packages/calendar/src/types.ts
yzalvov: Updated with enum.

packages/calendar/src/props.ts
yzalvov: Added to the interactive docs.

packages/calendar/src/types.ts
```diff
@@ -192,6 +192,16 @@ export type TimeRangeTooltipProps = Omit<TimeRangeDayData, 'date' | 'value'> & {
     value: string
 }
 
+export enum Weekday {
```
plouc, line 195: Unions are simpler to use IMHO, and for non typescript users, we have the same issue as before, it'll be just a number.

plouc (changes requested):

yzalvov:
Hi @plouc, is there anything else you'd want me to add?

plouc (approved):
LGTM

plouc:
Hi @yzalvov, nope, LGTM, will merge as soon as tests pass, thank you!

yzalvov:
Hi @plouc, thanks for your prompt reaction and thank you for the library!

FYI:

1. The tooltip seems to be broken after your recent updates. Neither default nor custom tooltip components are rendered for me on current master.
2. No updates of the interactive docs regarding this new `firstWeekday` feature is found available https://nivo.rocks/time-range

plouc:
I moved to vercel, but didn't update the DNS yet, you can see it here for now: https://nivo-nxi8sxerj-plouc.vercel.app/time-range/. Regarding the tooltip issue, I think it's an old issue due to a cyclic dependency (`@nivo/core -> @nivo/tooltip -> @nivo/core`), I need to fix this among other things.

plouc:
DNS updated, it might take some time to propagate.

yzalvov:
Hi @plouc, thanks so much for the library by the way 🤓 Helps a lot.. Just checking, is the broken tooltip situation resolved? If not, could you share a link to the issue please?

PR #2337, opened 2023-05-11
Move theming to its own package

plouc:
`@nivo/core` contains a bit of everything, and it can lead to circular dependencies (it's the case with `@nivo/tooltip` for example), so I decided to move it to its own package.

The added benefit is that now this part will be in TypeScript, as `@nivo/core` is unfortunately still not.

Extra features:

- We can now define a text outline for most text elements via `outlineWidth`, `outlineColor` and `outlineOpacity` via the theme object.

This introduces some breaking changes:

- When implementing custom layers/components, you might use `useTheme`, previously imported from `@nivo/core`, now it's defined in `@nivo/theming`
- The TypeScript definition of a theme changed a little bit, please see the new type in the `@nivo/theming` package.


pcorpet:
@plouc Any update on this one? It would be useful to get the d3-color vulnerability fix released.

RicardoAALL:
Also looking for a progress update on the d3-color vulnerability fix. Noticed in #2142 this has been an ongoing issue since at least September 2022. 11 months is quite a long time for a high-severity vulnerability to be live. I understand open-source projects such as this are usually passion projects but is there anything in particular, we might be able to help with via contribution/PR? I've tried to look into other PR's and related issues but perhaps I have missed something.

PR #2345, opened 2023-05-24
Asking for help

msafrun:
Hi Raphaël Benitte, please help me on how to implement vertical markers in BAR CHART.

My case is that I have data in one month from 1-30 in x axis then I would like to mark public holiday and weekend with markers. I can achieve it in LINE CHART, but in BAR CHART the markers only can be shown horizontally.

Thanks,
Haekal


plouc:
The code isn't up-to-date, and sorry but I don't really understand why using a PR.

PR #2357, opened 2023-06-05
Feature: Add option for positioning tooltip for LineCanvas

bdknox:
The tooltip defaults to the existing behavior to be at the cursor, but optionally can be placed at the point that the data in the tooltip is representing (similar to the default for SVG Lines)

https://github.com/plouc/nivo/assets/25183171/936cad49-41d4-446e-b68d-1bad13e5b5b4


barrymichaeldoyle:
bump

PR #2369, opened 2023-06-20
Fix arcLinkLabelsOffset for pie chart

aautem:
Noticed the `arcLinkLabelsOffset` control wasn't doing anything on the [pie chart demo](https://nivo.rocks/pie/).

Tracked it down to a couple of issues:

1. In the default props, the `arcLinkLabels` layer came before the `arcs` layer so arcs were being rendered over the top of the arc link labels.
2. The `offset` was not being included in the dependency array for `useTransitionPhases` so changing the `arcLinkLabelsOffset` required another prop to change before the offset would update in the UI.

### Before
Screenshot 2023-06-20 at 3 14 41 PM

### After
Screenshot 2023-06-20 at 3 13 22 PM

### DIFF
--- a/packages/pie/src/Pie.tsx
+++ b/packages/pie/src/Pie.tsx
@@ -122,30 +122,10 @@ const InnerPie = <RawDatum extends MayHaveLabel>({
     const boundDefs = bindDefs(defs, dataWithArc, fill)
 
     const layerById: Record<PieLayerId, ReactNode> = {
-        arcLinkLabels: null,
         arcs: null,
+        arcLinkLabels: null,
         arcLabels: null,
         legends: null,
-    }
-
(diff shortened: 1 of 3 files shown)


packages/pie/src/Pie.tsx
```diff
@@ -169,6 +149,26 @@ const InnerPie = <RawDatum extends MayHaveLabel>({
         )
     }
 
+    if (enableArcLinkLabels && layers.includes('arcLinkLabels')) {
```
aautem, line 152: Just reordering these sections so they line up with the order of the default `layers`. Looked like that was the intention here.

plouc:
Good catch @aautem, thank you for the fix!

Issue #2346, opened 2023-05-25
Website component search does not work


Denisiums:
**Describe/explain the bug**
Component search on website does not allow to type anything in (it crashes with "Uncaught TypeError: n is not a function
    a ComponentsExplorer.tsx:34
    r ComponentsSearch.tsx:11")

**To Reproduce**
1. Go to 'https://nivo.rocks/components/'
2. Click on 'a search bar'
3. Type something

**Expected behavior**
Search works

**Desktop (please complete the following information):**
 - Macos
 - Firefox, Chrome

PR #2371, opened 2023-06-20
Fix website component search

aautem:
Resolves https://github.com/plouc/nivo/issues/2346

Found that the `navigate` function being passed from `Components` to `ComponentsExplorer` was `undefined`. Importing `navigate` from `gatsby` in the `ComponentsExplorer` instead.


plouc:
Thank you for the fix @aautem!

Issue #2391, opened 2023-07-20
Stream chart is overflowing, particularly in "silhouette" mode


baldulin:
**Bug:** The stream charts take the lowest y coordinate to be 0. This does not work in a lot of modes. But this is best seen with the "silhouette" mode.

**To reproduce:** Just go to https://nivo.rocks/stream/ the default settings, pretty much show the issue already.

Edit: Forgot to say, great library, thx for all your work and effort everyone.

PR #2392, opened 2023-07-20
fix(stream): fix stream diagram overflowing

baldulin:
The stream diagram was overflowing. Particularly in the "silhouette" mode. This was introduced in d53681b439d3bac08e978491a56518b3fe9c98b5 where the d3 scale was replaced with the nivo equivalent. While ignoring that the min value might be negative thus must be put into the first parameter.

This closes: #2391


plouc:
Thank you for the fix @baldulin!

PR #2396, opened 2023-07-21
feat/export responsive bar props TypeScript types

ollwenjones:
People can use them if it's not just an inline Omit type.


plouc (approved):
LGTM

criveratro:
I've an issue were is being typed as:

```
ResponsiveBar: <RawDatum extends BarDatum>(props: Omit<any, "height" | "width">) => JSX.Element
```

Note the `any`.

But if I open the actual file with the type definition it seems correct:

```
ResponsiveBar: <RawDatum extends BarDatum>(props: Omit<BarSvgProps<RawDatum>, "height" | "width">) => JSX.Element;
```

Started happening after updating `@nivo/core` to `v0.83.1`, could be related to this change? Previously I was using `v0.80.0`.

Issue #2419, opened 2023-08-25
Bug in TimeRange related to firstWeekday


wirhabenzeit:
**Description**
The `firstWeekday` setting in `ResponsiveTimeRange` can result in the display of wrong week days. 

**To Reproduce**
1. On [TimeRange documentation](https://nivo.rocks/time-range/) the first day *April 2, 2018* is correctly displayed as a *Monday*. 
2. Switch the `firstWeekday` setting to `monday`: *April 2, 2018* is still correctly displayed as a *Monday*. 
3. Switch the `firstWeekday` setting back to `sunday`: Now *April 2, 2018* is wrongly displayed as a *Tuesday*.

**Expected behavior**
The week days should reflect the actual calendar irrespective of the `firstWeekday` setting.

**Screenshots**
Initially April 2 is displayed as a Monday. 
<img width="842" alt="Correct initial display" src="https://github.com/plouc/nivo/assets/17571643/32c15d14-46bf-455b-8ea0-aff541581cb4">
After switching the `firstWeekday` to `monday` it is still correct
<img width="879" alt="Still correct after switching to Monday" src="https://github.com/plouc/nivo/assets/17571643/0c2c016f-c5d7-47f1-8e98-1f007e131fe4">
After switching back to `sunday` April 2, 2018 is wrongly displayed as a Tuesday. 
<img width="836" alt="Wrong after switching back to Sunday" src="https://github.com/plouc/nivo/assets/17571643/32b19cf4-18ae-459e-8a7a-559b267b3627">

**Desktop (please complete the following information):**
 - OS: MacOS 13.5.1
 - Browser: Safari 
 - Version: 16.6

**Additional context**
- Timezone: Europe/Zurich

wirhabenzeit:
**Update**

I think I found the problem. The function 

https://github.com/plouc/nivo/blob/2671ca4df614190f597bb63030bb794d3106c36e/packages/calendar/src/compute/timeRange.ts#L184-L193

accidentally (?) mutates 

https://github.com/plouc/nivo/blob/2671ca4df614190f597bb63030bb794d3106c36e/packages/calendar/src/compute/timeRange.ts#L150-L158

when called in 

https://github.com/plouc/nivo/blob/2671ca4df614190f597bb63030bb794d3106c36e/packages/calendar/src/compute/timeRange.ts#L310

**Fix**

It seems `shiftArray` should not mutate its argument, i.e. for instance 

```ts
[...]

PR #2420, opened 2023-08-25
shiftArray in calendar/src/compute/timeRange.ts should not mutate its argument

wirhabenzeit:
This is a proposed fix for https://github.com/plouc/nivo/issues/2419

The function `shiftArray` mutates `ARRAY_OF_WEEKDAYS` which results in a wrong week day display in certain situations.

### DIFF
--- a/packages/calendar/src/compute/timeRange.ts
+++ b/packages/calendar/src/compute/timeRange.ts
@@ -184,12 +184,8 @@ const getTimeInterval = (firstWeekday: Weekday) => {
 function shiftArray<T>(arr: T[], x: number): T[] {
     if (!arr.length || !x) return arr
 
-    for (let i = 0; i < x; i++) {
-        const shifted = arr.shift() as T
-        arr.push(shifted)
-    }
-
-    return arr
+    x = x % arr.length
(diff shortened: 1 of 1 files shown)


packages/calendar/src/compute/timeRange.ts
```diff
@@ -184,12 +184,8 @@ const getTimeInterval = (firstWeekday: Weekday) => {
...
     if (!arr.length || !x) return arr
 
-    for (let i = 0; i < x; i++) {
-        const shifted = arr.shift() as T
-        arr.push(shifted)
-    }
-
-    return arr
+    x = x % arr.length
+    return arr.slice(x, arr.length).concat(arr.slice(0, x))
```
ordruk, line 188: Alternatively, you can create a new array and build it based on the shifted elements from the original array.

```ts
function shiftArray<T>(arr: T[], x: number): T[] {
    if (!arr.length || !x) return arr;

    const newArr = [...arr];

    for (let i = 0; i < x; i++) {
        const shifted = newArr.shift() as T;
        newArr.push(shifted);
    }

    return newArr;
}
```

ordruk: However, I liked your fix 🙏

Issue #2415, opened 2023-08-17
Warning: Support for defaultProps will be removed from function components


donfour:
**Describe/explain the bug**
Getting the following error when using Nivo with Next.js 13 inside `app` router:
```
app-index.js:31 Warning: Jr: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.
```

It doesn't affect the chart rendering though. It's just a bit annoying to see a long error in the console. Other libraries are also seeing a similar issue:
- https://github.com/recharts/recharts/issues/3615
- https://github.com/chakra-ui/chakra-ui/issues/7057

Wondering if Nivo has a plan to move from `defaultProps` to JS default parameters? It seems like quite a large refactor. Happy to contribute if needed. Thanks!

**To Reproduce**
Codesandbox: https://codesandbox.io/p/sandbox/frosty-tom-fgddxp

1. Go to the codesandbox
2. Open the preview in a new tab
3. Open dev tool console

**Expected behavior**
Rendering a chart should throw no error.

RobertAron:
I'm running into this issue as well. The warning is quite long also, which makes it difficult to parse other logs that appear in the inspector.

I understand this could be a large refactor... If anyone has suggestions for a hack to suppress warnings about this issue temporarily I would really appreciate it ❤️

bernatfortet:
Any updates on this? 
Has anyone found a way to hide the warnings?

andre19980:
I tried to do the refactor (based in the similar issue that @donfour linked to the description) but the deploy has failed and I guess I don't have access to vercel to see the logs :(
It's my first contribution, so I'm not really sure if I did all correctly

Sinakhx:
still looking forward for the fix

dreampulse:
I fixed that by patching the `console.log`, since the warning is polluting the developer experience.
(And for me, the issue is just occurring in external libraries)

```js
if (process.env.NODE_ENV !== "production") {
  // eslint-disable-next-line no-console
  const originalWarn = console.error;
  // eslint-disable-next-line no-console
  console.error = (...args) => {
    if (
      args[0].includes(
        "Support for defaultProps will be removed from function components in a future major release.",
      )
    ) {
      return;
    }
    originalWarn(...args);
  };
}
```

ClydSpyd:
Hello!

Any update on this issue?

andre19980:
[...]

Issue #2077, opened 2022-07-13
Warning: Failed prop type: The prop `labelYOffset` is marked as required in `Ti`, but its value is `undefined`.


asifkabani:
![image](https://user-images.githubusercontent.com/9083341/178789815-8781fe81-72d5-41ee-ae6f-6db5aaa4c56c.png)

This is my code:

```
<Line
            {...commonProperties}
            data={[
              {
                id: "positive",
                data: [
                  { x: 0, y: 0 },
                  { x: 1, y: 0.3 },
                  { x: 2, y: 0.5 },
                  { x: 3, y: 0 },
                ],
              },
              {
                id: "negative",
                data: [
                  { x: 3, y: 0 },
                  { x: 4, y: -0.5 },
                  { x: 5, y: -0.2 },
                ],
              },
            ]}
            curve="monotoneX"
            pointSymbol={CustomSymbol}
            pointSize={9}
            pointBorderWidth={1}
            pointBorderColor={{
              from: "color",
              modifiers: [["darker", 0.3]],
            }}
            colors={["#0BCEBB", "#FFD572"]}
            yScale={{
              type: "linear",
              stacked: false,
              min: -50,
              max: 50,
            }}
            enableGridX={false}
            enableArea={true}
            areaOpacity={0.2}
            useMesh={true}
            crosshairType="cross"
            axisBottom={null}
          />
```

erikasv:
if it's of any use, I found this warning is solved when the prop `pointLabelYOffset` is passed ( I wasn't passing it because it's optional and I have point labels disabled)

Saxtone:
I also experienced the same problem. It was because I had a custom `markers` object and the error was thrown because of it.

Adding these props fixed the warning:
```
markers={[
          {
            axis: 'x',
            value: 0,
            lineStyle: { stroke: colors.background, strokeWidth: 1 },
            legend: '',
            // @ts-expect-error: Adding these props to fix Nivo 0.88.0 bug with legendOffsetX being required
            legendOffsetX: 0,
            legendOffsetY: 0,
            legendOrientation: 'horizontal',
          },
        ]}
```

The last 3 about the `legend` were needed in order to remove the error from appearing.

plouc:
`prop-types` were removed via https://github.com/plouc/nivo/pull/2714

PR #2452, opened 2023-10-20
refactor default props to fix error message

andre19980:
### This PR is related to issue #2415.
Refactor default props declared like:
```
Component.defaultProps = {
    prop1: 0,
    prop2: 1,
    // ....
}
```
Now, the default props are defined directly inside the component
```
export const Component = props => {
    const {
      prop1 = 0,
      prop2 = 1,
      // ...
    } = props
    // ...
}
```

The given solution is based in the issue from another repository:
- chakra-ui/chakra-ui#7057

### DIFF
--- a/packages/core/src/components/cartesian/markers/CartesianMarkersItem.js
+++ b/packages/core/src/components/cartesian/markers/CartesianMarkersItem.js
@@ -171,10 +171,10 @@ const CartesianMarkersItem = ({
     lineStyle,
     textStyle,
     legend,
-    legendPosition,
-    legendOffsetX,
-    legendOffsetY,
-    legendOrientation,
+    legendPosition = 'top-right',
+    legendOffsetX = 14,
+    legendOffsetY = 14,
+    legendOrientation = 'horizontal',
 }) => {
     const theme = useTheme()
 
@@ -255,12 +255,6 @@ CartesianMarkersItem.propTypes = {
     legendOffsetX: PropTypes.number.isRequired,
     legendOffsetY: PropTypes.number.isRequired,
     legendOrientation: PropTypes.oneOf(['horizontal', 'vertical']).isRequired,
-}
-CartesianMarkersItem.defaultProps = {
-    legendPosition: 'top-right',
-    legendOffsetX: 14,
-    legendOffsetY: 14,
-    legendOrientation: 'horizontal',
 }
 
 export default memo(CartesianMarkersItem)
--- a/packages/core/src/components/defs/patterns/PatternDots.js
+++ b/packages/core/src/components/defs/patterns/PatternDots.js
@@ -1,7 +1,20 @@
 import { memo } from 'react'
 import PropTypes from 'prop-types'
 
-export const PatternDots = memo(({ id, background, color, size, padding, stagger }) => {
+export const PatternDotsDefaultProps = {
+    color: '#000000',
+    background: '#ffffff',
+    size: 4,
+    padding: 4,
+    stagger: false,
+}
+
+export const PatternDots = memo(props => {
+    const { id, background, color, size, padding, stagger } = {
+        ...PatternDotsDefaultProps,
+        ...props,
+    }
+
     let fullSize = size + padding
     const radius = size / 2
     const halfPadding = padding / 2
@@ -33,14 +46,6 @@ PatternDots.propTypes = {
     size: PropTypes.number.isRequired,
     padding: PropTypes.number.isRequired,
     stagger: PropTypes.bool.isRequired,
-}
-
-PatternDots.defaultProps = {
-    color: '#000000',
-    background: '#ffffff',
-    size: 4,
-    padding: 4,
-    stagger: false,
 }
 
 export const patternDotsDef = (id, options = {}) => ({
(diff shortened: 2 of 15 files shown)


plouc:
@andre19980, could you please rebase this PR?

andre19980:
@plouc it worked! Thanks :)

plouc:
@andre19980, it seems like the tooltip doesn't work anymore for the choropleth (SVG implementation), see https://nivo.rocks/choropleth/ VS https://nivo-git-fork-andre19980-refactor-default-props-of-b2d03b-plouc.vercel.app/choropleth/.

plouc:
A test is also failing for the line chart.

fernandojpps:
I hope you guys don't mind, but I saw this PR and decided to take a look.

From my investigation it looks like the problem is when we explicitly pass `undefined` as a prop.

The failing tests from the line chart explicitly define `axisLeft` and `axisBottom` as `undefined`. If we comment those lines all tests pass but this actually reveals a bigger issue.

When we do something like `{...defaultProps, ...props}` we are overriding all props that are defined in `props`, even if they are defined as `undefined` 👀.
As an example of this behaviour, you can check that `{ ...{ a: 1 }, ...{ a: undefined} }` evaluates to `{a: undefined}`.

One solution for this problem could be to clear props undefined before using them, like:
```js
function getOnlyDefinedProps(props) {
  return Object.fromEntries(
    Object.entries(props).filter(([_, v]) => v !== undefined)
  )
}

// We could also memoize the result, so that the function only runs when props change
function useDefinedProps(props) {
  return useMemo(() => getOnlyDefinedProps(props), [props])
}

function Component(props) {
  const definedProps = useDefinedProps(props)
  const readyToUseProps = {...defaultProps, ...definedProps}
 //...
}
```

I hope this helps!

@andre19980 @plouc Let me know if you want me to help with this PR.

andre19980:
Awesome, @fernandojpps! Thank you for the help!
Your solution seems to fix the failing test, but I wonder if filtering the received props is actually the right behavior, 'cause maybe the person using the component wants to send an undefined prop. In this case, the component shouldn't discard it, especially if it's gonna affect the component's behavior, like prevent something from being rendered. Then I guess we would be creating a bug.

fernandojpps:
You're welcome @andre19980!

I understand your point. I actually find it weird that `undefined` can have two different meanings.

From my understanding, the current behavior with `defaultProps` is to use default values whenever a `prop` is `undefined`. It doesn't matter if it was explicitly set as `undefined` or not.

I created a simple React Component to test how it actually behaves, and here are the results.

```jsx
const UndefinedTest = ({name = "John"}) => <p>Hello {name}</p>

<UndefinedTest name={undefined} /> //<p>Hello John</p>
<UndefinedTest />                  //<p>Hello John</p>
<UndefinedTest name="Joe" />.      //<p>Hello Joe</p>
```

I believe my suggestion helps keep the behavior aligned with the default React behavior, and in turn with developer expectations.

plouc:
@andre19980, @fernandojpps,

Thank you for your help with this issue, another option would be to explicitly assign a default value for each property needing one, that's what I did for other components, I found it clear enough, even if it's a bit more tedious to maintain, it's easy to see when you have a default or not. That being said, I'm not against the suggested approach as long as it leads to the same default behavior.

andre19980:
@fernandojpps you're totally right, my bad 🥸 . Hope you don't mind, but I followed @plouc's approach, just cause I also thought it would be more explicit and clearer

nehalist:
Any chance this is going to be merged?

Sinakhx:
I'm still getting the error

nehalist:
> I'm still getting the error

Since there hasn't been made a new release since this was merged, it's not that surprising. @plouc any plans on releasing a new version in the near future? Thanks a lot!

plouc:
@nehalist, I'll try to create a new release next week.

nehalist:
Thanks a lot! :+1:

b310L:
When will the new version be released?

GreenAsJade:
This PR seems to be the likely cause of #2539 .   Any chance of someone "here" taking a look?

andre19980:
hey @GreenAsJade thanks for reporting the issue! So I took a look into the code to see what may be causing the problem and found out that the `Line` component has its props defined with `isRequired` in the `PropTypes` declaration. One can check in the file `packages/line/src/props.js`

Captura de Tela 2024-03-19 às 12 20 35

I was able to fix the warnings by removing the requirement conditions, but I'm not sure if I'm allowed to delete these protections. @plouc I saw in the documentation that many of those props are `optional`, so I'm a bit confused if they should be marked with `isRequired`. To solve the warnings, I was thinking of editing the components where I did the refactor in the default props and remove the `isRequired` condition from the props which are written as _optional_ in the documentation. Would that be ok and safe?

Btw, I think this problem is also causing the bug #2077, so I think it would solve both problems

plouc:
Properties were required because we had defaults, due to how we now define the default values, it won't pass the prop-types validation, I think we should remove the prop-types for the package, as it was done for others, relying on TS instead.
I removed the prop-types for most packages when I migrated them to TS, `@nivo/line` hasn't been migrated yet, but we could still remove them.

csandman:
> Properties were required because we had defaults, due to how we now define the default values, it won't pass the prop-types validation, I think we should remove the prop-types for the package, as it was done for others, relying on TS instead. I removed the prop-types for most packages when I migrated them to TS, `@nivo/line` hasn't been migrated yet, but we could still remove them.

That would probably be the ideal approach. Most people rely on TS for prop type validation nowadays anyway, and the errors are pretty annoying.

Issue #2297, opened 2023-04-23
Bar Chart - renderToStaticMarkup not rendering properly


mrkirchner:
**Describe/explain the bug**
When rendering a barchart with renderToStaticMarkup the bars will not render.  At one point there was this bug but was closed https://github.com/plouc/nivo/issues/1778.

Even a pr was added but never merged due to package changes.
https://github.com/plouc/nivo/pull/1712

**To Reproduce**
https://codesandbox.io/s/bar-staticmarkup-x6vzv?file=/src/App.js

1. Render a Bar chart using renderToStaticMarkup to see no bars.

**Expected behavior**
Bars render properly

**Screenshots**
<img width="362" alt="Screenshot 2023-04-23 at 3 13 13 PM" src="https://user-images.githubusercontent.com/13805753/233866619-76f2266f-bc2e-4923-938a-f5802c5b05ff.png">

**Desktop (please complete the following information):**
 - OS: OSX
 - Browser Chrome/Firefox/Safari
 - Version 80

mrkirchner:
@plouc does Nivo have anyone fixing issues?  This issue also seems related https://github.com/plouc/nivo/issues/2368

mrkirchner:
Bump?

mrkirchner:
@plouc or @shoito any word on this ? It prevents using bar on server side.  Original pr was one liner but package was moved.

OleksandrRakovets:
Sorry it took so long 😮‍💨 

Been using a custom build with this change in production for over a year w/o any issues.

Hope it'll get merged this time 🤞

plouc:
This issue should have been solved by https://github.com/plouc/nivo/pull/2464.

@OleksandrRakovets, thank you!

Issue #2310, opened 2023-04-27
v0.81.0 can lead to import issues


plouc:
## Related to:
- https://github.com/plouc/nivo/issues/2307
- https://github.com/plouc/nivo/issues/2309
- https://github.com/plouc/nivo/issues/2305

## What's the issue?

Users of nivo are experiencing import issues since the release `0.81.0`, this is due to a breaking change in `d3-color`, which was updated due to security issues.

[Most D3 packages are now esm only](https://devclass.com/2021/06/15/d3-7-0-goes-all-in-on-ecmascript-modules/) and `d3-color` is one of them, other libraries depending on D3 experienced the same problem, victory even [wrote a post about this](https://formidable.com/blog/2022/victory-esm/) explaining it in details, and [visx has an opened issue about it](https://github.com/airbnb/visx/issues/1637).

Nivo provides esm/cjs/umd, but only esm can be supported now due to this change, and it also seems that sometimes people get the cjs version rather than the esm one, that's another thing that needs to be fixed.

## What should I do?

- You could stick to `0.80.0`, as the D3 packages used in this version are not esm only, the drawback is that there are some critical security issues in those. Also `0.80.0` contains a [regression for canvas based components](https://github.com/plouc/nivo/issues/2162) 😓 
- Try the workarounds suggested in the issues mentioned above.
- Help to find a proper fix :) Share workarounds, even if those are temporary fixes, it could help other users to upgrade.

Aeolun:
It's a bit bizarre, but maybe it's possible to change the import for d3-scale/color to the victory-vendor package that transpiles the d3 dependencies to CJS if necessary?

leeuwd:
Nivo `1.83.0`: 

```
Error: require() of ES Module [redacted]/node_modules/@nivo/colors/node_modules/d3-color/src/index.js from  [redacted]/node_modules/@nivo/colors/dist/nivo-colors.cjs.js not supported.
Instead change the require of index.js in  [redacted]/node_modules/@nivo/colors/dist/nivo-colors.cjs.js to a dynamic import() which is available in all CommonJS modules.
```

APIWT:
I'm a little confused -- even after updating to 0.81.0, `npm audit` shows the high vunerability in `d3-color`:

```
npm audit
# npm audit report

d3-color  <3.1.0
Severity: high
d3-color vulnerable to ReDoS - https://github.com/advisories/GHSA-36jr-mh4h-2g58
fix available via `npm audit fix --force`
[...]

PR #2464, opened 2023-11-16
Fix Bar SSR

### LINKED ISSUE #1778: [Bar] ReactDOMServer.renderToStaticMarkup doesn't render data after nivo/Bar version 0.73.0
I need to convert to png multiple charts for my project, and for that i use ReactDOMServer.renderToStaticMarkup .
The issue is since nivo/bar version 0.73.0 the data are not displayed on first render even with the animate props set to false.

Here's a link to a Sandbox demo :
https://codesandbox.io/s/bar-staticmarkup-x6vzv?file=/src/App.js

If you change nivo/bar version to 0.72.0 or less the data is displayed.

Normal display and dangerouslySetInnerHTML of the staticMarkup :
image

OleksandrRakovets:
Resolves #1778, #2297

Was using a custom package with this change for over a year with no issues.

### DIFF
--- a/packages/bar/src/Bar.tsx
+++ b/packages/bar/src/Bar.tsx
@@ -231,6 +231,7 @@ const InnerBar = <RawDatum extends BarDatum>({
         }),
         config: springConfig,
         immediate: !animate,
+        initial: animate ? undefined : null,
     })
 
     const commonProps = useMemo(


petrvecera:
Oh I incorrectly assumed that this will fix the SSR on NextJS -  https://github.com/plouc/nivo/issues/2310  but that's different issues. Back to the dynamic rendering! :D 

_Btw I hate the packages (d3) which provide only esm from the newer versions. We are stuck on some projects with old versions because we can't allocate weeks of work for fixes and rewrites on projects with millions lines of code just because some dude decides that everyone should migrate esm ..._

Issue #2503, opened 2024-01-29
Error when expanding Axis options on website


scalabw:
**Describe/explain the bug**
An error appearing on the website when trying to expand any axis options "axisTop" "axisBottom" "axisLeft" "axisRight".
It make the website unusable and you are also not able to see the options for those axis.

**To Reproduce**
You can go to any graph on the website (such as this one: [Line](https://nivo.rocks/line/))
Then click on the option "axisTop" this will turn all the website blank and show an error in the console (as you can see in the screenshot section)

**Screenshots**
<img width="1718" alt="image" src="https://github.com/plouc/nivo/assets/22681512/92a83ac5-9752-493c-9e8f-2bf718ce4af8">
<img width="1715" alt="image" src="https://github.com/plouc/nivo/assets/22681512/dd8b6a13-d72c-4b3b-8172-712c40307ad0">

msolano00:
I can confirm this.

msolano00:
I can confirm this. @plouc

scalabw:
I created a pull request that solve this issue #2504

PR #2504, opened 2024-01-29
Fix: add initial property for truncateTickAt

scalabw:
This PR will  the error appearing on the website when trying to expand any axis options "axisTop" "axisBottom" "axisLeft" "axisRight"

It solve the issue [#2503](https://github.com/plouc/nivo/issues/2503)

### DIFF
--- a/website/src/data/components/line/defaults.ts
+++ b/website/src/data/components/line/defaults.ts
@@ -34,6 +34,7 @@ export default {
         tickRotation: 0,
         legend: '',
         legendOffset: 36,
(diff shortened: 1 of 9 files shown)


plouc:
Technically, the default value should be `undefined`, but I guess it's fine in the context of the demo website.

scalabw:
> Technically, the default value should be `undefined`, but I guess it's fine in the context of the demo website.

I tried using undefined as default value but it throw an error:
image

Issue #2520, opened 2024-02-27
Theming website - expanding tooltip crashes the web page


LaurelineP:
## Context: nivo website theming
https://nivo.rocks/guides/theming/
### Issue: 
When reaching out for the tooltip customisation, the web page crashes
### Scenario:
- go to the [theming page of the  website](https://nivo.rocks/guides/theming/)
- scroll down to the bottom to reach the `tooltip` section
- expand the tile `tooltip object optional`
- will reveal another expandable section `tooltip.container`
- click on it
     _➡️ this displays a blank page_
     _➡️ this logs an error in the console_
    
![image](https://github.com/plouc/nivo/assets/32878345/2df0f8de-4771-44c0-bd37-5a6c23ca04c0)

 ### Personal Setup
 - OS: Mac
 - Browser: Chrome
 - Version: 121.0.6167.184
=

scalabw:
I opened a PR to solve that here: https://github.com/plouc/nivo/pull/2521

plouc:
Fixed by https://github.com/plouc/nivo/pull/2521 thanks to @scalabw.

PR #2521, opened 2024-02-27
fix tooltip default color

scalabw:
This PR will the error appearing on the website when trying to expand tooltip.container in "/guides/theming/"

It should solve the issue #2520 

image

### DIFF
--- a/website/src/components/guides/theming/defaults.ts
+++ b/website/src/components/guides/theming/defaults.ts
@@ -110,7 +110,7 @@ const annotationsDefaults: CompleteTheme['annotations'] = {
 const tooltipDefaults: CompleteTheme['tooltip'] = {
     container: {
         background: '#ffffff',
-        color: "#000000",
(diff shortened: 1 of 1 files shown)


website/src/components/guides/theming/defaults.ts
```diff
@@ -110,7 +110,7 @@ const annotationsDefaults: CompleteTheme['annotations'] = {
 const tooltipDefaults: CompleteTheme['tooltip'] = {
     container: {
         background: '#ffffff',
-        color: extendedTheme.textColor,
```
plouc, line 113: I think it should be `extendedTheme.text.fill`.

plouc (changes requested):

website/src/components/guides/theming/defaults.ts
scalabw: Done!

PR #2524, opened 2024-03-02
Touch crosshair for line graphs

WilliamABradley:
This adds the ability to shift the crosshair on mobile by touching and holding along the graph surface just like Google's Stock and Exchange rate graphs, which you can run your finger along on mobile.

This option is only available in mesh mode, as it appears that slices would have to be refactored to figure out when touchMove moves across different elements.

Also, it seems that in order to build, I needed to add:
```
    "@babel/preset-env": "^7.21.5",
    "@babel/preset-react": "^7.18.6",
    "@babel/preset-typescript": "^7.21.5",
```

Not sure if the package lock is wonk, or if the issue is me being on Windows.


packages/line/src/Mesh.js
```diff
@@ -68,6 +72,41 @@ const Mesh = ({
         [onClick]
     )
 
+    const handleTouchStart = useCallback(
```
plouc, line 75: `showTooltipAt`, `tooltip`, `margin` and `setCurrent` dependencies are missing (same for `handleTouchMove`).

packages/tooltip/src/context.ts
```diff
@@ -7,7 +7,11 @@ export interface TooltipActionsContextData {
         position: [number, number],
         anchor?: TooltipAnchor
     ) => void
-    showTooltipFromEvent: (content: JSX.Element, event: MouseEvent, anchor?: TooltipAnchor) => void
+    showTooltipFromEvent: (
```
plouc, line 10: Seems like `MouseEvent | TouchEvent` would me more accurate, what do you think?

packages/voronoi/src/Mesh.tsx
```diff
@@ -102,13 +125,55 @@ export const Mesh = <Datum,>({
...
+            }
+            if (onTouchEnd) {
+                let previousNode: Datum | undefined = undefined
+                if (currentIndex !== null) {
+                    previousNode = nodes[currentIndex]
+                }
+                previousNode && onTouchEnd(previousNode, event)
+            }
+        },
+        [setCurrentIndex, currentIndex, onTouchEnd, nodes]
```
plouc, line 176: `touchCrosshair` dependency is missing.

packages/voronoi/src/Mesh.tsx
```diff
@@ -102,13 +125,55 @@ export const Mesh = <Datum,>({
...
+                setCurrentIndex(index)
+            }
+            if (node) {
+                onTouchStart?.(node, event)
+            }
+        },
+        [getIndexAndNodeFromTouchEvent, setCurrentIndex, onTouchStart]
+    )
+
+    const handleTouchMove = useCallback(
```
plouc, line 150: `touchCrosshair` dependency is missing.

packages/voronoi/src/Mesh.tsx
```diff
@@ -102,13 +125,55 @@ export const Mesh = <Datum,>({
...
             setCurrentIndex(index)
             if (node) {
                 onClick?.(node, event)
             }
         },
-        [getIndexAndNodeFromEvent, setCurrentIndex, onClick]
+        [getIndexAndNodeFromMouseEvent, setCurrentIndex, onClick]
+    )
+
+    const handleTouchStart = useCallback(
```
plouc, line 137: `touchCrosshair` dependency is missing.

packages/line/index.d.ts
```diff
@@ -185,6 +186,9 @@ export interface LineProps {
     onMouseMove?: PointMouseHandler
     onMouseLeave?: PointMouseHandler
     onClick?: PointMouseHandler
+    onTouchStart?: PointTouchHandler
```
plouc, line 189: It would be nice to add these new properties to the website, but the way to do so is unfortunately not documented.

plouc: I think we could also add tests for those, if possible.

plouc (changes requested):

packages/line/src/Mesh.js
WilliamABradley: Looks like the were some missing deps on the mouse ones too, patched in https://github.com/WilliamABradley/nivo/commit/f2e5fbb6007e02bbf0a7ea0f4f363916fe7ee138

packages/voronoi/src/Mesh.tsx
WilliamABradley: Updated in https://github.com/WilliamABradley/nivo/commit/f2e5fbb6007e02bbf0a7ea0f4f363916fe7ee138

packages/voronoi/src/Mesh.tsx
WilliamABradley: Updated in https://github.com/WilliamABradley/nivo/commit/f2e5fbb6007e02bbf0a7ea0f4f363916fe7ee138

packages/tooltip/src/context.ts
WilliamABradley: Well, technically it means a logic change, but it can be tidier to support unwrapping the client coords inside the function. 

(MouseEvent has one clientX/Y at the top level, while TouchEvent supports multi-touch, so all of the clientX/Y coords are in an array of touches)

Patched in https://github.com/WilliamABradley/nivo/commit/4ad5f67ffb61af4f8d5cb29b18fd55c0259a6ab3

packages/line/index.d.ts
WilliamABradley: Added tests and docs!

WilliamABradley:
@plouc I've address the feedback 🙂 
Not sure why vercel preview is failing and I can't access any logs. Do you know why?

packages/line/index.d.ts
plouc: Thank you!

plouc (approved):
LGTM, thank you for the adjustments!

WilliamABradley:
I can take a look at the website build tonight 🙂 

Also, I think I need to add support for slices to this feature, as my use case doesn't actually need the voronoi mesh (I think maybe it is worth describing the purpose of slices in the documentation as I didn't understand why it was needed, and what this difference between slices and mesh was).

Correct me if I'm wrong, slices allows for a vertical/horizontal crosshair on one axis, rather than Voronoi which is closest point reguardless of axis? It felt unnatural for my FX graph for the crosshair to be skipping around (Due to different high and low points)

Are you happy for me to add that slice support to this PR or would it make sense to make another?

plouc:
> I can take a look at the website build tonight 🙂

Thank you :)

> Also, I think I need to add support for slices to this feature, as my use case doesn't actually need the voronoi mesh (I think maybe it is worth describing the purpose of slices in the documentation as I didn't understand why it was needed, and what this difference between slices and mesh was).
> 
> Correct me if I'm wrong, slices allows for a vertical/horizontal crosshair tracking on one axis, rather than Voronoi which is closest point reguardless of axis? It felt unnatural for my FX graph for the crosshair to be skipping around (Due to different high and low points), which I only gathered from turning on the separate debug flags.

Slices can be useful when you want to visualize a "slice" of data all at once when hovering a point belonging to that slice, while with the voronoi mesh we only show one data point. The voronoi mesh is also useful for very dense charts, as it can become difficult to hover a specific point, but it could be improved, for example by adding a secondary radius to restrict the detection, so that it's not constantly jumping around.
 
> Are you happy for me to add that slice support to this PR or would it make sense to make another?

This PR is fine.

WilliamABradley:
Gatsby straight up refuses to comply on Windows 🤷  
Luckily I finally got it fixed and running on my work Macbook 😄

I got the touch support for slices working nicely, and I added an explanation to isInteractive about the pros/cons for mesh vs slices

plouc (approved):
LGTM, I'm not a big fan of passing all the slices and accessing the DOM directly to find the current slice, but I don't have anything else to suggest.

WilliamABradley:
Yeah, it is a little gross, but the alternative would be a large refactor to use an interceptor like mesh, but that would be a lot of work and introduce bugs.

Issue #2020, opened 2022-05-27
Totals for stacked bars


ghost:
**Is your feature request related to a problem? Please describe.**
I don't know if there is a way to show the totals of stacked bars as asked in 

**Describe the solution you'd like**
Being able to show the total of the portions of the stacked bar over the bars or as the number of the last segment.

**Additional context**
I opened this issue, because there is a closed similar issue (https://github.com/plouc/nivo/issues/951) where the last message says: "Closing this issue after a prolonged period of inactivity. If this issue is still present in the latest release, please create a new issue with up-to-date information. Thank you!"

I personally need a solution to this in a project I'm working and I suspect other people are looking for one too.

damianocasula:
I mention @Mulli and @timothyarmes who asked for this feature in the other issue (https://github.com/plouc/nivo/issues/951).

Mulli:
@[damianocasula](https://github.com/damianocasula) I am still waiting for a solution... Thanks for re-opening

tkonopka:
Hi @damianocasula, hi @Mulli. Do you mean something like this? 

![2022-05-28-bar-totals](https://user-images.githubusercontent.com/7260190/170834387-752b8ce7-07e8-45c0-bf44-d343444d2b7e.png)

https://codesandbox.io/s/hopeful-mcclintock-44m0w8?file=/src/App.js

This implementation uses a custom layer. Note the sandbox only works for vertical bar plots, but hopefully you can customize it for your needs.

Mulli:
Hi Tomasz
Thanks for taking the time to show the example
I shall give it a try

Best
Mulli Bahr
+972-50-9262025
https://site2goal.co.il


‫בתאריך שבת, 28 במאי 2022 ב-19:47 מאת ‪Tomasz Konopka‬‏ <‪
***@***.***‬‏>:‬

ChuckJonas:
Converted this solution to typescript (and did some very slight refactoring).

```typescript
const BarTotalsLayer: BarLayer<{ value: number }> = ({
  bars,
  xScale,
  yScale,
}) => {
  const labelOffset = 10;
  const labelFontSize = 12;
  if (bars.length === 0) return null;
  // compute totals for each index/bar

  const totals = bars.reduce<Record<number, number>>((acc, bar) => {
    console.log(bar.data.indexValue);
    const indexValue = bar.data.indexValue;
    if (!(indexValue in acc)) {
      acc[indexValue] = 0;
    }
[...]

Issue #951, opened 2020-05-11
Totals for stacked bars


timothyarmes:
It would be really useful to show the total at the top of each bar so that we can show a sales breakdown by category along with total earnings.

alexiastandish:
@timothyarmes did you ever figure out a workaround for this? Currently I'm using a custom bar component but with that it no longer utilizes the built in tooltip functionality

Joelelmiller:
Looking for a work around for this too. Would be a great addition.

jeffodle:
Anyone find a solution to this?

alexiastandish:
@Joelelmiller @jeffodle I did but it's pretty hacky... 

`layers={['grid', 'axes', 'bars', Line, 'markers', 'legends']}`

I added a custom Line component and then I used the bar data that is passed to the custom component through nivo to render a custom bar with a <text> element that shows the total of the bar. Custom <Line/> component below... 

`
const Line = ({
    bars
}) => {
    if (bars.length / 3 <= 14) {
        const barsList = [];
        bars.map((currBar) => {
            const key = currBar.key.slice(currBar.key.indexOf('.') + 1);
            const keyCheck = barsList.find((element) => element.key === key);
            const keyCheckIndex = barsList.findIndex((element) => element.key === key);

            if (keyCheck) {
                barsList[keyCheckIndex]['bars'].push(currBar.data.id);
            } else {
                barsList.push({
                    key: key,
                    bars: [currBar.data.id]
                });
            }
        });

        return ( <
            Fragment > {
                bars.map((bar) => {
                    const barListKey = barsList.findIndex((element) => element.key === bar.key.slice(bar.key.indexOf('.') + 1));
                    const lastBarInGroup = barsList[barListKey].bars[barsList[barListKey].bars.length - 1];

                    if (bar.data.id === lastBarInGroup) {
                        return ( <
                            svg key = {
                                bar.key
                            }
                            style = {
                                {
                                    overflow: 'visible'
                                }
                            } >
                            <
                            g >
                            <
                            rect x = {
                                bar.x
                            }
                            y = {
                                bar.y - 20
                            }
                            width = {
                                bar.width
                            }
                            height = "14px"
                            fill = "transparent" / >
                            <
                            text x = {
                                bar.x + bar.width / 2
                            }
                            y = {
                                bar.y + 25 / 2 - 20
                            }
                            style = {
                                {
                                    fontSize: '12px',
                                    textAlign: 'center'
                                }
                            }
[...]

PR #2525, opened 2024-03-02
feat(Bar): add bar totals

joaopedromatias:
This adds a `totals` layer to the bar chart component.

The goal is to provide the total value for each index of the chart.

In order to activate it, users have to add the `layers` prop having the `totals` value to the BarSvg or BarCanvas component, like following:

```ts
export const WithTotals = {
    return (
        <Bar
            {...props}
            layers={['grid', 'axes', 'bars', 'markers', 'legends', 'annotations', 'totals']}
        />
}
```

The issues #951 and #2020 are related to this feature.

Captura de Tela 2024-03-02 às 19 39 08


packages/bar/src/Bar.tsx
```diff
@@ -362,6 +364,19 @@ const InnerBar = <RawDatum extends BarDatum>({
         )
     }
 
+    if (layers.includes('totals')) {
```
plouc, line 367: Usually, I tend to expose a prop rather than having to deal with layers directly, something like `enableTotals` with the layer enabled by default, I feel it's a bit simpler to use than having to redefine all layers, what do you think?

packages/bar/src/Bar.tsx
```diff
@@ -362,6 +364,19 @@ const InnerBar = <RawDatum extends BarDatum>({
...
+        layerById.totals = (
+            <BarTotals
+                key={'totals'}
```
plouc, line 370: nit: `key="totals"`.

packages/bar/src/BarCanvas.tsx
```diff
@@ -52,6 +60,93 @@ const findBarUnderCursor = <RawDatum,>(
...
+
+    const totalsByIndex = new Map<string | number, number>()
+
+    const barWidth = bars[0].width
+    const barHeight = bars[0].height
+    const yOffsetVertically = -10
+    const xOffsetHorizontally = 20
+    const fontSize = 12
+
+    ctx.fillStyle = '#222222'
```
plouc, line 81: Styles should be defined by the theme IMO.

packages/bar/src/BarCanvas.tsx
```diff
@@ -52,6 +60,93 @@ const findBarUnderCursor = <RawDatum,>(
...
+) {
+    if (bars.length === 0) return
+
+    const totalsByIndex = new Map<string | number, number>()
+
+    const barWidth = bars[0].width
+    const barHeight = bars[0].height
+    const yOffsetVertically = -10
+    const xOffsetHorizontally = 20
+    const fontSize = 12
```
plouc, line 79: The font-size should be defined by the theme IMO.

packages/bar/src/BarCanvas.tsx
```diff
@@ -52,6 +60,93 @@ const findBarUnderCursor = <RawDatum,>(
...
+    layout: BarCommonProps<RawDatum>['layout'] = defaultProps.layout,
+    groupMode: BarCommonProps<RawDatum>['groupMode'] = defaultProps.groupMode
+) {
+    if (bars.length === 0) return
+
+    const totalsByIndex = new Map<string | number, number>()
+
+    const barWidth = bars[0].width
+    const barHeight = bars[0].height
+    const yOffsetVertically = -10
```
plouc, line 77: I think it would be best to make these offsets configurable.

packages/bar/src/BarCanvas.tsx
```diff
@@ -52,6 +60,93 @@ const findBarUnderCursor = <RawDatum,>(
...
+
+            ctx.fillText(
+                String(totalsByIndex.get(indexValue)),
+                xPosition + (layout === 'vertical' ? barWidth / 2 : xOffsetHorizontally),
+                yPosition + (layout === 'vertical' ? yOffsetVertically : barHeight / 2)
+            )
+        })
+    }
+
+    if (groupMode === 'grouped') {
```
plouc, line 115: nit: `else if`?

packages/bar/src/BarCanvas.tsx
```diff
@@ -52,6 +60,93 @@ const findBarUnderCursor = <RawDatum,>(
...
+        const greatestValueByIndex = new Map<string | number, number>()
```
plouc, line 116: I think it could help top briefly explain the purpose of:
- `totalsByIndex`
- `totalsPositivesByIndex`
- `greatestValueByIndex`
- `numberOfBarsByIndex`
In comments.

packages/bar/src/BarTotals.tsx
```diff
@@ -0,0 +1,154 @@
...
+    groupMode = defaultProps.groupMode,
+}: BarTotalsProps<RawDatum>) => {
+    if (bars.length === 0) return <></>
+    const totals: JSX.Element[] = []
+
+    const totalsByIndex = new Map<string | number, number>()
+
+    const barWidth = bars[0].width
+    const barHeight = bars[0].height
+    const yOffsetVertically = -10
```
plouc, line 27: As for the canvas implementation, styles should come from the theme and offsets should be configurable.

packages/bar/src/BarCanvas.tsx
```diff
@@ -32,6 +33,13 @@ import { renderAxesToCanvas, renderGridLinesToCanvas } from '@nivo/axes'
 import { renderLegendToCanvas } from '@nivo/legends'
 import { useTooltip } from '@nivo/tooltip'
 import { useBar } from './hooks'
+import {
+    updateGreatestValueByIndex,
+    updateNumberOfBarsByIndex,
+    updateTotalsByIndex,
+    updateTotalsPositivesByIndex,
+} from './BarTotals'
```
plouc, line 41: Imports could be defined in a separate file, so that we don't have to import an SVG component for canvas.

packages/bar/src/BarTotals.tsx
```diff
@@ -0,0 +1,154 @@
...
+    groupMode?: BarCommonProps<RawDatum>['groupMode']
+}
+
+export const BarTotals = <RawDatum extends BarDatum>({
+    bars,
+    xScale,
+    yScale,
+    layout = defaultProps.layout,
+    groupMode = defaultProps.groupMode,
+}: BarTotalsProps<RawDatum>) => {
```
plouc, line 19: Seeing both implementations (canvas + SVG), it seems that the computation is quite similar, I think we could compute the totals in some kind of hook (`useBarTotals` or something similar), and then we would only care about rendering for each implementation. What do you think?

packages/bar/src/BarTotals.tsx
```diff
@@ -0,0 +1,154 @@
...
+                    {totalsByIndex.get(indexValue)}
+                </text>
+            )
+        })
+    }
+
+    return <>{totals}</>
+}
+
+export const updateTotalsByIndex = (
```
plouc, line 121: If we extract the computation in a hook, we might not need all these helpers.

storybook/stories/bar/Bar.stories.tsx
```diff
@@ -294,6 +294,15 @@ export const WithSymlogScale: Story = {
     ),
 }
 
+export const WithTotals: Story = {
```
plouc, line 297: 👍🏻

website/src/data/components/bar/meta.yml
```diff
@@ -1,76 +1,78 @@
 flavors:
-  - flavor: svg
-    path: /bar/
-  - flavor: canvas
-    path: /bar/canvas/
-  - flavor: api
-    path: /bar/api/
+    - flavor: svg
```
plouc, line 2: If possible, I'd prefer to keep the formatting as is, as it's not nice for git history.

packages/bar/src/Bar.tsx
plouc: It also makes it easier to document, right now, appart from the story, it'll be hard to tell that this feature exists.

plouc (changes requested):
Thank you for this, should be very useful as it's been requested quite often!

website/src/data/components/bar/meta.yml
joaopedromatias: I'm sorry it was prettier

packages/bar/src/BarTotals.tsx
joaopedromatias: Done! Great tip, thanks

packages/bar/src/BarTotals.tsx
joaopedromatias: I created a central hook for svg and canvas but I kept the helpers in order to document with comments and not repeating code.

packages/bar/src/BarTotals.tsx
joaopedromatias: Done! Thanks for the tip

packages/bar/src/Bar.tsx
joaopedromatias: done, this way is much better, thank you!

I made the default behavior as not enabled in order to don't affect production charts. What do you think?

packages/bar/src/BarTotals.tsx
plouc: Makes sense.

packages/bar/src/Bar.tsx
```diff
@@ -420,6 +436,21 @@ const InnerBar = <RawDatum extends BarDatum>({
 
                 return layerById?.[layer] ?? null
             })}
+            {barTotals.map(barTotal => (
```
plouc, line 439: Ah, sorry, I should have been more explicit about using a flag to enable the feature, I think it should come on top of the layers feature (with the layer present by default), layers can be useful not only to control if something is present or not, but also to control the order of the layers, which can be important in certain situations.

packages/bar/src/BarCanvas.tsx
```diff
@@ -367,6 +371,25 @@ const InnerBarCanvas = <RawDatum extends BarDatum>({
...
+            bars,
+            xScale,
+            yScale,
+            layout,
+            groupMode,
+            totalsOffset
+        )
+
+        ctx.fillStyle = theme.text.fill
+        ctx.font = `bold ${theme.text.fontSize}px sans-serif`
```
plouc, line 385: I think you could use the following, which is used for similar components:

```
ctx.font = `${theme.labels.text.fontSize}px ${theme.labels.text.fontFamily}`
```

packages/bar/src/Bar.tsx
joaopedromatias: Got it, so the prop should control the layer array. Just to check if I got that right, should it be enabled by default?

plouc: Those would be independent, for example, this is how it's done for the labels of the `Chord` diagram:

```
if (layers.includes('labels') && enableLabel) {
```

And `labels` is part of the layers by default.

packages/bar/src/BarCanvas.tsx
joaopedromatias: I just implemented this way for svg and canvas

packages/bar/src/Bar.tsx
joaopedromatias: Got it! So it is implemented using layers, but controlled mainly through prop.

packages/bar/src/Bar.tsx
```diff
@@ -368,6 +370,11 @@ const InnerBar = <RawDatum extends BarDatum>({
         )
     }
 
+    if (layers.includes('totals') && enableTotals) {
+        const barTotals = computeBarTotals(bars, xScale, yScale, layout, groupMode, totalsOffset)
```
plouc, line 372: Ideally, this should be memoized, in a hook, there's a bit of logic involved so it might be best to not have to recompute it all the time, as even hovering a bar would lead to recompute this.

packages/bar/src/BarTotals.tsx
```diff
@@ -0,0 +1,29 @@
+import { useTheme } from '@nivo/core'
+import { BarTotalsData } from './compute/totals'
+
+interface Props {
+    barTotals: BarTotalsData[]
+}
+
+export const BarTotals = ({ barTotals }: Props) => {
```
plouc, line 8: Sorry, I didn't think about it when first reviewing the PR, but how does this looks with animated charts? all other components support transitions, so it could be weird if we just have this not animated.

packages/bar/src/props.ts
```diff
@@ -54,7 +54,7 @@ export const defaultProps = {
 
 export const svgDefaultProps = {
     ...defaultProps,
-    layers: ['grid', 'axes', 'bars', 'markers', 'legends', 'annotations'] as BarLayerId[],
+    layers: ['grid', 'axes', 'bars', 'markers', 'legends', 'annotations', 'totals'] as BarLayerId[],
```
plouc, line 57: I think `legends` and `annotations` should come first.

packages/bar/src/props.ts
```diff
@@ -69,7 +69,7 @@ export const svgDefaultProps = {
 
 export const canvasDefaultProps = {
     ...defaultProps,
-    layers: ['grid', 'axes', 'bars', 'legends', 'annotations'] as BarCanvasLayerId[],
+    layers: ['grid', 'axes', 'bars', 'legends', 'annotations', 'totals'] as BarCanvasLayerId[],
```
plouc, line 72: I think `legends` and `annotations` should come first.

plouc (changes requested):

packages/bar/src/BarTotals.tsx
joaopedromatias: I just added it following the implementation structure for the bars itself.

packages/bar/src/Bar.tsx
joaopedromatias: I did it and passed the computation to `useBar` hook

packages/bar/src/Bar.tsx
```diff
@@ -276,13 +283,48 @@ const InnerBar = <RawDatum extends BarDatum>({
         targetKey: 'data.fill',
     })
 
+    const totalsTransition = useTransition<
```
plouc, line 286: As this is optional, I think it should be defined in its own component.

packages/bar/src/Bar.tsx
```diff
@@ -276,13 +283,48 @@ const InnerBar = <RawDatum extends BarDatum>({
...
+        BarTotalsData,
+        {
+            x: number
+            y: number
+            labelOpacity: number
+        }
+    >(barTotals, {
+        keys: barTotal => barTotal.key,
+        from: barTotal => ({
+            x: layout === 'vertical' ? barTotal.x : barTotal.x - 50,
```
plouc, line 296: I don't think we should have hardcoded values.

packages/bar/src/Bar.tsx
```diff
@@ -276,13 +283,48 @@ const InnerBar = <RawDatum extends BarDatum>({
...
+            y: barTotal.y,
+            labelOpacity: 1,
+        }),
+        update: barTotal => ({
+            x: barTotal.x,
+            y: barTotal.y,
+            labelOpacity: 1,
+        }),
+        leave: barTotal => ({
+            x: layout === 'vertical' ? barTotal.x : barTotal.x - 50,
```
plouc, line 311: I don't think we should have hardcoded values.

packages/bar/src/Bar.tsx
```diff
@@ -362,6 +404,23 @@ const InnerBar = <RawDatum extends BarDatum>({
         )
     }
 
+    if (layers.includes('totals') && enableTotals) {
+        layerById.totals = (
+            <Fragment key="totals">
+                {totalsTransition((style, barTotal) =>
+                    createElement(BarTotal, {
```
plouc, line 411: As the component is not customizable, we can use JSX directly rather than `createElement`, probably easier for typings.

packages/bar/src/BarCanvas.tsx
```diff
@@ -285,6 +311,8 @@ const InnerBarCanvas = <RawDatum extends BarDatum>({
         ]
     )
 
+    const formatValue = useValueFormatter(valueFormat)
```
plouc, line 314: Rather than duplicating the formatting logic, I think it should be part of the hook computing the totals, which would provide an extra `formattedValue`.

website/src/data/components/bar/props.ts
```diff
@@ -416,6 +416,16 @@ const props: ChartProperty[] = [
         control: { type: 'inheritedColor' },
         group: 'Labels',
     },
+    {
```
plouc, line 419: `totalsOffset` should be added as well.

packages/bar/src/types.ts
```diff
@@ -259,6 +258,9 @@ export type BarCommonProps<RawDatum> = {
     renderWrapper?: boolean
 
     initialHiddenIds: readonly (string | number)[]
+
+    enableTotals?: boolean
```
plouc, line 262: No need for `?` for values having a default, the top components are using `Partial`, and underlying components are going to inherit the default value, so this won't be optional anymore.

plouc (changes requested):

packages/bar/src/BarTotals.tsx
plouc: Thank you.

packages/bar/src/Bar.tsx
plouc: Perfect, thank you.

packages/bar/src/compute/totals.ts
```diff
@@ -56,7 +57,7 @@ export const computeBarTotals = <RawDatum extends BarDatum>(
                 key: 'total_' + indexValue,
                 x: xPosition,
                 y: yPosition,
-                value: indexTotal,
+                value: formatValue(indexTotal),
```
plouc, line 64: I think we should keep both the value and the `formattedValue`, that's usually what we do in other components.

plouc (approved):
LGTM, just a small comment about keeping the value along with the formattedValue, good job! I tried the feature in the vercel preview, and it's really nice! I think we could allow the customization of the component (later).

Maxim278:
@plouc is there an information about when this change is going to be released?

dbuhon:
It works now on `v0.86.0` with:
```ts
<ResponsiveBar 
  enableTotals
  layers={[..., 'totals']}
  ...
/>
```

plouc:
> It works now on `v0.86.0` with:
> 
> ```ts
> <ResponsiveBar 
>   enableTotals
>   layers={['bars', 'totals', ...]}
>   ...
> />
> ```

Please note that you don't need to specify the layers as the `totals` layer is included by default, unless you want to change their order or add a custom layer.

Maxim278:
is it possible to format totals value?

criveratro:
How can I format the total value? Let's say I want to show `1K` instead of `1000`.

PR #2528, opened 2024-03-06
Fix (peer) dependencies

### LINKED ISSUE #2401: 0.83.0 Doesn't work with pnpm + webpack
**Describe/explain the bug**
When using 0.83.0 with wepback + _pnpm_, something goes wrong in the dep tree, and the result is that all the Container.js contexts hooks produce undefined values wherever they are called, resulting in lots of runtime errors and blank charts.

Runtime Errors are like
```sh
Cannot read properties of undefined (reading 'animate')
    TypeError: Cannot read properties of undefined (reading 'animate')
```

**To Reproduce**
Does not reproduce using npm, but switching this repo from `npm` to `pnpm` reproduced it immediately.

https://github.com/ollwenjones/nivo-0830-test

**Expected behavior**
Since nivo uses pnpm internally (per contributing page) one would expect it to work with pnpm

**Screenshots**

Screenshot 2023-07-24 at 4 06 06 PM

 - OS: Building on Mac OS or Windows, doesn't matter here.
     - Node 16, pnpm 17, webpack 5 (setup per Create React App)
 - Browser: Chrome
 - Version 114.0.5735.198

**Additional context**
Really tried to solve this, and will contribute a fix if I can figure it out, but had to timebox it. My guess is some dependency problem caused by pnpm isolating modules is being masked by these react-contexts-are-blank runtime errors.  Tried to find a duplicate version of React, but was unable to in the time that I had.

marvinruder:
* Declare `prop-types` as a dependency of `@nivo/core` (as recommended in their README)
* Declare `react` as a peer dependency of `@nivo/tooltip`
* Declare correct version of `pnpm`
* Update lockfile

Fixes #2401

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,5 +90,5 @@
     "type": "opencollective",
     "url": "https://opencollective.com/nivo"
   },
-  "packageManager": "pnpm@8.3.1"
+  "packageManager": "pnpm@8.10.2"
 }
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -32,10 +32,10 @@
         "d3-scale-chromatic": "^3.0.0",
         "d3-shape": "^1.3.5",
         "d3-time-format": "^3.0.0",
-        "lodash": "^4.17.21"
+        "lodash": "^4.17.21",
+        "prop-types": "^15.7.2"
     },
     "peerDependencies": {
-        "prop-types": ">= 15.5.10 < 16.0.0",
         "react": ">= 16.14.0 < 19.0.0"
     },
     "publishConfig": {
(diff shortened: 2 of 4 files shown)


marvinruder:
Just imported the packages `@nivo/core@https://pkg.csb.dev/plouc/nivo/commit/eca9bac0/@nivo/core/_pkg.tgz` and  `@nivo/sunburst@https://pkg.csb.dev/plouc/nivo/commit/eca9bac0/@nivo/sunburst/_pkg.tgz` into my project and I can confirm that this indeed solves #2401 for me without any additional declarations in `.yarnrc.yml`.

plouc (approved):
LGTM, thank you for investigating and solving the issue.

PR #2541, opened 2024-03-14
feat(Line): passing the full Point info to the `AccessorFunc` for Label

akassaei:
💥 Breaking change
This is a rework PR for what was proposed for #1981 .

We achieve the same goal but instead pass the whole `Point` informations to the `AccessorFunc` directly.
Compared to the first version of the PR, this one comes with :
- Updates related to the Website to not break existing examples
- Unit tests for the case with a function and the case with a string passed to the `pointLabel` prop

### DIFF
--- a/packages/line/index.d.ts
+++ b/packages/line/index.d.ts
@@ -96,7 +96,7 @@ export interface Point {
     }
 }
 
-export type AccessorFunc = (datum: Point['data']) => string
+export type AccessorFunc = (datum: Point) => string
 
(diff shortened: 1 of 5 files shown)


dubzzz:
@plouc Sorry for the direct ping, but I feel this new version of https://github.com/plouc/nivo/pull/1981 follows your recommendations done at https://github.com/plouc/nivo/pull/1981#issuecomment-1989659736. Anything to change?

plouc:
Thank you @dubzzz and @akassaei!

PR #2558, opened 2024-04-03
Fix area rendering disparity of line canvas compared to line svg

MatinAniss:
This PR fixes the inconsistent rendering of `enableArea` when comparing SVG to Canvas, seen below is a example of the issue. The SVG line renders the area correctly when using a stacked line graph but the current implementation of the Canvas line renders it improperly.

This fix will make the canvas version conform with how the svg version is displayed.

SVG Line Area example (Current Good ✔️):
svg-rendered-normal

Canvas Line Area example (Current Not Good ❌):
canvas-rendered-broken

Canvas Line Area example (Fixed Good ✔️):
cavas-rendered-fixed

### DIFF
--- a/packages/line/src/LineCanvas.js
+++ b/packages/line/src/LineCanvas.js
@@ -193,12 +193,12 @@ const LineCanvas = props => {
                 ctx.globalAlpha = areaOpacity
 
                 areaGenerator.context(ctx)
-                series.forEach(serie => {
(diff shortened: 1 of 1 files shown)


MatinAniss:
Here is another example from the `nivo.rocks` website which shows this improvement, it looks much better!

Fixed New:
image

Old:
image

PR #2566, opened 2024-04-19
feat(bar): added aria hidden and disabled flag

jingcai86:
we need support for aria hidden/disabled bar for accessibility in case if there is a empty bar


plouc:
@jingcai86, thank you for your contribution, these new properties should be added to the documentation as well.

Issue #2591, opened 2024-05-16
🚩Custom Legend causes infinite re-render! 🚩


ZombieChowder11:
**Describe/explain the bug**
Using `forwardLegendData` with `ResponsivePie` causes the page re-render infinitely!

**To Reproduce**
Sandbox reproducing the issue: https://codesandbox.io/p/sandbox/focused-euler-sq9s3p

1. Go to the sandbox
2. Open the console and watch the page re-render.

**Expected behavior**
The expected behavior is for the page **NOT** to re-render infinitely whenever someone creates a custom legend. 

My code is based on the official storybook documentation: https://github.com/plouc/nivo/blob/master/storybook/stories/pie/Pie.stories.tsx#L308

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome
 - Version: 124.0.6367.207

plouc:
Duplicate of https://github.com/plouc/nivo/issues/2569.

Issue #2569, opened 2024-04-25
Infinite renders with custom legend


ManishBisht777:
**Describe/explain the bug**
On using `forwardLegendData` with `ResponsivePie` for custom legend the component rerenders infinitely.

**To Reproduce**
This is the codesandbox link for reproduction
https://codesandbox.io/p/devbox/objective-tree-h63wjz?file=%2Fsrc%2FApp.tsx

**Expected behavior**
according to https://github.com/plouc/nivo/blob/master/storybook/stories/pie/Pie.stories.tsx inside the `PieWithCustomLegend` function, it should work fine and render a table as per the storybook

**Screenshots**
![image](https://github.com/plouc/nivo/assets/89926834/65af5e4f-5000-4242-8f29-0cabe1d2b2fd)

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Chrome
 - Version: 0.85.1

ZombieChowder11:
Have you found a solution to this problem?

I'm having a similar issue and can't find information about it anywhere. If I find a solution, I'll post back here.

ManishBisht777:
> Have you found a solution to this problem?
> 
> I'm having a similar issue and can't find information about it anywhere. If I find a solution, I'll post back here.

hey @ZombieChowder11 we couldn't find a proper solution or workaround for this, we switched to Echarts since we are just trying out different libs for our CRM dashboard

ZombieChowder11:
> > Have you found a solution to this problem?
> > I'm having a similar issue and can't find information about it anywhere. If I find a solution, I'll post back here.
> 
> hey @ZombieChowder11 we couldn't find a proper solution or workaround for this, we switched to Echarts since we are just trying out different libs for our CRM dashboard

I'm going to try and escalate this to one of the supporters. Otherwise I'm gonna have to switch libraries as well. 
@ManishBisht777 Thanks for replying tho'

PR #2592, opened 2024-05-16
fix:removing forwardLegendDataRef from hook dependency

ZombieChowder11:
**Related to issue #2591 #2569** 

This fix should remove the infinite re-renders when creating a custom Legend for responsive pies.

**forwardLegendDataRef** is not needed in the dependency of the useEffect hook

### DIFF
--- a/packages/pie/src/hooks.ts
+++ b/packages/pie/src/hooks.ts
@@ -180,7 +180,7 @@ export const usePieArcs = <RawDatum>({
     useEffect(() => {
         if (typeof forwardLegendDataRef.current !== 'function') return
         forwardLegendDataRef.current(legendData)
-    }, [forwardLegendDataRef, legendData])
+    }, [legendData])
 
     return result
 }


plouc:
@ZombieChowder11, thank you for submitting a fix.

I thought a ref's reference was stable, only `current` changes as I understand it.

plouc:
I found an interesting post about this: https://epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array/, so while it seems like this change is not going to fix the problem, not passing the ref and passing the ref object (not `current`) seems to lead to the same behavior.

plouc:
The problem might be about legend data changing.

plouc:
I think you could try locally the setup you shared in the issue you created with the updated code, creating a new story for example.

ZombieChowder11:
> I think you could try locally the setup you shared in the issue you created with the updated code, creating a new story for example, also log in `useEffect` to see when it runs.

Something that came to my mind was to either use a useMemo hook or a callback for the legend data changing instead of a useEffect.

plouc:
Given the simplicity of the data for the legends, maybe deeply comparing the data would be better, rather than relying on other computations which might be necessary for the chart, but won't impact the legend's data. It would probably be less prone to infinite rendering, it's really easy to end up in this situation if you don't declare the props statically/memoize them.

Issue #1378, opened 2021-01-06
Pie chart radial labels overlap


gindulis:
**To Reproduce**
Please see:
https://codesandbox.io/s/nivopie-total-inside-donut-434-forked-eq1yi?file=/src/index.js
![image](https://user-images.githubusercontent.com/2493521/103785203-994e2b80-5043-11eb-9ebd-b7385bec39a6.png)

**Expected behavior**
Radial labels should not overlap each other

Any way to make them not overlap? Using the `radialLabelsSkipAngle` is not an option they should be present at all times. All other options does not resolve the issue.

suchak1:
bump

tclarkston:
bump

RamblinGambler:
bump

darron1217:
same issue here

whirledsol:
bump

Jovells:
bump @plouc

dev-creata:
Hi @plouc Hi I'm trying to contribute to fix this Issue Can you point me to where I can find the function that computes the path? I found the compute.ts file but making edits to it has no effect on the render (I'm using the storybook)

davidkim10:
I wish I contributed my solution when I wrote it 2 years ago but I solved this issue by adjusting the computeRadialLabels function within nivo-pie. By adjusting this function and adjusting the `padAngle` value I was able to fix the overlap.

<img width="644" alt="Screen Shot 2020-05-29 at 5 09 11 PM" src="https://user-images.githubusercontent.com/56006438/158417767-766b03aa-b347-4dcb-9cad-35919aa29477.png">

<img width="680" alt="Screen Shot 2020-05-29 at 5 09 47 PM" src="https://user-images.githubusercontent.com/56006438/158418368-4b2d290a-2f11-4875-9d77-307a60167bce.png">

<img width="666" alt="Screen Shot 2020-05-29 at 5 11 28 PM" src="https://user-images.githubusercontent.com/56006438/158418390-b59d6b69-7979-414d-93f1-1fd977d5bd28.png">

<img width="658" alt="Screen Shot 2020-05-29 at 5 13 41 PM" src="https://user-images.githubusercontent.com/56006438/158418418-3c990fff-80e3-464b-93c2-4efc1f7d10a7.png">

Im trying to find the exact fix from a project I worked on 2 years ago but honestly cant recall my thought process at the time but i adjusted nivo-pie.esm.js

[...]

Issue #143, opened 2018-02-02
feature request: fix overlapping pie radial labels


knackjax:
if the pie is drawn with a bunch of slices that are small, the result is the radial labels are all overlapping as shown in this example

https://www.webpackbin.com/bins/-KylBOM-OLUpXZ8zOhY-

is there a way to fix this? I was thinking maybe varying the horizontal and diagonal lengths between slices might work...

thanks

DaveyEdwards:
Did you ever come up with a solution to this?  Ive just started noticing this and going to try a couple hack fixes.  Your url is broken now so here is a image of the issue for others to see (It gets worse the more small % valued parts there are)


![image](https://user-images.githubusercontent.com/15203899/39828865-62349d04-5371-11e8-865f-e73e67f2d26b.png)

knackjax:
Yes I went in and modified the PieRadialLabels js with an offset variable that gives greater numbers if the height calculated by sin(angle) is a large value.

It is pretty hacky so I didn't bother to contribute back.

linkBase = 5 and linkExponent = 2 seem to work best for my usage but you might have to tweak for your own use

>                 {data.filter(d => skipAngle === 0 || d.angleDegrees > skipAngle).map((d,i) => {
>                     const angle = midAngle(d) - Math.PI / 2
> 
>                     const offset = Math.pow(linkBase, Math.abs(Math.sin(angle)) * linkExponent)
> 
>                     const positionA = positionFromAngle(angle, radius + linkOffset)
>                     const positionB = positionFromAngle(
>                         angle,
>                         radius + linkOffset + linkDiagonalLength + offset
>                     )
>                     let positionC
>                     let labelPosition
>                     let textAnchor
>                     if (angle + Math.PI / 2 < Math.PI) {
>                         positionC = { x: positionB.x + linkHorizontalLength + offset, y: positionB.y }
>                         labelPosition = {
>                             x: positionB.x + linkHorizontalLength + textXOffset + offset,
>                             y: positionB.y,
>                         }
>                         textAnchor = 'start'
>                     } else {
>                         positionC = { x: positionB.x - linkHorizontalLength - offset, y: positionB.y }
>                         labelPosition = {
>                             x: positionB.x - linkHorizontalLength - textXOffset - offset,
>                             y: positionB.y,
>                         }
>                         textAnchor = 'end'
>                     }
![image](https://user-images.githubusercontent.com/20896041/39830082-5f067914-538e-11e8-9fca-8c34d1d7d315.png)

[...]

PR #2599, opened 2024-05-30
fix(pie chart): fix overlapping labels

calum-atkinson:
Fix based on the solution found in the following issue: https://github.com/plouc/nivo/issues/1378

Effectively, the solution recalculates the points on the label when segments of the pie are small (π/10 radians chosen as a useful value) so that more labels are able to be displayed without interference with one another.

It's not an exhaustive solution as when the segments become very small (<1% of the pie) there can still be overlapping, but this solution should improve the general case. A more complete solution may need an awareness of the neighbouring segments and the space around the pie.

### DIFF
--- a/packages/arcs/src/arc_link_labels/compute.ts
+++ b/packages/arcs/src/arc_link_labels/compute.ts
@@ -1,4 +1,4 @@
-import { positionFromAngle } from '@nivo/core'
+import { positionFromAngle, radiansToDegrees } from '@nivo/core'
 import { Arc, Point } from '../types'
 import { getNormalizedAngle } from '../utils'
 import { ArcLink } from './types'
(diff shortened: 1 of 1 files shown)


brian-codes:
Also issue #143

MartinHutyra:
@calum-atkinson is it working as expected? If so, @plouc can you please do the review?

brian-codes:
@plouc do you have an example dataset it's worse with?

plouc:
@brian-codes, I just played with the deployed demo, and generated random dataset, see this screenshot for example:

CleanShot 2024-07-24 at 18 36 00@2x

IMO, it would be better if labels were all the same and personally, I wouldn't want this behavior on by default.

PR #2608, opened 2024-06-13
add lodash to nivo/line's package.json

jonaengs:
https://github.com/plouc/nivo/blob/da6d6b751cff5570fb7378d57b52e3e4bdbcf872/packages/line/src/hooks.js#L6

`nivo/line` imports `uniqueId` from lodash in the hooks.js file. However, lodash is not declared as a dependency in `nivo/line`'s package.json file. This was causing my project to crash when using the package with pnpm. 

The PR fixes the issue by adding lodash as a dependency


plouc:
@jonaengs, the lock file needs to be updated accordingly, most jobs are failing due to this.

jonaengs:
Hey @plouc thanks for the tip. I'm not super experienced with javascript development, so all input is appreciated:) I think I've updated the lockfile correctly, but I see that the workflows are still failing, so maybe I missed something? I've allowed edits by maintainers for the PR, so if you spot my mistake and want to quickly fix it yourself, feel free. Otherwise I'll do my best to fix the issues myself. Though I might need some more pointers before I'm able to make meaningful progress.

PR #2651, opened 2024-09-24
Add getColor to props provided to custom layers

lemps:
I was creating a custom layer for my bar chart and didn't find a way to access the colours passed to the chart.

Figured that the most correct approach is to actually pass `colors` and `colorBy` props to the layer so I could use `const getColor = useOrdinalColorScale(colors, colorBy)` in my custom layer.

### DIFF
--- a/packages/bar/src/types.ts
+++ b/packages/bar/src/types.ts
@@ -124,6 +124,7 @@ interface BarCustomLayerBaseProps<RawDatum>
 
     xScale: AnyScale
     yScale: AnyScale
+    getColor: OrdinalColorScaleConfig<ComputedDatum<RawDatum>>
 }
 
 export interface BarCustomLayerProps<RawDatum>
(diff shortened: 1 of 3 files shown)


plouc:
`getColor` is already returned by the `useBar` hook, so I'd suggest to pass this instead of recomputing the same color scale inside the custom layer.

lemps:
Thanks for the feedback. Updated the PR to pass getColor along instead

packages/bar/src/types.ts
```diff
@@ -124,6 +124,7 @@ interface BarCustomLayerBaseProps<RawDatum>
 
     xScale: AnyScale
     yScale: AnyScale
+    getColor: OrdinalColorScaleConfig<ComputedDatum<RawDatum>>
```
plouc, line 127: The type is `OrdinalColorScale<ComputedDatum<RawDatum>>` at this point it's not the config anymore.

plouc (changes requested):

tombennet:
@plouc This would be useful for us too - can we get this one merged? Thanks!

plouc (approved):
LGTM

Issue #2597, opened 2024-05-28
Line Chart : crash when hovering an empty chart that is interactive


akassaei:
**Describe/explain the bug**
When hovering a Line Chart with empty data, the chart crashes.

**To Reproduce**
Here is the link to a Codesandbox showing the issue : https://codesandbox.io/p/sandbox/charming-mendel-3wsx4v?file=%2Fsrc%2Fcomponents%2FLineChart.js%3A110%2C25

1. Create a Line Chart with empty Data
2. Hover the Chart with your mouse
3. See the crash

**Expected behavior**
No crash 😅 

**Additional context**
I spotted where the issue is coming from : basically the `findNode` method in Mesh hooks relies on `d3-delaunay` and specifically on `delaunay.find` method. With no data points given, the `.find` returns NaN and it isn't a case checked in the code so we try to access the `node` related to the `index` and it crashes because it's `undefined`.https://github.com/plouc/nivo/blob/da6d6b751cff5570fb7378d57b52e3e4bdbcf872/packages/voronoi/src/hooks.ts#L172-L179

I've already opened an issue on `d3-delaunay` side because I don't think `NaN` is really an expected behavior 😅 You can find the issue opened here : https://github.com/d3/d3-delaunay/issues/144

On our side, I've patched Nivo to return `null` when `nodes.length === 0`. So I have a fix ready if you want me to handle it, but I'm not sure that's how you want to handle it 😄

gercog1:
hey @akassaei!

Did you find a quick workaround for this one? 

I also struggling ;)

robiscoding:
Any update here?

akassaei:
@gercog1 @robiscoding Sorry didn't see your messages
On my side I've directly patched the Nivo package on our side using `patch-package`, as said in the original description I've just changed the code to return `null` when `nodes.length === 0`

@plouc is that something you can look at please ?
I can definitely open a PR for this with same change I've done in my patch

plouc:
@akassaei, a PR would be nice :)

akassaei:
Hey @plouc 👋 
A PR was made just here: https://github.com/plouc/nivo/pull/2669
Whenever you have time it would be awesome to have a review please 🙏

plouc:
Fixed by #2669

diego9497:
Is this already part of any release?, I'm still having the issue

PR #2669, opened 2024-11-14
🐛  [#2597] Prevent Line Chart from crashing when having no data

akassaei:
This PR fixes issue described on #2597 
The issue we had was that we are calling `delaunay.find` function on empty data, it was fixed on their side but new version was not deployed yet so we still need to make a fix around this.
The simple fix I found was to early return when we have `nodes.length === 0`.
I've also added a simple test around Line Chart


plouc:
@akassaei, thank you for your contribution!

plouc (commented):
LGTM

akassaei:
Thanks a lot for the review and the time that you put in this @plouc 🙏 ❤️

PR #2670, opened 2024-11-14
feat: Pie: support multiple `activeId`s

assaf-aviram:


packages/pie/src/hooks.ts
```diff
@@ -20,6 +20,9 @@ import {
     CommonPieProps,
 } from './types'
 
+const idIsActive = (id: DatumId, activeId: null | DatumId | DatumId[]) =>
```
plouc, line 23: Rather than supporting a scalar or an array, I think it would be better to rename the property to `activeIds` and to support arrays exclusively.

It would also be nice to add tests about this new feature and to adjust the documentation.

storybook/stories/pie/Pie.stories.tsx
```diff
@@ -282,6 +282,76 @@ export const ControlledActiveId: Story = {
     render: () => <ControlledPies />,
 }
 
+const ControlledPiesMultipleActiveIds = () => {
+    const [activeId, setActiveId] = useState<string>(commonProperties.data[1].id)
+    const [additionalActiveIds, setAdditionalActiveIds] = useState<string>([])
```
plouc, line 287: The type should be `string[]` I think.

plouc (changes requested):

packages/pie/src/hooks.ts
plouc: I understand, but it makes the code and typing more complex, I'd prefer to have something a bit simpler to maintain, also the name of the property is now confusing IMHO.

assaf-aviram: do you think it's a better idea to keep `activeId` as it was, and add a new `additionalActiveIds ([])`?
so we don't break existing charts

plouc:
I'm closing this PR since it's been inactive for a while now, I can reopen in case it's updated.

PR #2681, opened 2025-01-13
marimekko: all dimension input data is propagated

### LINKED ISSUE #2677: marimekko: cannot set custom color via datum
**Describe/explain the bug**
I want to use a Marimekko chart and configure custom colors using the `{ datum: someField }` syntax.  This works nicely in other charts, but for Marimekko this seems to not be behaving the same. 

**To Reproduce**

https://codesandbox.io/p/sandbox/cntjxj

**Expected behavior**

Being able to configure the color per dimension by using the "datum" color variant

 - OS: Linux
 - Browser Chromium
 - Version 130

**Additional information**

I'm not very familiar with the codebase, but my initial investigation lead to `hooks.ts:233` where `getColor` gets the `dimensionDatum` which is in the process of being constructed. This does not have access to properties that are defined on the original `dimensions` and that are not otherwise used like `id` and `value`. 

Therefore, this also partially affects the `custom` color variant as it doesn't have access to all the input properties.

Linschlager:
Currently, only the dimensionId as well as the patched value function are propagated to the computed data.
In order to use custom colors per dimension, it is necessary to propagate the entire object to the computed one so the dimension data is still available.

Resolves #2677

### DIFF
--- a/packages/marimekko/src/hooks.ts
+++ b/packages/marimekko/src/hooks.ts
@@ -146,7 +146,7 @@ export const useThicknessScale = <RawDatum>({
 export const useComputedData = <RawDatum>({
     data,
     stacked,
-    dimensionIds,
+    rawDimensions,
     valueFormat,
     thicknessScale,
     dimensionsScale,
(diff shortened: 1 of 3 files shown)


plouc:
Hi @Linschlager,

First of all, thank you for this contribution!

I'm not convinced that merging the raw dimension directly into the `DimensionDatum` is the best approach.
I think this could be confusing, as we could end up with some arbitrary properties in `DimensionDatum`, especial ly for users who don't use TypeScript.
It would probably be best to isolate, for example adding `DimensionDatum.original` or something similar, and maybe to use a generic for the dimensions if needed.
This approach would require using the `colors` property to get the color from `DimensionDatum.original`, but I think it would be _cleaner_.

Please let me know what you think.

plouc (commented):
LGTM

PR #2689, opened 2025-02-18
Remove fill property in BarTotals.tsx

MatthewDlr:
Hey there 👋🏻,

I recently noticed that the color of the total label in stacked bar charts is being overridden by 'fill: theme.text.fill'. 
It seems more logical to use the 'theme.labels.text' props instead, since it's specifically for labels. 

What are your thoughts on this @plouc?


plouc:
Yes, I think it makes sense.

PR #2696, opened 2025-03-06
feat(sunburst): add innerRadius and renderRootNode props

uson1x:
Hey

Thank you for your great work on this library.

I saw a few innerRadius sunburst PRs and decided to make a fresh version with a storyboard for you.

Please, let me know if you need to change anything in order to get this PR to be ready to be merged. I'd be happy to rewrite the PR as needed.

Right now we are pressed to use a custom version of nivo because in our project we need to control the innerRadius, but it would be so much easier to work with if it was in the main repository.

Thank you for taking a look at it!


plouc:
@uson1x, I reviewed the previous PR as it already came from another PR 😅
I think you could submit a PR once the other one is merged, with the added story.

plouc:
As the previous one is inactive, we should probably go for yours instead.

packages/sunburst/package.json
```diff
@@ -35,12 +35,16 @@
...
         "lodash": "^4.17.21"
     },
     "peerDependencies": {
         "react": ">= 16.14.0 < 19.0.0"
     },
     "publishConfig": {
         "access": "public"
+    },
+    "devDependencies": {
+        "@types/d3-scale": "^4.0.8"
```
plouc, line 48: I usually put the types in `dependencies`, as for `@types/d3-hierarchy`, in general types are only needed at compile time, but in case these typings are required for the types exported by nivo, this could be an issue.

packages/sunburst/src/hooks.ts
```diff
@@ -32,6 +35,8 @@ export const useSunburst = <RawDatum>({
     valueFormat?: DataProps<RawDatum>['valueFormat']
     radius: number
     cornerRadius?: SunburstCommonProps<RawDatum>['cornerRadius']
+    innerRadius?: SunburstCommonProps<RawDatum>['innerRadius']
```
plouc, line 38: The properties should be added to the dependencies of `useMemo`, if you check the live demo on vercel, you can see that changing these values doesn't do anything until you change other properties.

packages/sunburst/src/hooks.ts
```diff
@@ -68,6 +75,12 @@ export const useSunburst = <RawDatum>({
         // are going to be computed first
         const sortedNodes = sortBy(descendants, 'depth')
 
+        const innerRadiusOffset = radius * Math.min(innerRadius, 1)
+
+        const maxDepth = Math.max(...sortedNodes.map(n => n.depth))
+
+        const scale = d3ScaleRadial().domain([0, maxDepth]).range([innerRadiusOffset, radius])
```
plouc, line 82: nit: `radiusScale`?

packages/sunburst/src/hooks.ts
```diff
@@ -82,8 +95,12 @@ export const useSunburst = <RawDatum>({
             const arc: Arc = {
                 startAngle: descendant.x0,
                 endAngle: descendant.x1,
-                innerRadius: Math.sqrt(descendant.y0),
-                outerRadius: Math.sqrt(descendant.y1),
+                innerRadius:
+                    renderRootNode && descendant.depth === 0 ? 0 : scale(descendant.depth - 1),
```
plouc, line 99: nit: `const isRootNode = renderRootNode && descendant.depth === 0`, which can then be used for both inner/outer radii.

pnpm-lock.yaml
plouc, line 1: It seems weird to have so many changes on the lock file.

packages/sunburst/src/Sunburst.tsx
```diff
@@ -25,6 +25,8 @@ const InnerSunburst = <RawDatum,>({
     data,
     id = defaultProps.id,
     value = defaultProps.value,
+    innerRadius = defaultProps.innerRadius,
+    renderRootNode = defaultProps.renderRootNode,
```
plouc, line 29: I don't think we should use the same rendering as for the arcs for the root node, it generates a shape that doesn't really make sense:

CleanShot 2025-04-18 at 21 34 05@2x

Which means it cannot have a color, interactivity doesn't work...

packages/sunburst/src/Sunburst.tsx
```diff
@@ -25,6 +25,8 @@ const InnerSunburst = <RawDatum,>({
     data,
     id = defaultProps.id,
     value = defaultProps.value,
+    innerRadius = defaultProps.innerRadius,
```
plouc, line 28: It would be good to add some tests for these new properties.

plouc (changes requested):

pnpm-lock.yaml
uson1x, line 1: Sadly wasn't able to produce pnpm-lock.yaml with minimal changes, tried pnpm 9.15.0 (just to see if 9 works differently) and 10.10.0 (which is specified in package.json). Please let me know If there is a way

uson1x (commented):
Hi @plouc 

Thanks a lot for the review, I have hopefully addressed the feedback.
Please, let me know if I missed something.

pnpm-lock.yaml
plouc: The changes should be ok, I experience big changes quite often as well, which aren't problematic.

plouc (changes requested):

plouc:
@uson1x, I think you forgot to run `make fmt`.

packages/sunburst/src/Sunburst.tsx
plouc: It's a bit better, but I think using an arc doesn't work very well tbh.
When enabling/disabling the root node, we can clearly see that the shape for the root node isn't a circle but more like a donut, since it's an arc of 360 degrees.
While the donut shape isn't necessarily critical, the impact it has on the position of the root node label is IMHO, the text should be at the center.
An option (and a bit of a hack) would be to adjust `computeArcCenter` so that it handles arcs of 360 deg with an inner radius of 0 differently (and `arcLabelsRadiusOffset` should probably be ignored for this case). That's probably the easiest option.

plouc: If we go for this approach, we should test the position of the text in the root node, and also make sure that a ring of 360 deg still behave as before.

PR #2701, opened 2025-03-29
fix(sankey): use readonly array types in props

pcorpet:
The component actually clones data that it mutates, and only iterate over the other arrays, so making them readonly is restricting the props types without any change.


plouc:
@pcorpet, yes, in fact all charts should receive read-only data, it's also important for memoization, thank you!

PR #2704, opened 2025-04-02
feat(funnel): add support for size overrides

langarus:
Enhancements to the Funnel Component

I've introduced new a new prop to allow the customisation of funnel width:

### Sizes prop
Can be one of the following:
- an `array of numbers`
- a data selector `{datum: "size"}`
- a function of this format `(data: Datum) => number`

Let me know if you have any questions or suggestions! 🚀


packages/funnel/src/Funnel.tsx
```diff
@@ -23,6 +23,7 @@ const InnerFunnel = <D extends FunnelDatum>({
     spacing = svgDefaultProps.spacing,
     shapeBlending = svgDefaultProps.shapeBlending,
     valueFormat,
+    valueFormatter,
```
plouc, line 26: I think this should be added to the documentation, we might also want to add a story about it, or to make it available via the playground.

plouc: Some tests would also be nice.

packages/funnel/src/hooks.ts
```diff
@@ -334,7 +336,8 @@ export const useFunnel = <D extends FunnelDatum>({
     const getBorderColor = useInheritedColor(borderColor, theme)
     const getLabelColor = useInheritedColor(labelColor, theme)
 
-    const formatValue = useValueFormatter<number>(valueFormat)
+    const _formatValue = useValueFormatter<number>(valueFormat)
```
plouc, line 339: `useValueFormatter` already uses the provided formatter if it's a function:

```
export const getValueFormatter = format => {
    // user defined function
    if (typeof format === 'function') return format

    if (typeof format === 'string') {
        // time format specifier
        if (format.indexOf('time:') === 0) {
            return d3TimeFormat(format.slice('5'))
        }

        // standard format specifier
        return d3Format(format)
    }

    // no formatting
    return value => `${value}`
}
```

packages/funnel/src/types.ts
```diff
@@ -87,7 +93,7 @@ export interface FunnelCommonProps<D extends FunnelDatum> {
     layers: (FunnelLayerId | FunnelCustomLayer<D>)[]
 
     valueFormat: ValueFormat<number>
-
+    valueFormatter?: (value?: number) => string
```
plouc, line 96: You can probably use `ValueFormat` here.

packages/funnel/src/hooks.ts
```diff
@@ -396,7 +405,7 @@ export const useFunnel = <D extends FunnelDatum>({
                 data: datum,
                 width: partWidth,
                 height: partHeight,
-                color: getColor(datum),
+                color: datum.overrides?.color || getColor(datum),
```
plouc, line 408: `getColor` could be customized to pick a custom color defined on the datum. I think we could also add a test for this.

packages/funnel/src/types.ts
```diff
@@ -9,6 +9,10 @@ export interface FunnelDatum {
     id: string | number
     value: number
     label?: string
+    overrides?: {
```
plouc, line 12: While normalization is good, users often complained about having to format their data to match nivo's schema, maybe we could use a generic instead and allow users to pick the color (which is already supported) and size via a custom function.

packages/funnel/src/types.ts
```diff
@@ -42,6 +46,8 @@ export interface FunnelPart<D extends FunnelDatum> extends BoxPosition {
     borderOpacity: number
     labelColor: string
     formattedValue: number | string
+    tooltipName?: string
```
plouc, line 49: Are these used?

plouc (changes requested):

packages/funnel/src/types.ts
langarus: no, it was for some tests. I removed them

packages/funnel/src/types.ts
langarus: Thx for all the suggestions, apparently I was able to achieve all customisation of color and value formatting via the props `color` and `valueFormat` as you suggested.

I was still missing the `size` formatting, so I've created functionality similar to colors.

packages/funnel/src/hooks.ts
langarus: You're right, I removed the override

packages/funnel/src/hooks.ts
langarus: true, removed

packages/funnel/src/Funnel.tsx
```diff
@@ -24,6 +24,7 @@ const InnerFunnel = <D extends FunnelDatum>({
     shapeBlending = svgDefaultProps.shapeBlending,
     valueFormat,
     colors = svgDefaultProps.colors,
+    sizes = svgDefaultProps.sizes,
```
plouc, line 27: `size` should be added to the website, in https://github.com/plouc/nivo/blob/master/website/src/data/components/funnel/props.ts. 

I think it would also be good to add tests for the different options.

packages/funnel/src/hooks.ts
```diff
@@ -263,6 +268,60 @@ export const computePartsHandlers = <D extends FunnelDatum>({
...
+    config: OrdinalSizeScaleConfig<Datum>
+): config is OrdinalSizeScaleConfigDatumProperty => {
+    return (config as OrdinalSizeScaleConfigDatumProperty).datum !== undefined
+}
+
+// Get size based on configuration
+export const getOrdinalSizeScale = <Datum extends FunnelDatum>(
+    config: OrdinalSizeScaleConfig<Datum>
+) => {
+    // // User defined function
```
plouc, line 281: Duplicated `//`

packages/funnel/src/hooks.ts
```diff
@@ -263,6 +268,60 @@ export const computePartsHandlers = <D extends FunnelDatum>({
...
+            // Ensure we're using the id property from the datum to map to the appropriate size
+            return Number(scale(String(datum.id)))
+        }
+
+        return generator as OrdinalSizeScaleConfigCustomFunction<Datum>
+    }
+
+    if (isPlainObject(config)) {
+        // Use size from current datum
+        // use color from current datum
```
plouc, line 304: This should be removed.

packages/funnel/src/hooks.ts
```diff
@@ -263,6 +268,60 @@ export const computePartsHandlers = <D extends FunnelDatum>({
...
+    config: OrdinalSizeScaleConfig<Datum>
+) => {
+    // // User defined function
+    if (typeof config === 'function') {
+        return config
+    }
+
+    // Static number
+    if (typeof config === 'number') {
+        return () => config
```
plouc, line 288: I'm not sure we need this option, it could be easily achieved with the array, and it seems a bit weird to have a funnel with all parts having the same size.

packages/funnel/src/hooks.ts
```diff
@@ -263,6 +268,60 @@ export const computePartsHandlers = <D extends FunnelDatum>({
...
 }
 
+const isOrdinalSizeScaleConfigDatumProperty = <Datum>(
+    config: OrdinalSizeScaleConfig<Datum>
+): config is OrdinalSizeScaleConfigDatumProperty => {
+    return (config as OrdinalSizeScaleConfigDatumProperty).datum !== undefined
+}
+
+// Get size based on configuration
+export const getOrdinalSizeScale = <Datum extends FunnelDatum>(
```
plouc, line 278: If not set I think we should fallback to the `linearScale` currently used, and remove the simple number support.

packages/funnel/src/hooks.ts
```diff
@@ -370,19 +432,20 @@ export const useFunnel = <D extends FunnelDatum>({
     const parts: FunnelPart<D>[] = useMemo(() => {
         const enhancedParts = data.map((datum, index) => {
             const isCurrent = datum.id === currentPartId
+            const customSize = sizes ? getSize(datum) : linearScale(datum.value)
```
plouc, line 435: `useOrdinalSizeScale` should return the linearScale as a default IMHO.

packages/funnel/src/types.ts
```diff
@@ -5,12 +5,30 @@ import { InheritedColorConfig, OrdinalColorScaleConfig } from '@nivo/colors'
...
+    datum: string
+}
+
+export type OrdinalSizeScaleConfig<Datum = any> =
+    | OrdinalSizeScaleConfigStaticSize
+    | OrdinalSizeScaleConfigCustomFunction<Datum>
+    | OrdinalSizeScaleConfigCustomSizes
+    | OrdinalSizeScaleConfigDatumProperty
+
+type StandardFunnelDatum = {
```
plouc, line 22: nit: `BaseFunnelDatum`?

plouc (changes requested):

packages/funnel/src/hooks.ts
langarus: done

packages/funnel/src/hooks.ts
langarus: Done

website/src/data/components/funnel/props.ts
```diff
@@ -109,6 +109,35 @@ const props: ChartProperty[] = [
         flavors: allFlavors,
         defaultValue: defaults.colors,
     }),
+    {
+        key: 'sizes',
+        group: 'Style',
+        help: 'Define how to compute the size of funnel parts.',
+        description: `
```
plouc, line 116: 👍🏻

packages/funnel/src/Funnel.tsx
langarus: Done - this is how it looks locally. Let me know if you'd like me to make some changes

Untitled

plouc: Looks good, it's nice to have some examples.

packages/funnel/src/hooks.ts
langarus: Totally makes sense, I've removed it

packages/funnel/src/hooks.ts
langarus: Done, I've also tested it and it works fine.

Would it make sense in the future to add also a logaritmicScale for better smoothing/normalization between numbers with huge discrepancies? Allow the user to chose the normalization or we could also have "auto" where internally we could figure out the best one to use. So the prop sizes could also be a string like this `sizes: "linear" | "log" |"auto"` - besides the ones we already have (like function, array, selector).

 This is just an idea maybe for another PR.

packages/funnel/src/hooks.ts
langarus: great point, fixed it. Check the previous comment

packages/funnel/src/hooks.ts
```diff
@@ -18,6 +20,8 @@ import {
     FunnelAreaPoint,
     FunnelBorderGenerator,
     Position,
+    OrdinalSizeScaleConfig,
```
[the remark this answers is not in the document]
langarus: you're right, I've implemented size similar to `colors`, did not check the other components. It's updated now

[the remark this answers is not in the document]
plouc: Thank you.

plouc:
Thank you for your contribution @langarus!

PR #2709, opened 2025-04-16
Fix: No longer require legendOffsetX, legendOffsetY and legendOrientation props in Markers 

rinheimerle:
Came across [this issue](https://github.com/plouc/nivo/issues/2077#issuecomment-2747598791) while looking for some information on it as I am seeing these warnings in the browser console in my use case.


plouc:
`prop-types` were removed via https://github.com/plouc/nivo/pull/2714

PR #2710, opened 2025-04-20
feat(sankey): make it easy to use a different component for labels

pcorpet:
Before this change it was possible not to show the layer props and build an custom layer for labels, however the internal labels layer is pretty cool and compute a lot of useful thing. Exposing the component make it customizable when one wants only to change the render of each label

### DIFF
--- a/packages/sankey/src/SankeyLabels.tsx
+++ b/packages/sankey/src/SankeyLabels.tsx
@@ -1,7 +1,6 @@
 import { useSprings } from '@react-spring/web'
 import { useMotionConfig } from '@nivo/core'
 import { useTheme } from '@nivo/theming'
-import { Text } from '@nivo/text'
 import { DefaultLink, DefaultNode, SankeyCommonProps, SankeyNodeDatum } from './types'
 
 interface SankeyLabelsProps<N extends DefaultNode, L extends DefaultLink> {
@@ -9,6 +8,7 @@ interface SankeyLabelsProps<N extends DefaultNode, L extends DefaultLink> {
     layout: SankeyCommonProps<N, L>['layout']
     width: number
     height: number
+    labelComponent: SankeyCommonProps<N, L>['labelComponent']
     labelPosition: SankeyCommonProps<N, L>['labelPosition']
     labelPadding: SankeyCommonProps<N, L>['labelPadding']
     labelOrientation: SankeyCommonProps<N, L>['labelOrientation']
@@ -24,6 +24,7 @@ export const SankeyLabels = <N extends DefaultNode, L extends DefaultLink>({
     labelPadding,
     labelOrientation,
     getLabelTextColor,
+    labelComponent: LabelComponent,
 }: SankeyLabelsProps<N, L>) => {
     const theme = useTheme()
 
@@ -31,7 +32,7 @@ export const SankeyLabels = <N extends DefaultNode, L extends DefaultLink>({
     const labels = nodes.map(node => {
         let x
         let y
-        let textAnchor
+        let textAnchor: 'middle' | 'start' | 'end' | undefined
         if (layout === 'horizontal') {
             y = node.y + node.height / 2
             if (node.x < width / 2) {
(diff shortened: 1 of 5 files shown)


packages/sankey/src/SankeyLabels.tsx
```diff
@@ -97,7 +106,7 @@ export const SankeyLabels = <N extends DefaultNode, L extends DefaultLink>({
                 const label = labels[index]
 
                 return (
-                    <animated.text
+                    <LabelComponent
```
plouc, line 109: The problem with this is that if you don't use an animated element, the transform property and the fill style aren't going to work, because they're animated values.

packages/sankey/src/SankeyLabels.tsx
```diff
@@ -109,7 +118,7 @@ export const SankeyLabels = <N extends DefaultNode, L extends DefaultLink>({
                         }}
                     >
                         {label.label}
```
plouc, line 120: I think the component should decide what to do with the label, not have it as a children, for example, if you have long labels, you could want to split the text.

packages/sankey/src/types.ts
```diff
@@ -162,6 +168,7 @@ export interface SankeyCommonProps<N extends DefaultNode, L extends DefaultLink>
     labelPadding: number
     labelOrientation: 'horizontal' | 'vertical'
     labelTextColor: InheritedColorConfig<SankeyNodeDatum<N, L>>
+    labelComponent: ComponentType<SankeyLabelProps>
```
plouc, line 171: I usually only support a `FunctionComponent`, I don't think we should allow class components.

plouc (changes requested):
I think the component should be defined differently, not just as a wrapper, you could maybe have a look at how it's done in the `polar-axes` package for example, to support overriding the `tickComponent` in a `RadialAxisConfig`, the default being a `RadialAxisTick`.

packages/sankey/src/SankeyLabels.tsx
pcorpet: Actually, now with the `@nivo/text` package, it's cleaner. What do you think?

packages/sankey/src/types.ts
pcorpet: Fixed. thanks

packages/sankey/src/SankeyLabels.tsx
plouc: The props are good, with the exception of the label, which I think should be provided as well because the type of `children` is not a `string`,  which would make it a bit difficult to split the text for example.

packages/sankey/src/SankeyLabels.tsx
plouc: Don't you have issues with the type of `children`?

packages/sankey/src/types.ts
```diff
@@ -124,6 +126,11 @@ export interface CustomSankeyLayerProps<N extends DefaultNode, L extends Default
     isInteractive: boolean
 }
 
+export type SankeyLabelProps = Pick<
```
plouc, line 129: This is not needed anymore, right?

pcorpet: Well spotted. Removed.

packages/sankey/src/SankeyLabels.tsx
pcorpet: OK, I've sent the full node, which includes the label.

packages/sankey/src/SankeyLabels.tsx
```diff
@@ -13,6 +14,7 @@ interface SankeyLabelsProps<N extends DefaultNode, L extends DefaultLink> {
     labelPadding: SankeyCommonProps<N, L>['labelPadding']
     labelOrientation: SankeyCommonProps<N, L>['labelOrientation']
     getLabelTextColor: (node: SankeyNodeDatum<N, L>) => string
+    labelComponent: FunctionComponent<TextProps & { node: SankeyNodeDatum<N, L> }>
```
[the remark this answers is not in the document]
pcorpet: Done. Thanks for the fast iterations.

PR #2719, opened 2025-04-22
fix(website): fix typings

plouc:
The typings for the website are quite broken, some files are still in js, the purpose of this PR is to fix most of them, and try to add type checking for the website to the CI, although this might be difficult to do in one pass.

This PR also includes upgrading to node.js v22 and pnpm 10.


plouc:
Several types were fixed, but there are ~100 issues.

PR #2720, opened 2025-04-22
fix: add types to geo package

melkstam:
This solves an issue where the types for the `@nivo/geo` package was not included.

I tested it by opening `website/src/pages/choropleth/index.tsx` on master and on this branch. Where I did get an error about missing types in my editor on master, the error was not present on this branch.

To solve it, I copied the way it was done in `@nivo/line`, which seems to be one of the few places where an `index.d.ts` is used as well.


plouc:
@melkstam, good catch! Thank you again!

PR #2721, opened 2025-04-22
fix: add extension for lodash imports in funnel package

melkstam:
This fixes what I think is a simple merge miss.

https://github.com/plouc/nivo/pull/2704 added these two imports, but https://github.com/plouc/nivo/pull/2712 did at the same time add `.js` to all lodash imports.

This causes a build error in one of my projects.


plouc:
@melkstam, good catch! Thank you!

Issue #2707, opened 2025-04-07
Incorrect line points position


ignatevdev:
**Describe/explain the bug**
When plotting a line consisting of many points I experience a problem where some points are positioned incorrectly. Each point's `y` value is bigger than the previous one, but some points are renderred as if their `y` values are the same or the next one's is slightly smaller.

My guess was that the issue happens because of very small `y` values, but even if you multiply them to remove the decimal point completely the issue still persists.

**To Reproduce**
I've created a sandbox with reproduction of this problem. If you zoom in a little bit you will see that points grow in "steps", when each point should be higher than the previous one by approximately the same distance.
https://codesandbox.io/p/sandbox/dark-cache-z368sd

**Expected behavior**
Points should be drawn according to their values.

**Screenshots**
This is how the line generally looks like currently:
<img width="1064" alt="Image" src="https://github.com/user-attachments/assets/4a176203-ef21-4e3d-8d7b-2e46a47388b1" />

And this is an example of two specific points that are drawn incorrectly. 
By the values on the tooltip you can see that their y is different and the second point should be placed higher, but it's not.
<img width="858" alt="Image" src="https://github.com/user-attachments/assets/24427359-4ed4-403d-8b02-8c03259a52a7" />
<img width="800" alt="Image" src="https://github.com/user-attachments/assets/04ee1b47-5759-430f-95c4-1b837f36cb3f" />

**Desktop**
 - OS: macOS 15.1 
 - Browser: Chrome
 - Version: 134.0.6998.166

Thanks in advance!

sergey-tuna:
I can confirm that it's a very terrible issue. Please, help to resolve it.

plouc:
The issue seems to be related to [this line](https://github.com/plouc/nivo/blob/master/packages/scales/src/linearScale.ts#L32), rounding should be an option rather than a default.

sergey-tuna:
Yes, rounding should be an option, you are right.

ignatevdev:
Can confirm, after using a custom line layer with a custom scale that uses `range` instead of `rangeRound` the issue is fixed.
Should I make a PR to add "round" option to LinearScaleSpec? @plouc

plouc:
@ignatevdev, that would be great!

plouc:
Fixed by https://github.com/plouc/nivo/pull/2722.

PR #2722, opened 2025-04-23
feat(scales): add `round` option to linear scale spec

dkamyshov:
Hi @plouc,

Thank you for the great library. This PR addresses the issue raised in https://github.com/plouc/nivo/issues/2707 - we are also observing the same behaviour in our product.

- The name of the property is the same as `ScaleBandSpec.round` for consistency.
- The property is enabled by default to avoid breaking the existing codebases.

Before:

Screenshot 2025-04-23 at 11 01 11

After:

Screenshot 2025-04-23 at 11 01 24

Let me know if there is anything else that needs to be updated in order for this PR to land.

---

Rounding the output values to the nearest integer causes artifacts on charts with high-frequency data. The new option allows for this behaviour to be disabled.

### DIFF
--- a/packages/scales/package.json
+++ b/packages/scales/package.json
@@ -22,9 +22,11 @@
     ],
     "dependencies": {
         "@types/d3-scale": "^4.0.8",
+        "@types/d3-interpolate": "^3.0.4",
         "@types/d3-time": "^1.1.1",
         "@types/d3-time-format": "^3.0.0",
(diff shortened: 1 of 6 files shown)


plouc (approved):
LGTM, I'll try to add the option to the playground later.

plouc:
@dkamyshov, thank you for your contribution!

PR #2734, opened 2025-04-29
feat(line): improve types for custom layers.

pcorpet:
After version 0.92, our custom line charts are missing some pieces to get their types properly.


plouc (approved):
LGTM

PR #1978, opened 2022-04-22
@nivo/rects and @nivo/icicles package

lsagetlethias:
image
image

Highly inspired from @nivo/sunburst, @nivo/arcs, and https://observablehq.com/@d3/zoomable-icicle.

The main goal was to add a "flat" version of the Sunburst chart.

This PR is pretty complete but some TODO remains which are not critical and I don't know how to resolve them for now:
- `@nivo/rects`:
  - [ ] Add transition mode like what's arcs package has builtin
  - [ ] Create a `useAnimatedRect` hook for single rect animation
- `@nivo/icicles`
  - [ ] Figure out why text colors (e.g. "white", "black") are not working with `borderColor` props (react-spring is during `useTransition` with text spring values I guess?)
  - [ ] Canvas version
- website
  - [ ] add icons and captures for Icicles
  
 @plouc Tell me if anything is missing.
 
 Shout out to my buddy and co author of this packages: @mehdilouraoui.

PR #2745, opened 2025-05-01
@nivo/icicle & @nivo/rects

plouc:
Adjustments to the icicles package created by @lsagetlethias and @mehdilouraoui via https://github.com/plouc/nivo/pull/1978.

An advanced example of what can be achieved with this new package:

CleanShot 2025-05-08 at 08 31 23@2x


plouc:
@lsagetlethias, thank you again for introducing this new package! It's now available as `@nivo/icicle@0.94.0`.

I've made a few adjustments to the props, and added zooming as a built-in feature, I also implemented an HTML version which is more flexible for customizations (as you can see in [this example](https://nivo.rocks/storybook/?path=/story/iciclehtml--custom-node) 😄).

PR #2757, opened 2025-05-12
Add support for a ref for several charts

plouc:
- `Bar`
- `AreaBump`
- `Bump`
- `BoxPlot`
- `Bullet`
- `Calendar`
- `CalendarCanvas`
- `TimeRange`
- `Chord`
- `ChordCanvas`
- `PolarBar`
- `Sankey`


ImLunaHey:
@plouc curious why `forwardRef` was added when it's [deprecated in react 19](https://react.dev/blog/2024/12/05/react-19#ref-as-a-prop) and will be removed in future versions. seems like a step back towards react 19 support.

plouc:
@ImLunaHey, I still want to support refs for older react versions, which require using `forwardRef`.
My understanding is that it will be deprecated in future versions.

Issue #2781, opened 2025-07-17
Bug - Bullet adding random more bars at the end


JackieNim:
**Describe/explain the bug**
Given a range of [5,10,15] and a measure of 7, 4 bars appear
Also happens with [1,2,162] with a measure of [161]

**To Reproduce**
https://codesandbox.io/p/devbox/nivo-bullet-example-forked-tknyy4?workspaceId=ws_24jadn2Vj71gxoQLuKt5Wd

**Expected behavior**
Only 3 bars should appear when 3 values are given to range

JackieNim:
Bump - Waiting for PR to be merged

ACerka-Springbuk:
We're also seeing this problem. I was thinking it was related to [this change](https://github.com/plouc/nivo/commit/eeded8ebd82c69757998b97a998135ef64b890a7#diff-e0e8f10532080421981b5bb34e337e49b77fd66947a66db91519e4b26a4adce0L25-R25) where the round was added to the create linear scale. The bullet chart now rounds the end point to a fairly arbitrary point. When you're using the sequence values ("seq:cool") it's easily missed and we probably wouldn't have noticed it as easily because it just extends the sequence. But when you're using an array of colors for the specific ranges, the 'rounded' area seems to use the first color for the extra area that gets added. Either way this will cause issues for people as there may be other UI elements placed around the chart designed to appear at specific locations on the chart that won't line up anymore.

JackieNim:
> We're also seeing this problem. I was thinking it was related to [this change](https://github.com/plouc/nivo/commit/eeded8ebd82c69757998b97a998135ef64b890a7#diff-e0e8f10532080421981b5bb34e337e49b77fd66947a66db91519e4b26a4adce0L25-R25) where the round was added to the create linear scale. The bullet chart now rounds the end point to a fairly arbitrary point. When you're using the sequence values ("seq:cool") it's easily missed and we probably wouldn't have noticed it as easily because it just extends the sequence. But when you're using an array of colors for the specific ranges, the 'rounded' area seems to use the first color for the extra area that gets added. Either way this will cause issues for people as there may be other UI elements placed around the chart designed to appear at specific locations on the chart that won't line up anymore.

I made a [PR](https://github.com/plouc/nivo/pull/2783) here but I dont think it'll be merged anytime soon

PR #2783, opened 2025-07-25
Fix the default value of nice in linearScale

JackieNim:
This reverts the default value of `nice` to false 

Orginally set to true in this [PR](https://github.com/plouc/nivo/commit/0e5d3672eb920ebe1adce731212277ebd67cdb8a#diff-aee0d14423c59380cbad3f23d2e15d0ecee7eb919256d94d3dcf59aa6ad36865)

This also fixes the issue #2781 where an extra section would be added to the bullet.

Before:
image

After:
image

### DIFF
--- a/packages/bullet/src/hooks.ts
+++ b/packages/bullet/src/hooks.ts
@@ -22,7 +22,7 @@ export const useEnhancedData = (
                 const min = minValue ?? Math.min(...all)
 
                 const scale = createLinearScale(
-                    { clamp: true, min, max, type: 'linear', round: true },
+                    { clamp: true, min, max, type: 'linear', round: false },
                     { all, max, min },
                     layout === 'horizontal' ? width : height,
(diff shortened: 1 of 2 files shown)


JackieNim:
@plouc Please take a look

ek-hystax:
I ran into a similar issue with my line chart after updating to version 0.99.0. Restoring the previous behavior was possible by setting the `nice` property back to false.

| Before update | After |
|--------|--------|
|  |  | 

@JackieNim Thanks for opening this PR - it helped me identify the issue and figure out how to fix it

JackieNim:
Bump

plouc:
I don't think we should change the default for all packages, it would be best to just override for the bullet chart.

JackieNim:
> I don't think we should change the default for all packages, it would be best to just override for the bullet chart.

Updated

PR #2785, opened 2025-07-31
fix(line): include lodash in dependencies

blx:
Lodash is imported in packages/line/src/hooks.ts (as 'lodash/uniqueId.js') but was not included in `line`'s dependencies.

Other packages within nivo, like `bar` or `annotations`, also import from lodash, but they already include lodash in their deps.

Without this fix, I can't import the ResponsiveLine component in the nextjs project I'm working on (which already depends on lodash from elsewhere, but is not happy for `@nivo/line` to try to import lodash without declaring it as a dep).

### DIFF
--- a/packages/line/package.json
+++ b/packages/line/package.json
@@ -47,7 +47,8 @@
         "@nivo/voronoi": "workspace:*",
         "@react-spring/web": "9.4.5 || ^9.7.2 || ^10.0",
         "@types/d3-shape": "^3.1.6",
-        "d3-shape": "^3.2.0"
+        "d3-shape": "^3.2.0",
+        "lodash": "^4.17.21"
     },
     "peerDependencies": {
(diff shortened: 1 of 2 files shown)


blx:
bump

caillou:
I ran into this issue as well. I think it would be nice to have this merged in.

It worked as long as my project had `lodash` itself. I dropped `lodash`, and it generated an error.

For now, my work-around is the following:

```yaml
# pnpm-workspace.yaml

# @nivo/line 0.99.0 imports lodash/uniqueId.js without declaring lodash; every other @nivo package
# that deep-imports lodash declares it. Supply the missing declaration rather than adding lodash to
# our own dependencies, which would misstate the graph.
packageExtensions:
  '@nivo/line':
    dependencies:
      lodash: '*'
```

Issue #2685, opened 2025-02-05
Add option for calling `d3-scale` `.nice()` for axis log type


lharress:
**Is your feature request related to a problem? Please describe.**
The line logarithmic scale type automatically calls `.nice()` from `d3-scale` ([source](https://github.com/plouc/nivo/blob/master/packages/scales/src/logScale.ts#L50)) which (depending on the data set and base) adds additional ticks to the axis and grid. In some scenarios this is not ideal.

**Describe the solution you'd like**
Optionally call `.nice()` based on an axis config like the time scale ([source](https://github.com/plouc/nivo/blob/master/packages/scales/src/timeScale.ts#L43-L44))

**Describe alternatives you've considered**
Have tried various other options including manually setting the grid/ticks but it's cumbersome and don't work as well with logarithmic scaling as the data tends to also need changing.

**Additional context**
Happy to create the PR myself, just keen to see if there's any objection.

PR #2795, opened 2025-09-29
make nice optional for createLogScale

webxl:
Similar to what was done for https://github.com/plouc/nivo/issues/2685 but this would be for regular log scales. I can make `nice` default to true to keep it consistent.


webxl:
Never mind. Looks like it was already fixed [here](https://github.com/plouc/nivo/commit/0e5d3672eb920ebe1adce731212277ebd67cdb8a)

PR #2824, opened 2026-03-26
fix(line): correct pointSymbol datum type to match runtime data

### LINKED ISSUE #2818: Mismatching type for `pointSymbol` from `ResponsiveLine`
**Describe/explain the bug**

The type for the `pointSymbol` prop from `ResponsiveLine<T>` is wrong.
Example:

```ts
type ChartDatum = {
  x: string;
  y: number;
  stalled: boolean;
};

type ChartSeries = {
  id: string;
  data: readonly ChartDatum[];
};

// ...

return (
  <ResponsiveLine<ChartSeries>
    pointSymbol={({ datum, size, borderWidth }) => {
      // @ts-expect-error
      console.log('Stalled data is at:', datum.stalled); // datum.stalled -> any
      console.log('But the typing thinks it is at:', datum.data?.stalled); // datum.data?.stalled -> boolean

      return (
        <circle
          r={size / 2}
          fill="#0e7490"
          stroke="#22d3ee"
          strokeWidth={borderWidth} />
      );
    }}
```

**To Reproduce**

1. `git clone git@github.com:macabeus/mizuchi.git`
2. `git checkout a7365f5`
3. Check the `as never` on `src/report-generator/ui/components/AttemptsChart.tsx`

**Expected behavior**
No `as never` should be necessary to type this code

restareaByWeezy:
Closes #2818

The `pointSymbol` prop was typed as `DotsItemSymbolComponent<Point<Series>>`, but at runtime the `datum` passed to the symbol component is actually `Point<Series>['data']` (i.e. the raw data item), not the full `Point` object. You can see this in `Points.tsx` where `datum: point.data` is mapped before passing to `DotsItem`.

Changed the type to `DotsItemSymbolComponent<Point<Series>['data']>` so `datum.stalled` (or whatever custom fields are on the data) is correctly typed without needing `datum.data.stalled`.

Also removed the `as any` cast on the `symbol` prop in `Points.tsx` since the types now line up.


plouc:
Thank you  for the fix @restareaByWeezy!

PR #2825, opened 2026-03-26
fix(calendar): show tooltip when valueFormat returns a string

### LINKED ISSUE #2812: calendar: CalendarTooltip doesn't support valueFormat
**Describe/explain the bug**

When using `Calendar` or `TimeRange`, you can provide a `valueFormat` function:

```js
(v: number) => string;
```

In `CalendarTooltip.tsx`, value is assumed to **always** be a number:

https://github.com/plouc/nivo/blob/2cf224355087b02aab204ced0d04d05d3d081b08/packages/calendar/src/CalendarTooltip.tsx#L6

This will return `null` when using a `valueFormat` function because it returns a string, not a number, which fails to show a tooltip.

**To Reproduce**

https://codesandbox.io/p/sandbox/mystifying-dhawan-4qhh8x

1. Go to csb
2. Hover or try to click on time range day
3. Observe no tooltip is shown

**Expected behavior**

A tooltip should be shown

**Workaround**

Copy original tooltip and make your own without the `isNaN` check

restareaByWeezy:
Closes #2812

CalendarTooltip was checking `isNaN(Number(value))` before rendering, but when `valueFormat` is used, the value is already a formatted string (e.g. "1,234 items"). `Number("1,234 items")` is NaN, so the tooltip silently returned null.

Since the `value` prop is typed as `string` and already goes through `valueFormat` before reaching the tooltip, the `isNaN` check doesn't make sense here. Removed it so the tooltip renders whenever `value` is defined.

### DIFF
--- a/packages/calendar/src/CalendarTooltip.tsx
+++ b/packages/calendar/src/CalendarTooltip.tsx
@@ -3,6 +3,6 @@ import { CalendarTooltipProps } from './types'
 import { memo } from 'react'
 
 export const CalendarTooltip = memo(({ value, day, color }: CalendarTooltipProps) => {
-    if (value === undefined || isNaN(Number(value))) return null
+    if (value === undefined) return null
     return <BasicTooltip id={day} value={value} color={color} enableChip={true} />
 })


plouc:
Good catch.

